PR #57, opened 2019-05-27
Support the rest of binding types

### LINKED ISSUE #45: Support the rest of binding types
Extend `Ormolu.Printer.Meat.Declaration.Value` by finishing `p_valDecl'` to support properly all bindings.

mrkkrp:
Close #45.


mrkkrp:
OK, I believe this is not perfect, but it's a decent approximation and certainly will do for most practical cases. We can always return and adjust stuff later.

PR #59, opened 2019-05-28
Implement TypeApplications and StaticPointers support

utdemir:
Hey,

I tried to use ormolu on one my projects, and saw a few unimplemented constructs and decided to implement it.

These are pretty small changes, since StaticPointers is like an ordinary function application and TypeApplications is like an ordinary parameter.

I do not know if you prefer separate PR's for these two, but the change is small enough that a single PR can be easily reviewed/merged.

---

Also a quick question, I want to try implementing a few more constructs, however they are more complex. I believe I can still come up with working implementations, but they will be less than ideal and you will need to go through them again. Would you prefer that, or do you prefer you or someone else to implement them from scratch?

### DIFF
--- a/data/examples/declaration/value/function/static-pointers-out.hs
+++ b/data/examples/declaration/value/function/static-pointers-out.hs
@@ -1,0 +1,18 @@
+{-# LANGUAGE StaticPointers #-}
+foo :: StaticPtr Int
+foo = static 5
+
+bar :: StaticPtr [Int]
+bar =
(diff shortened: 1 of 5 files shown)


mrkkrp:
>  you will need to go through them again

Well, that's called "reviewing"! You're very welcome to work on this and it'll be still faster for me to review your work than doing it from scratch, especially that I cannot work on this full-time anyway.

PR #60, opened 2019-05-29
Initial support for do notation

utdemir:
I hacked something together until we can parse `do` notation.

It currently doesn't look pretty because I couldn't find a way to put the `do` keyword in the same line as the previous line; since, as far as I understand, `ormolu` thinks everything should be multi-lined since the whole construct spans multiple lines. I'd appreciate if you can point me to a solution.

### DIFF
--- a/data/examples/declaration/value/function/do-out.hs
+++ b/data/examples/declaration/value/function/do-out.hs
@@ -1,0 +1,22 @@
+{-# LANGUAGE RecursiveDo #-}
+bar = do
+  foo
+  bar
+
+baz = mdo
+  bar a
+  a <- foo
+  b <-
+    bar
+      1
+      2
+      3
+  return (a + b)
+
+baz = do
+  a <- foo
+  let b = a + 2
+      c = b + 3
+  bar c
+  let d = c + 2
+  return d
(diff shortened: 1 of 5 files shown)


mrkkrp:
I'll get to reviewing this and thinking how to make it pretty but probably not today. A bit of patience!

mrkkrp:
I rebased against current master. The PR doesn't seem to trigger CI runs because it's from a fork. I'll see about enabling building from forks.

The idempotency properly requires multi-line things to map to multi-line outputs and the same for single line. So if you have `do { foo ; bar }` it must print back something on single line, probably using this notation with braces.

Other than that, perhaps we could leave the issue with leaving `do` on the same line for later.

utdemir:
Thank you.

I am happy to fix the PR to handle single-line do properly too, it shouldn't take much effort. I'll take a look at it either tomorrow or the next day.

mrkkrp:
I thought of doing some work on top of the PR and then merge it. I think it may require some non-trivial changes to get right.

mrkkrp:
@utdemir Here we go, I think I implemented all the bits related to "normal" do blocks, that is, without curly braces. I'll leave it to you to add support for the alternative syntax. Let me know if you have any feedback or questions.

mrkkrp:
On a second thought though, I think since the tests pass now, we could merge this one for now and leave the remaining bits to be implemented in a future PR.

PR #67, opened 2019-06-05
Change formatting of operators and infix constructors

mrkkrp:
This formatting scheme is more appropriate if you think of chaining multiple infix expressions and combining it with do-blocks. The operator should hang on previous line instead going first on the next line, e.g.:

```
foo +
  bar
```

instead of

```
foo
  + bar
```

### DIFF
--- a/data/examples/declaration/value/function/guards-out.hs
+++ b/data/examples/declaration/value/function/guards-out.hs
@@ -6,9 +6,9 @@ foo x
 bar :: Int -> Int
 bar x
   | x == 5 =
-    foo x
-      + foo 10
+    foo x +
+      foo 10
   | x == 6 =
-    foo x
-      + foo 20
+    foo x +
+      foo 20
   | otherwise = foo 100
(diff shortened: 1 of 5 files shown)


aspiwack:
Actually, I think that I prefer infix-operator-first. It seems to me the most common convention, both in Haskell and pen-and-paper mathematics.

What prompted the change?

mrkkrp:
The decision is apparent if we think of this sort of situation involving `do`-notation:

```haskell
foo = something $ do
  ...
```

The new formatting scheme results in the style above, while the old one would result in this:

```haskell
foo = something
  $ do
  ...
```

While perhaps it could be argued that we could make some cases special, I think it's simpler to just adjust it like this and enjoy uniform handling.

aspiwack:
I realise it's hard to deal with, and not a high priority. But, just for the record, what I typically expect to see is that:

```haskell
foo = something + do
   …
```

While

```haskell
foo =
   something long
     + something long too
```

So really, it seems to me that do blocks (and, probably, any hanging symbol) wants to pretend that is always written on a single line, for the sake of its context. So, you'd format an expression containing a do block as if there was just the `do` symbol, and ignore the rest of the block.

It doesn't seem quite possible, if I understand correctly, to handle within the current code base. But I think it is something to keep in mind somehow.

Issue #49, opened 2019-05-26
Implement rendering of type class and instance declarations


mrkkrp:
Extend `p_tyClDecl` and `p_hsDecl` in `Ormolu.Printer.Meat.Declaration` in order to support type class and instance declarations.

yumiova:
I have a working prototype of type class (and type family) instance declarations in [yumiova:yumiova/instance-declaration](https://github.com/yumiova/ormolu/tree/yumiova/instance-declaration). It is not, however, something I consider currently ready for merging.

yumiova:
OK, I am more or less done with the definition of instance declarations (aside from perhaps adding more tests.) I also ended up implementing associated types, which made it seem natural to implement type family instances as well. However, I have not implemented associated data types. Such a change would either require code duplication or a slightly invasive change to `p_dataDecl`.

If you want, I can send a pull request for this individually. It makes no difference to me, since I want to try to implement type class declarations regardless. Either both can be implemented in one great pull request, or it can be two separate pull requests.

Also, looking at the forks, it seems like @fghibellini has a [branch](https://github.com/fghibellini/ormolu/tree/instance-decls) that works on the same thing? I hope I haven't accidentally caused some misunderstanding by not mentioning my interest in this soon enough.

mrkkrp:
I'd encourage you to send PRs with ready bits as soon as possible.

yumiova:
I am currently working on type class declarations in [yumiova:yumiova/class-declaration](https://github.com/yumiova/ormolu/tree/yumiova/class-declaration).

mrkkrp:
Right now, what remains to implement for this one is just type class declarations.

mrkkrp:
Done in #79.

PR #72, opened 2019-06-06
Implement rendering of instance declarations

yumiova:
This pull request implements half of #49. Specifically, it deals with pretty printing instance declarations and most structures needed to support that. This includes pretty printing of associated types (and therefore also type family instances.) However, this _does not_ include pretty printing of associated data types.

Pretty printing of type class declarations is to be dealt with at a later time.

Instances are formatted such that indentation hangs after the `instance`. Namely,

```haskell
instance Eq a
         => Eq [a] where
```

To support pretty printing associated types, the function `p_tyFamInstEqn` was exported from the appropriate module. This was done to avoid redundant code duplication.

As for supporting pretty printing of associated data types: there is a potential (but unimplemented) solution that works while minimizing code duplication by changing the `p_dataDecl` function, which currently has the signature:

```haskell
p_dataDecl :: Located RdrName -> LHsQTyVars GhcPs -> HsDataDefn GhcPs -> R ()
```

If the parameters to `p_dataDecl` instead were any `HsType`, then the associated data patterns could be pretty printed using `p_dataDecl`. Fortunately, type variables can be directly represented in `HsType`s.

Furthermore, if `p_dataDecl` took an extra parameter to represent the context (or style) it is placed in, it would be trivial to support open data family instances as well, providing a final signature:

```haskell
p_dataDecl :: FamilyStyle -> Located RdrName -> [LHsType GhcPs] -> HsDataDefn GhcPs -> R ()
```

In principle, making this change should not take me long at all, and would give complete support for both associated data and open data family instances.

That said, I feel like larger changes to e.g. type signatures and use cases like this should be agreed upon ahead of time. Hence, why I have not implemented this yet.

This branch has had its commit history squashed. See [yumiova:yumiova/instance-declaration](https://github.com/yumiova/ormolu/tree/yumiova/instance-declaration) for the original commit history.

### DIFF
--- a/data/examples/declaration/instance/associated-data-out.hs
+++ b/data/examples/declaration/instance/associated-data-out.hs
@@ -1,0 +1,15 @@
+{-# LANGUAGE TypeFamilies #-}
+instance Foo Int where
+  data Bar Int = IntBar Int Int
+
+instance Foo Double where
+  newtype Bar Double
+    = DoubleBar
+        Double
+        Double
+
+instance Foo [a] where
+  data Bar [a]
+    = ListBar [Bar a]
+  data Baz [a]
+    = ListBaz
--- a/data/examples/declaration/instance/associated-data.hs
+++ b/data/examples/declaration/instance/associated-data.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE TypeFamilies #-}
+
+instance Foo Int where data Bar Int  =  IntBar Int Int
+
+instance Foo Double where
+    newtype
+      Bar
+        Double
+        =
+          DoubleBar
+            Double
+            Double
+
+instance Foo [a]
+  where
+    data Bar [a] =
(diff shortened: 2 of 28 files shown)


yumiova:
Just realised some tests I need to add; give me a few minutes.

mrkkrp:
Take your time, I'm done for today, will review it tomorrow. I also sent you an invitation to become a collaborator. This will give you write access to the repo.

mrkkrp:
This looks good. I went ahead and relaxed comparison of ASTs to allow differences in `SourceText`s, which should be pretty harmless.

I think it makes sense to change `p_dataDecl` in the way you describe and so gain support for data family instances too. Could you add this in this PR? Once that is added, I think we can merge.

yumiova:
OK, this took me a little bit longer than expected, but support for associated data types and open data family instances should now be working.

Issue #50, opened 2019-05-26
Implement support for rendering of standalone deriving declarations


mrkkrp:
Extend `p_hsDecl` in `Ormolu.Printer.Meat.Declaration` to support rendering of standalone deriving declarations.

mrkkrp:
Done in #50.

PR #78, opened 2019-06-09
[#50] Standalone Deriving

sphaso:
Hi, I hope you don't mind if I take a stab at this.     
I opened the PR just to check if I'm going in the right direction, there's a few things still missing:    
- still need to figure out when is `XDerivDecl` matched
- there might be a refactoring opportunity for the deriving strategy cases (e.g. in [Data.hs:140-165](https://github.com/tweag/ormolu/blob/fab9f15ad6e34165a57b0151de3503710b82898a/src/Ormolu/Printer/Meat/Declaration/Data.hs#L140-L165)) although I don't think all strategies are available \ compatible for standalone deriving (right?)
- it looks like `located (hsib_body (hswc_body deriv_type)) p_hsType` is too crude here since it's not parsing the `via` between the types and thus not triggering a newline

### DIFF
--- a/data/examples/declaration/deriving/multiline-out.hs
+++ b/data/examples/declaration/deriving/multiline-out.hs
@@ -1,0 +1,44 @@
+deriving instance
+  Eq
+    Foo
+
+deriving stock instance
+  Show
+    Foo
+
+deriving anyclass instance
+  ToJSON
+    Foo
+
+deriving newtype instance
+  Data
+    Foo
+
+deriving instance
+  {-# OVERLAPPABLE #-}
+  Ord
+    Foo
+
+deriving instance
+  {-# OVERLAPPING #-}
+  Num
+    Foo
+
+deriving instance
+  {-# OVERLAPS #-}
+  Read
+    Foo
+
+deriving instance
+  {-# INCOHERENT #-}
+  Show
+    Foo
+
+deriving via
+  Foo
+    Int
+  instance
+    Triple
+      A
+      B
+      C
--- a/data/examples/declaration/deriving/multiline.hs
+++ b/data/examples/declaration/deriving/multiline.hs
@@ -1,0 +1,36 @@
+deriving instance Eq
+                    Foo
+
+deriving stock instance
+  Show
+    Foo
+deriving anyclass instance
+  ToJSON
+    Foo
+deriving newtype instance
+  Data
+    Foo
+
+deriving instance
+  {-# OVERLAPPABLE #-}
+    Ord
+      Foo
+deriving instance
+  {-# OVERLAPPING #-}
+    Num
+      Foo
+deriving instance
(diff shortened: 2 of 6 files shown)


yumiova:
Just letting you know that `XDerivDecl` is irrelevant for us. Essentially, all syntax tree cases that begin with `X` are intended to allow libraries to add additional information about the syntax trees. However, we simply use GHC's parser directly, which means that most of these cases are *never used*.

This includes `XDerivDecl`, so the way you handled that is in fact correct.

sphaso:
@yumiova thanks for the info!

src/Ormolu/Printer/Meat/Declaration/Deriving.hs
```diff
@@ -0,0 +1,38 @@
...
+  d@DerivDecl {} -> p_standaloneDerivDecl d
+  XDerivDecl _ -> notImplemented "XDerivDecl standalone deriving"
+
+p_standaloneDerivDecl :: DerivDecl GhcPs -> R ()
+p_standaloneDerivDecl DerivDecl {..} = line $ do
+    txt "deriving"
+    case deriv_strategy of
+      Nothing ->
+        breakpoint
+      Just l -> locatedVia Nothing l $ \case
```
mrkkrp: I don't understand how it renders the other cases given that this case expression doesn't implement some matches? How is it capable of rendering "deriving via" instances for example? Is `via` included in the `deriv_type` somehow? Once we (or well, I) understand this, it'll be easier to tell whether or not we can re-use existing rendering logic for `DerivStrategy`.

If `located (hsib_body (hswc_body deriv_type)) p_hsType` is too crude I propose we infer approximately the span that includes everything by using the `combineSrcSpans'` function. It works like this: combine spans of individual elements and then feed that to `switchLayout` that should be put around the entire thing I guess. Here is an example:

https://github.com/sphaso/ormolu/blob/feature%2F50_standalone_deriving/src/Ormolu/Printer/Meat/Declaration/Data.hs#L51

Another important thing that is missing is handling of `deriv_overlap_mode`. Overlapping modes should be handled similar to how it's done for instance declarations:

https://github.com/sphaso/ormolu/blob/feature%2F50_standalone_deriving/src/Ormolu/Printer/Meat/Declaration/Instance.hs#L33

Perhaps some part of it could be moved to a standalone function and reused.

mrkkrp:
@sphaso Are you going finish it up?

sphaso:
@mrkkrp yes, that would be my intention. I was too busy these past days.
I'll work on it during ZuriHac. See you there? :)

mrkkrp:
@sphaso Sounds good!

mrkkrp:
I squashed it, rebased the code, and updated the examples. Could you tweak it a bit so the updated examples pass?

PR #79, opened 2019-06-09
Implement rendering of type class declarations

yumiova:
This pull request implements the latter half of #49. Specifically, it deals with pretty printing type class declarations. This includes pretty printing the associated type and data families.

Like #72, type class declarations are formatted such that indentation hangs after the `class`:

```haskell
class ( Ord a
      , Show a
      )
      => Foo a where
```

Functional dependencies over multiple lines use a style that is debatable. Specifically,

```haskell
class Bar a b
      | a -> b
      , b -> a where
```

Obviously feel free to request a change of style if desired.

The only other thing of note aside from that is this: I ended up duplicating the the `tyVarsToTypes` helper function. Alternatively, I could move it and export it from either the `Ormolu.Printer.Meat.Common` or `Ormolu.Printer.Meat.Type` modules.

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Class.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Class.hs
@@ -1,0 +1,94 @@
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards   #-}
+
+-- | Rendering of type class declarations.
+
+module Ormolu.Printer.Meat.Declaration.Class
+  ( p_classDecl
+  )
+where
+
+import Class
+import Control.Arrow
+import Control.Monad
+import Data.Foldable
+import Data.Function
+import Data.List (sortBy)
+import Ormolu.Printer.Combinators
+import Ormolu.Printer.Meat.Declaration.Signature
+import Ormolu.Printer.Meat.Declaration.Value
+import Ormolu.Printer.Meat.Declaration.TypeFamily
+import Ormolu.Printer.Meat.Common
+import Ormolu.Printer.Meat.Type
+import Ormolu.Utils
+import GHC
+import RdrName (RdrName (..))
+import SrcLoc (Located, combineSrcSpans)
+
+p_classDecl
+  :: LHsContext GhcPs
+  -> Located RdrName
+  -> LHsQTyVars GhcPs
+  -> [Located (FunDep (Located RdrName))]
+  -> [LSig GhcPs]
+  -> LHsBinds GhcPs
+  -> [LFamilyDecl GhcPs]
+  -> [LTyFamDefltEqn GhcPs]
+  -> R ()
+p_classDecl ctx name tvars fdeps csigs cdefs cats catdefs = do
+  let HsQTvs {..} = tvars
+      combinedSpans =
+        getLoc ctx `combineSrcSpans`
+        getLoc name `combineSrcSpans`
+        foldr (combineSrcSpans . getLoc) noSrcSpan hsq_explicit `combineSrcSpans`
+        foldr (combineSrcSpans . getLoc) noSrcSpan fdeps
+  txt "class "
+  sitcc $ do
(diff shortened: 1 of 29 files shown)


src/Ormolu/Printer/Meat/Declaration/Class.hs
```diff
@@ -0,0 +1,105 @@
...
+p_funDep :: FunDep (Located RdrName) -> R ()
+p_funDep (before, after) = do
+  spaceSep p_rdrName before
+  txt " -> "
+  spaceSep p_rdrName after
+
+----------------------------------------------------------------------------
+-- Helpers
+
+tyVarsToTypes :: LHsQTyVars GhcPs -> [LHsType GhcPs]
```
mrkkrp: Yeah, please go ahead and move these to `Ormolu.Printer.Meat.Type` to avoid duplication.

yumiova:
I should be able to apply the requested changes by thursday evening (i.e. before friday.) That said, I am not entirely sure if I understand the third comment (about functional dependencies.) Consequently, it would be nice if you could give relevant input and the desired output exactly.

mrkkrp:
I think I meant that if we feed

```haskell
class (Stream s, MonadPlus m) => MonadParsec e s m
      | m -> e s where
```

we currently get

```haskell
class (Stream s, MonadPlus m)
      => MonadParsec e s m | m -> e s where
```

while the expected output would be something like

```haskell
class (Stream s, MonadPlus m) => MonadParsec e s m
      | m -> e s where
````

I.e. the beakpoint should be before `|`, not before `=>` in this case.

**Comment by yumiova** [comment]
OK, I have applied some changes to the pull request. Basically, there are two changes:

1. I obviously added some basic comments. Most of the comments are in pretty non-controversial locations. I tried to add some comments in weird locations at times, though most of the output just becomes weirdly formatted from that, so I chose to refrain. I believe this has no relation to the changes from this pull request.
2. In the original code, I stupidly forgot to call switchLayout when pretty printing the contexts and functional dependencies. Consequently, if any part of the entire class declaration (including the methods) was multi-line, the contexts and functional dependencies would be layed out over multiple lines. Obviously, I have fixed this.

   Otherwise, I thought it best to not over-complicate the handling of breakpoints inside the `class ... where`. This also seems to be in-line with what Ormolu generally already follows, which is my main priority. Consequently, if any part inside the `class ... where` area is multi-line, the entire block gets breakpoints at every relevant place. In other words,

   ```haskell
   class (Stream s, MonadPlus m)
         => MonadParsec e s m
         | m -> e s where
   ```

   Personally, I am still a little luke-warm about this formatting. However, it—again—seems to be in-line with what Ormolu already does. Obviously, feel free again to request a change of style.

**Comment by yumiova** [comment]
It seems like the general style Ormolu is using for pretty printing e.g. type families is

```haskell
type family
  Id a
    = result
    | result -> a where
  Id a =
    a
```

Maybe type classes and instances should be formatted in a similar manner? As always, maintaining consistency is my main goal. If formatting should change, then I recommend that I do that in another pull request, since the eventual change would not only change type classes, but also instances.

yumiova:
After looking at recent changes to master, I am making slight changes to keep things consistent with the rest of the codebase. Maybe it will be ~1 hour or so.

yumiova:
OK, I am done applying some simple refactorings and cleanup after recent updates to master. I believe this is ready for review once again.

yumiova:
I am not entirely sure what happened, but recent commits seems to have been reverted after a force-push from @mrkkrp? I know for a fact that the checks passed before.

What exactly did the force-push do? I can replicate the force-pushed changes across the updated commits, if you would like.

mrkkrp:
@yumiova I'm not sure, I rebased this PR against current master and maybe adjusted some bits. Anyway, I pushed some more stuff and now it's green and I'm going to merge it.

yumiova:
For the record, the commit 3685ac0 also had a test for type operators & infix symbols, which was lost in 3f7c8bd. There is probably no reason to care that much, but just wanted to mention it.

PR #81, opened 2019-06-10
Format SigPat patterns

mitchellwrosen:
This patch formats `SigPat` patterns, which appear (for example) when using `ScopedTypeVariables` + `do`-notation:

```haskell
do
  input :: String <- getLine
  putStrLn input
```

I reused part of the `p_typeSig` formatter, which I named `p_typeAscription`, that just formats the `:: type` half of such an expression/declaration.


mrkkrp:
Could you perhaps add some tests too?

mitchellwrosen:
Of course. I'll add some, then force-push over this commit.

PR #84, opened 2019-06-10
Fix rendering of case alternatives with guards

mrkkrp:
Also rolls back handing things in if branches, because it doesn't look so good?

cc @facundominguez


data/examples/declaration/value/function/case-multi-line-out.hs
```diff
@@ -19,3 +19,12 @@ baz x = case x of
 quux :: Int -> Int
 quux x = case x of
   x -> x
+
+withGuards :: Int -> Int
+withGuards x = case x of
+  x
```
mrkkrp: This may look a bit strange but it's in fact the right thing to do if we want to keep diffs minimal and independent of length of `x`.

PR #92, opened 2019-06-12
Fix rendering of infix constructions in lhs of definitions

### LINKED ISSUE #36: Type families with infix left-hand equations are reformatted in prefix style
```haskell
type family (x :: N) + (y :: N) :: N where
  'Zero + y = y
  'Succ n + y = 'Succ (n + y)

type family (x :: N) `LEQ` (y :: N) :: Bool where
  'Zero `LEQ` y = 'True
  'Succ n `LEQ` 'Zero = 'False
  'Succ n `LEQ` 'Succ m = n `LEQ` m
```

is reformatted as

```haskell
type family (+) (x :: N) (y :: N) :: N where
  (+) 'Zero y = y
  (+) 'Succ n y = 'Succ (n + y)

type family LEQ (x :: N) (y :: N) :: Bool where
  LEQ 'Zero y = 'True
  LEQ 'Succ n 'Zero = 'False
  LEQ 'Succ n 'Succ m = n LEQ m
```

I think the original style should be preserved in this case.

mrkkrp:
Close #36.

### DIFF
--- a/data/examples/declaration/instance/associated-types-out.hs
+++ b/data/examples/declaration/instance/associated-types-out.hs
@@ -3,7 +3,10 @@ instance Foo Int where
   type Bar Int = Double
 
 instance Foo Double where
-  type Bar Double =
-    [Double]
-  type Baz Double =
-    [Double]
+  type
+    Bar
+      Double =
+      [Double]
+  type
+    Baz Double =
+      [Double]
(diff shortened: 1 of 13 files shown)


mrkkrp:
Function definitions are now rendered properly but still need to handle type families. This is for another day.

Issue #70, opened 2019-06-06
Cover entire GHC AST


mrkkrp:
This is the umbrella issue about covering various syntactic constructions in GHC AST. It should be closed when we cover everything.

- [x] `HsModule` Top level module thing
  - [x] Module header and export list
  - [x] Imports
  - [x] Declarations
  - [x] Deprecation messages
- [x] `TyClD` Type or class declaration
  - [x] `FamDecl` Type family declaration
  - [x] `SynDecl` Type declaration
  - [x] `DataDecl` Data type declaration
  - [x] `ClassDecl` [Type class declaration](https://github.com/tweag/ormolu/issues/49)
- [x] `InstD` Instance declaration
  - [x] `ClsInstD` [Type class instance declaration](https://github.com/tweag/ormolu/issues/49)
  - [x] `DataFamInstD` Data family instance declaration
  - [x] `TyFamInstD` Type family instance declaration
- [x] `DerivD` [Deriving declaration](https://github.com/tweag/ormolu/issues/50)
- [x] `ValD` Values (binds)
  - [x] `FunBind` Function-like binding
  - [x] `PatBind` Pattern binding
  - [x] `HsExpr` Lots of stuff implemented now
  - [x] `HsExpr` [Various kinds of do-notation except for `DoExpr` and `MDoExpr`](https://github.com/tweag/ormolu/issues/47)
  - [x] `Pat` [Some types of patterns](https://github.com/tweag/ormolu/issues/46)
- [x] `SigD` [Signature declaration](https://github.com/tweag/ormolu/issues/48)
  - [x] `TypeSig` Type signatures
  - [x] `PatSynSig` Pattern-synonym signatures
  - [x] `ClassOpSig` Signatures for class methods
  - [x] `FixSig` Fixity declaration
  - [x] `InlineSig` Inline pragmas
  - [x] `SpecSig` Specialization pragmas
  - [x] `SpecInstSig` A specialisation pragmas for instance declarations
  - [x] `MinimalSig` Minimal complete definition pragmas
  - [x] `SCCFunSig` A "set cost centre" pragmas for declarations
  - [x] `CompleteMatchSig` A complete match pragma
- [x] `DefD` Default declaration
- [x] `ForD` [Foreign declaration](https://github.com/tweag/ormolu/issues/53)
  - [x] `ForeignImport`
  - [x] `ForeignExport`
- [x] `WarningD` [Warning declaration](https://github.com/tweag/ormolu/issues/52)
[...]

PR #101, opened 2019-06-14
Add MinimalSig printer

basile-henry:
This is part of #70

This is based off of #103 in order to use the latest pragma combinator.

### DIFF
--- a/data/examples/declaration/signature/minimal/minimal-out.hs
+++ b/data/examples/declaration/signature/minimal/minimal-out.hs
@@ -1,0 +1,12 @@
+class Foo a where
+  {-# MINIMAL (==) | ((/=), foo) #-}
+  {-# MINIMAL
+    a
+    | ( b, c, d
+      | e
+      , f
+      )
+    | g
+    #-}
+  (==) :: a -> a -> Bool
+  (/=) :: a -> a -> Bool
--- a/data/examples/declaration/signature/minimal/minimal.hs
+++ b/data/examples/declaration/signature/minimal/minimal.hs
@@ -1,0 +1,10 @@
+class Foo a where
+
+  {-# MINIMAL     (==)   |(  (/=) ,       foo) #-}
+
+  {-#    MINIMAL a | (b , c      , d | e
+     , f)
+    |   g  #-}
+
+  (==) :: a -> a -> Bool
+  (/=) :: a -> a -> Bool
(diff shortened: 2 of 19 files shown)


data/examples/declaration/signature/minimal/minimal.hs
```diff
@@ -0,0 +1,6 @@
+
+  {-# MINIMAL     (==)   |(  (/=) ,       foo) #-}
+
+  {-#    MINIMAL a | (b , c      , d | e
```
mrkkrp: What is this syntax with parentheses?

mrkkrp: OK, I think I understand how the syntax, although I've never seen it used.

basile-henry: I think it's just to enforce precedence, otherwise **and** has a higher precedence over **or** and you can't enforce custom minimal definition rules.

data/examples/declaration/signature/minimal/minimal-out.hs
```diff
@@ -0,0 +1,12 @@
+class Foo a where
+  {-# MINIMAL (==) | ((/=), foo) #-}
+  {-# MINIMAL
+    a
+    | ( b, c, d
```
mrkkrp: I don't understand why `d | e` is not on the same line even though it is on the same line in the input. I'm not willing to spend time figuring this out though.

basile-henry: It's because `e, f` is on multiple lines. So by extension the `|` in `b, c, d | e, f` is also on multiple lines.

It's similar to say formatting:

```haskell
[a, b, (c,
 d)]
```

to:

```haskell
[ a
, b
, ( c
  , d
  )
]
```

Or something similar

mrkkrp: So it associates like this `d | (e, f)`?

basile-henry: The `,` (**and**) has a stronger precedence over `|` (**or**), so yes it associates like this: `(b, c, d) | (e, f)`
I just used the AST for the binary function given to me by GHC, I didn't explicitly write any precedence code myself :smile:

mrkkrp: Yeah, I see that. Just wanted to confirm that my understanding is correct.

PR #102, opened 2019-06-15
[#98] Fix where blocks with do notation

### LINKED ISSUE #98: Combining where and do will fail
Input:

```haskell
f :: Maybe Int
f = do
  return c
  where
    c = 0
```

```
λ ormolu A.hs
<rendered>:3:1-5
parse error on input ‘where’
```

ruhatch:
The issue here was that `placeHanging` removed the `inci` that would normally be applied to the whole function body and the local bindings. This PR adds an `inci` around the local bindings, conditional on being in `Hanging` mode.

Closes #98

### DIFF
--- a/data/examples/declaration/value/function/do-where-out.hs
+++ b/data/examples/declaration/value/function/do-where-out.hs
@@ -1,0 +1,5 @@
+f :: Maybe Int
+f = do
+  return c
+  where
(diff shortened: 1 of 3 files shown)


domenkozar:
@mrkkrp this one is painful :)

Issue #52, opened 2019-05-26
Implement documentation and warning declarations


mrkkrp:
Extend `p_hsDecl` in `Ormolu.Printer.Meat.Declaration` in order to support `WarningD` and `DocD` declarations. I'm not sure what is the role of `DocD` and if it's at all necessary given that we render all comments automatically. Perhaps it could be skipped.

ruhatch:
I think `DocDecl`s will be removed in the future and only get parsed in `haddock` mode, so I'm not sure we need to support them.

mrkkrp:
This seems to be fully-implemented now.

PR #103, opened 2019-06-15
[#52] Implement WARNING/DEPRECATED in module header

ruhatch:
This adds `WARNING` and `DEPRECATED` pragmas to the module header.

It will normalise singleton lists to just a string, and it will normalise to a single line layout unless there's explicit multi-lining in the body of the pragma (i.e. not just from hanging brackets.

The `pragma` combinator can be used for other pragmas later on.

### DIFF
--- a/src/Ormolu/Printer/Meat/Module.hs
+++ b/src/Ormolu/Printer/Meat/Module.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PatternSynonyms   #-}
 {-# LANGUAGE RecordWildCards   #-}
@@ -9,8 +10,10 @@ module Ormolu.Printer.Meat.Module
   )
 where
 
+import BasicTypes hiding (InlinePragma)
 import Control.Monad
 import Data.Maybe (isJust)
+import Data.Text (Text)
 import GHC
 import Ormolu.Imports
 import Ormolu.Printer.Combinators
@@ -35,6 +38,7 @@ p_hsModule loc@(L moduleSpan hsModule) = do
       Nothing -> pure ()
       Just hsmodName' -> line $ do
         located hsmodName' p_hsmodName
+        maybe (pure ()) (located' p_warningTxt) hsmodDeprecMessage
         case hsmodExports of
           Nothing -> return ()
           Just hsmodExports' -> do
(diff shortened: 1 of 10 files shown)


ruhatch:
@mrkkrp Might we want to deal with a multiline case here, e.g.
```
module Test
  {-# DEPRECATED
        [ "This module is deprecated."
        , "Please use some other module instead."
        ]
  #-}
  ( foo
  , bar
  , baz
  )
where
```

ruhatch:
I went ahead and added something like this, so there's a couple of points to discuss. Firstly, should
```
module Test {-# WARNING "This module is very internal"
  #-} where
```
result in
```
module Test {-# WARNING "This module is very internal" #-} where
```
or 
```
module Test
  {-# WARNING
        "This module is very internal"
    #-} where
```
and secondly, what do you think about the three `inci`s to align strings with 2 spaces after `PRAGMA`?

src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -59,6 +63,26 @@ p_hsModule loc@(L moduleSpan hsModule) = do
...
+  WarningTxt _ lits -> p_pragma "WARNING" lits
+  DeprecatedTxt _ lits -> p_pragma "DEPRECATED" lits
+  where
+    p_pragma :: Text -> [Located StringLiteral] -> R ()
+    p_pragma pragmaText lits = switchLayout (litsSpan lits) $ do
+      breakpoint
+      inci $ pragma (litsSpan lits) pragmaText (p_lits lits)
+
+    litsSpan :: [Located StringLiteral] -> SrcSpan
+    litsSpan lits = combineSrcSpans (getLoc $ head lits) (getLoc $ last lits)
```
basile-henry: We might actually be doing this wrong. Maybe we should combine all the `SrcSpan` from the list. We should use:
https://github.com/tweag/ormolu/blob/master/src/Ormolu/Utils.hs#L19

ruhatch: Hmm, as long as we know the spans are in order, then it's more efficient to do it the way we're doing it now. But maybe we don't know that for sure :man_shrugging:.

@mrkkrp what do you think?

mrkkrp: I changed the code to use all spans in #101.

mrkkrp:
This is already merged as part of #101.

PR #106, opened 2019-06-15
Add CompleteMatchSig printer

basile-henry:
This is part of #70

This is based off of #103 in order to use the latest pragma combinator.


ruhatch:
@basile-henry you should be able to use the `pragma` combinator from #103 :)

PR #107, opened 2019-06-15
Expand all constructor that are missing (no Show instance)

domenkozar:
Give a user the chance to see what AST is not supported.


mrkkrp:
Thanks! I'm not sure why it failed as unauthorized for you and run for others. AFAIU it should still run from forks just without the `org-global` context.

PR #109, opened 2019-06-16
Add AnnD printer

basile-henry:
This is part of #70

This is based off of #103 in order to use the latest `pragma` combinator.


mrkkrp:
I don't think I've ever seen this ANN thing used in code. What is it used for?

yumiova:
An example of how it can be used is in disabling HLint linters:

```haskell
{-# ANN module "HLint: ignore Use string literal" #-}
main :: IO ()
main = putStrLn ['H', 'e', 'l', 'l', 'o', ',', ' ', 'w', 'o', 'r', 'l', 'd', '!']
```

PR #111, opened 2019-06-16
Support infix operator & multi-line class names

yumiova:
This pull request implements certain changes that were accidentally left out of #79. Specifically, it makes the handling of type variables & names for type class pretty printing reuse `p_infixDefHelper` to support infix operator names & parameters over multiple lines. In addition, minor cleanups and refactorings were left out of #79.

### DIFF
--- a/data/examples/declaration/class/associated-data.hs
+++ b/data/examples/declaration/class/associated-data.hs
@@ -14,7 +14,10 @@ class Bar a
 -- | Something more.
 class Baz a where
     -- | Baz bar
-    data BazBar a b c
+    data BazBar
+        a
+        b
+        c
 
     -- | Baz baz
     data family BazBaz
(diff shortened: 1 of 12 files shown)


data/examples/declaration/class/associated-data.hs
```diff
@@ -6,15 +6,19 @@ class Foo a where data FooBar a
 class Bar a
   where
     -- | Bar bar
-    data BarBar a
```
mrkkrp: IIRC I introduced one-line data families because the original examples didn't have that sort of example, only multiline. Why do you want to bring it back?

data/examples/declaration/class/associated-types-out.hs
```diff
@@ -5,7 +5,9 @@ class Foo a where
 -- | Something.
 class Bar a where
   -- | Bar bar
-  type BarBar a
+  type
```
mrkkrp: The same here, we need single line examples of this too.

yumiova:
Note that none of the changes applied here are new; they are all the ones that were lost during squashing of the previous branch.

yumiova:
OK, I have reverted some of the changes that I didn't realise you made earlier, since to me it seemed like they were part of the mess caused accidentally by the squashing in #79.

PR #116, opened 2019-06-17
[#94] Improve handling of newline before match group bodies

### LINKED ISSUE #94: Improve handling of newline before functions bodies
Currently we use a "normal" `breakpoint` between equal sign `=` of function definitions and body functions. This means that there will be a space if the body is a one-liner and a newline otherwise. This leads to this sort of transformation:

```haskell
-- input:
instance Stream BL.ByteString where
  reachOffset o pst =
    reachOffset' splitAtBL BL.foldl' BL8.unpack (chr . fromIntegral) (10, 9) o pst

-- output:
instance Stream BL.ByteString where
  reachOffset o pst = reachOffset' splitAtBL BL.foldl' BL8.unpack (chr . fromIntegral) (10, 9) o pst
```

Which is obviously wrong because now the line is too long. We should start respecting those newlines.

N.B. I guess this is applicable to any match groups, not only function definitions.

ruhatch:
The PR changes the body layouting to use a span that includes the last point in
the name/pattern part of a match. This forces the layout to be multiline iff the
body starts on a newline.

Closes #94

### DIFF
--- a/data/examples/declaration/class/default-implementations-out.hs
+++ b/data/examples/declaration/class/default-implementations-out.hs
@@ -13,7 +13,8 @@ class Bar a where
 -- | Baz
 class Baz a where
   foobar :: a -> a
-  foobar a = barbaz (bazbar a)
+  foobar a =
+    barbaz (bazbar a)
   -- | Bar baz
   barbaz
     :: a -> a
--- a/data/examples/declaration/class/default-signatures-out.hs
+++ b/data/examples/declaration/class/default-signatures-out.hs
@@ -25,4 +25,5 @@ class Bar a where
   -- Even more pointless comment
   bar
     a
-    b = read a <> read b
+    b =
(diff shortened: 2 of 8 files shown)


mrkkrp:
Thanks. I’ll review this one and all the others this week.

mrkkrp:
This still doesn't solve the problem completely. For example if I have this in a where clause:

```
   where
-    queryLocated (L mspn _) =
-      case cast mspn :: Maybe SrcSpan of
-        Nothing -> mempty
-        Just (UnhelpfulSpan _) -> mempty
-        Just (RealSrcSpan spn) -> D.singleton spn
+    queryLocated (L mspn _) = case cast mspn :: Maybe SrcSpan of
+      Nothing -> mempty
+      Just (UnhelpfulSpan _) -> mempty
```

PR #117, opened 2019-06-17
Add Splice printer

basile-henry:
This PR does the following:

- [x] Format typed and untyped splices as expression
- [x] Format TH quotes
- [x] Fix QuasiQuote (by not formatting the inner string at all)

I haven't been able to format the declaration TH quotes:

```
foo = [d|
  bar :: Char
  bar = 'c'
  |]
```

In order to format the inner `HsDecl`, we need access to `p_hsDecl` from `Declaration.Value`.
I have 2 potential solutions to fix this but I'd like @mrkkrp's point of view before I go ahead and implement one of the two solutions.

1. We inline the `Declaration.Value` module in `Declaration`. I've quickly tried this and it looks like it would basically force us to move all the modules `Declaration.*` into `Declaration` since they basically all use `Declaration.Value` one way or another.
2. We could try to do some _dependency injection_ of the `p_hsDecl` function and pass it as an argument to the functions that need it. I assume this would end up being needed in most places the `R` monad goes so it might be interesting to put the `p_hsDecl` into the `Reader` monad that `R` is.

This second option is probably the crazier (most unusual) one, but I expect it will be a smaller difference to the codebase.
Maybe you have some other ideas to solve this?

### DIFF
--- a/data/examples/declaration/splice/untyped-splice-out.hs
+++ b/data/examples/declaration/splice/untyped-splice-out.hs
@@ -1,0 +1,9 @@
+{-# LANGUAGE TemplateHaskell #-}
+x = $(foo bar)
+
+x =
+  $( foo
+       bar
+   )
+
+x = $foo
--- a/data/examples/declaration/splice/bracket-out.hs
+++ b/data/examples/declaration/splice/bracket-out.hs
@@ -1,0 +1,25 @@
+{-# LANGUAGE TemplateHaskell #-}
+foo =
+  [e|
+  foo bar
+  |]
+
+foo =
+  [e|
+  foo bar
+  |]
+
+foo = [t|Char|]
+
+foo =
+  [d|
+  foo :: Int -> Char
+
+  bar = 42
+
+  |]
+
+foo =
+  [||
+  foo bar
+  ||]
--- a/data/examples/declaration/splice/bracket.hs
+++ b/data/examples/declaration/splice/bracket.hs
@@ -1,0 +1,17 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+foo   = [|      foo bar
+  |]
+
+foo = [e| foo bar
+  |]
+
+foo = [t|       Char |]
+
+foo = [d|
+           foo:: Int       -> Char
+           bar = 42
+  |]
+
+foo = [|| foo bar
+  ||]
--- a/data/examples/declaration/splice/quasiquote-out.hs
+++ b/data/examples/declaration/splice/quasiquote-out.hs
@@ -1,0 +1,11 @@
+{-# LANGUAGE QuasiQuotes #-}
+x = [foo|    foo bar   |]
+
+x =
+  [e|    foo
+ bar  {- -}
+  |]
+
+[d|    foo bar
+
+|]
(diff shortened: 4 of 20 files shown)


data/examples/declaration/splice/untyped-splice-out.hs
```diff
@@ -0,0 +1,9 @@
+{-# LANGUAGE TemplateHaskell #-}
+x = $(foo bar)
+
+x =
+  $( foo
+     bar
```
basile-henry: This one seems weird to me. I think it should be something like this instead:

```haskell
x =
  $( foo
       bar
   )
```

What's inside the quote is simply a `HsExpr`, and I didn't want to modify that formatting logic in this PR. Am I using it wrong? Do I need to give more context about the indentation somehow?

mrkkrp:
I'll review this one and everything else tomorrow.

basile-henry:
@mrkkrp I have attempted the second option I gave, it is all in the latest commit. Tell me what you think about it!

It's a bit of a weird design, so if you want me to remove that commit from this PR, I completely understand! :smile: 

If you prefer going with the first solution (moving many `Declaration.*` modules into `Declaration`), I can try that instead!

ruhatch:
@basile-henry would it not work to pass `p_hsDecl` to `p_valDecl` as a function, like you suggested in Zurich? I'd rather it not be promoted into the wider context of the `R` monad if possible!

data/examples/declaration/splice/untyped-splice-out.hs
mrkkrp: This can be fixed by using `sitcc` I guess.

mrkkrp:
> We could try to do some dependency injection of the p_hsDecl function and pass it as an argument to the functions that need it. I assume this would end up being needed in most places the R monad goes so it might be interesting to put the p_hsDecl into the Reader monad that R is.

I'm going to look at that now. It's certainly not a good idea to put `p_hsDecl` into `R` because it does not belong there. `R` enables printing, but should not contain actual rendering functions, I find it confusing and agree with @ruhatch's comment above.

I'll review the PR now and force push any corrections that I'll find necessary. After that I'm going to merge it, so no action from you is needed.

mrkkrp:
Another consideration to keep in mind is that the set of primitives and internal combinators should be kept as minimal as possible.

basile-henry:
@ruhatch I have tried passing it explicitly as I first suggested. But `p_valDecl` is used in many places, so you would end up passing the `p_hsDecl` pretty much everywhere.
I can try to finish that version tonight so we can compare.
I don't feel strongly about this R version at all, I think it's the version of the change with the smallest diff, but that's definitely the wrong metric to optimise for. 😅

mrkkrp:
@basile-henry Have you actually confirmed that the string transformation we were doing changes the string that user gets in the end?

basile-henry:
@mrkkrp `.hs-boot` looks pretty good! 😄 I'd never tried it myself before, it doesn't look too bad.

For the QuasiQuote, I haven't found anything about GHC doing some preprocessing on indentation like you said it might. I haven't actually tested this myself, but the fact that some QuasiQuote packages provide helper functions to do just that, like the [unindent](http://hackage.haskell.org/package/interpolate-0.2.0/docs/Data-String-Interpolate-Util.html#v:unindent) function from the `interpolate` package, would suggest GHC provides the raw string as is to the quasiquoter.

basile-henry:
This whole `interpolate` package would be annoying to use if GHC did do some preprocessing! 😅

mrkkrp:
I just checked it myself and indeed the string should be printed back in its raw form.

Issue #47, opened 2019-05-26
Support rendering of remaining types of expressions


mrkkrp:
Some types of expressions are not implemented yet. For example `do`-notation. Extend the `p_hsExpr` function in `Ormolu.Printer.Meat.Declaration.Value` to support all of them.

yumiova:
Pretty printing of expressions is something I have been working on in [yumiova:yumiova/misc-expression](https://github.com/yumiova/ormolu/tree/yumiova/misc-expression). I forgot to mention this up until now.

Basically, I am more or less done with _most_ types of expressions. Some expressions need some more thought; I will be sending a pull request with information about this soon.

ruhatch:
@yumiova are you taking on `do`-notation yet, or can I claim that and track it in another issue?

mrkkrp:
There is an attempt to implement it in #60, but that will require some adjustments.

yumiova:
@ruhatch After seeing the work of #60 in supporting do notation, I intended to wait and see what happens. Once do notation is figured out, I would like to explore implementing list comprehensions and arrow notation. I have no intention of implementing Template Haskell splices and brackets.

ruhatch:
Ah, okay! Might there be a simpler way to get a birds-eye view of what's been covered? Maybe something in the readme or wiki that links the issues?

mrkkrp:
Maybe we could have an umbrella issue with a list of checkboxes. When the entire AST is covered the issue can be closed. I'll probably create it this evening.

mrkkrp:
Now that I merged #60 what remains here is various not-yet-covered types of do notation (includes arrow notation, comprehensions, etc.) and perhaps something in the `p_hsExpr`.

yumiova:
My work on pretty printing list comprehensions and arrow notation will have to be delayed just a little bit. Mainly because I am working on #79, as well as upcoming exams I need to practice for. Consequently, I will only be able to work on list comprehensions around the coming friday/saturday. If anyone needs list comprehensions and/or arrow notation before then, the prompt is open. Otherwise, I will definitely get it done by saturday.

yumiova:
Pretty printing of parts of list comprehensions will finally get implemented with #119. This does not, however, include transform list comprehensions, largely because dealing with that well is much trickier. See #119 for more information. Unless someone else does it or someone explicitly presents a case where they have use for it, I will probably wait for _a long time_ before implementing transform list comprehensions, as it otherwise really does not seem important to me.

[...]

PR #119, opened 2019-06-17
Implement pretty printing of basic & parallel list comprehensions

yumiova:
This pull requests implements some parts of list comprehensions as specified in #47. Specifically, it implements normal & parallel list comprehensions. However, it does not implement transform list comprehensions.

There are no real problems to speak of, as the formatting is obvious and consistent with what the Haskell community already does. See the tests for an example.

The reason transform list comprehensions are not implemented yet is because the GHC AST for representing list comprehensions is a horrible mess, where slightly different list comprehensions end up causing massively different syntax trees. The simplest way of solving this would likely be a preprocessing step that reads the syntax trees and produces a simple, linear list of bindings. I decided not to implement this yet to avoid delaying this commit further.

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -14,7 +14,7 @@ import Bag (bagToList)
 import BasicTypes
 import Control.Monad
 import Data.Data
-import Data.List (sortOn)
+import Data.List (intersperse, sortOn)
 import Data.Text (Text)
 import GHC
 import Ormolu.Printer.Combinators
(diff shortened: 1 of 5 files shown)


src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -186,8 +186,7 @@ p_grhs _ (XGRHS NoExt) = notImplemented "XGRHS"
 
 p_stmt :: Stmt GhcPs (LHsExpr GhcPs) -> R ()
 p_stmt = \case
-  LastStmt NoExt _ _ _ ->
-    notImplemented "LastStmt" -- only available after renamer
+  LastStmt NoExt body _ _ -> located body p_hsExpr
```
mrkkrp: What is the reason you added this? AFAIK `LastStmt` is only available after renamer and so in our cases it'll never be passed.

yumiova: The return value (e.g. `(a, b)` in `[(a, b) | a <- x, b <- y]`) seems to always be `LastStmt`.

yumiova: At the very least, that is what all of my experimentation seems to indicate.

yumiova: If you want to see this fact in action, you can remove the `LastStmt` case.

Issue #110, opened 2019-06-16
More precise handling of newlines between methods inside of a type class declaration


mrkkrp:
Type classes were implemented in #79, but there is still an issue related to white space between methods and data/type family declarations inside of class declarations. I think those should be grouped similar to what we do for top-level functions:

```haskell
class Foo a where
  
  -- | Here goes the foo.
  foo : a -> String
  foo = default_implementation

  -- | Here goes bar.
  bar :: a -> String
  bar = ...
```

Note the newline between `foo` and `bar` and lack of thereof between e.g. type signature of `foo` and its default implementation.

Right now I think everything is joined without newlines, as one block of text.

yumiova:
Indeed, everything is formatted without newlines. I did it that way since no standards had been established yet, and previous Haskell formatters all seemed to use this without-newlines style.

mrkkrp:
@yumiova Feel free to grab this one!

yumiova:
I can implement this change. Should the first line be redundant as shown in the example?

yumiova:
OK, this may take a little bit longer than expected. Somehow, I managed to forget that certain declarations need to be bound together. It's not a particularly difficult task, but finding how to integrate it without completely tearing apart the codebase requires some thought on my part.

yumiova:
I figured out how to implement this without tearing apart the codebase, so this gets implemented in #120.

mrkkrp:
This appears to be resolved now.

PR #120, opened 2019-06-17
Add newlines between most declarations in a type class

yumiova:
This pull request adds newlines between most declarations in a type class, as specified in #110. Certain other kinds of declarations (such as associated type signatures and bindings) are grouped together without newlines. The process for determining which declarations are grouped reuses the same logic as for modules.

Sorting of type class declarations now wraps everything up in a `HsDecl` instead of pre-pretty-printing everything ahead of time. Since the way in which declarations inside of type classes are pretty printed may vary (e.g. type families), a specialized pretty printer called `p_associatedDecl` is partially copied from `p_hsDecl`. This is a minor bit of code duplication. However, since the behavior is relatively different and constrained, I believe this minor code duplication is appropriate. Alternatively, I could have passed a context/style parameter into `p_hsDecl`. Ping me if this seems preferable.

The `p_hsDecls` pretty printer is reused for handling newlines between declarations. In order to reuse `p_hsDecls` with `p_associatedDecl`, I generalised the pretty printer to accept any declaration pretty printer and called it `p_hsDecls'`. Alternatively, again, `p_hsDecls` could accept a context/style parameter.

Otherwise, all cases where `separatedDecls` matches `TypeSig` will now also match `ClassOpSig`, since that is how all type signatures are represented in type classes.

### DIFF
--- a/src/Ormolu/Printer/Meat/Common.hs
+++ b/src/Ormolu/Printer/Meat/Common.hs
@@ -27,8 +27,8 @@ import RdrName (RdrName (..))
 -- | Data and type family style.
 
 data FamilyStyle
-  = Associated
-  | Free
+  = Associated                  -- ^ Declarations in type classes
+  | Free                        -- ^ Top-level declarations
 
 p_hsmodName :: ModuleName -> R ()
 p_hsmodName mname = do
--- a/src/Ormolu/Printer/Meat/Declaration/Class.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Class.hs
@@ -13,18 +13,16 @@ import Class
 import Control.Arrow
 import Control.Monad
 import Data.Foldable
-import Data.Function
 import Data.List (sortBy)
+import Data.Ord (comparing)
+import GHC
 import Ormolu.Printer.Combinators
-import Ormolu.Printer.Meat.Declaration.Signature
-import Ormolu.Printer.Meat.Declaration.Value
-import Ormolu.Printer.Meat.Declaration.TypeFamily
 import Ormolu.Printer.Meat.Common
 import Ormolu.Printer.Meat.Type
 import Ormolu.Utils
-import GHC
 import RdrName (RdrName (..))
 import SrcLoc (Located, combineSrcSpans)
+import {-# SOURCE #-} Ormolu.Printer.Meat.Declaration
 
 p_classDecl
   :: LHsContext GhcPs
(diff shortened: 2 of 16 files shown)


src/Ormolu/Printer/Meat/Common.hs
```diff
@@ -89,6 +90,25 @@ p_ieWildcard = \case
...
+-- | A helper for pretty printing groups of certain syntactical elements.
+-- Groups are printed without any newlines in-between, whereas separate groups
+-- have an additional newline for clarity.
+
+p_groupingHelper
+  :: (a -> R ())                -- ^ How to format elements
+  -> (a -> a -> Bool)           -- ^ Whether two declarations should be grouped
+  -> [a]                        -- ^ List of syntactic elements
+  -> R ()
+p_groupingHelper pretty group xs =
```
ruhatch: It seems that `group` is a misnomer here, it is actually a predicate that says if they should be separated!

src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -46,38 +45,11 @@ p_hsModule loc@(L moduleSpan hsModule) = do
...
-      \(d, md) -> do
-        case md of
-          Nothing -> located d p_hsDecl
-          Just d' ->
-            if separatedDecls (unL d) (unL d')
-              then line (located d p_hsDecl)
-              else located d p_hsDecl
+    p_groupingHelper
+      (located' p_hsDecl)
+      (\d d' -> separatedDecls (unL d) (unL d'))
```
ruhatch: It seems like you only ever call this function with `separatedDecls`, so maybe you should inline it to `p_groupingHelper` for now. Once there's an actual use case for generalising the grouping predicate then you can resurrect that logic.

src/Ormolu/Printer/Meat/Declaration/Class.hs
```diff
@@ -61,34 +62,45 @@ p_classDecl ctx name tvars fdeps csigs cdefs cats catdefs = do
...
-      inci (sequence_ decls)
-    else newline
+  let sigs = (getLoc &&& fmap (SigD NoExt)) <$> csigs
+      vals = (getLoc &&& fmap (ValD NoExt)) <$> toList cdefs
+      tyFams = (getLoc &&& fmap (TyClD NoExt . FamDecl NoExt)) <$> cats
+      tyFamDefs =
+        ( getLoc &&& fmap (InstD NoExt . TyFamInstD NoExt . defltEqnToInstDecl)
+        ) <$> catdefs
+      allDecls =
+        snd <$> sortBy (compare `on` fst) (sigs <> vals <> tyFams <> tyFamDefs)
```
ruhatch:
```
(compare `on` fst) == comparing fst
```

src/Ormolu/Printer/Meat/Declaration/Class.hs
```diff
@@ -61,34 +62,45 @@ p_classDecl ctx name tvars fdeps csigs cdefs cats catdefs = do
...
-p_famDefDecl FamEqn {..} = do
-  txt "type"
-  breakpoint
-  let eqn = FamEqn { feqn_pats = tyVarsToTypes feqn_pats, .. }
-      hsib = HsIB { hsib_ext = NoExt, hsib_body = eqn }
-  inci (p_tyFamInstEqn hsib)
-  newline
-p_famDefDecl XFamEqn {} = notImplemented "XFamEqn"
+p_associatedDecl :: HsDecl GhcPs -> R ()
+p_associatedDecl = \case
```
ruhatch: This duplication does seem reasonable to me

yumiova:
I am currently making some changes in order to adapt the codebase to recent changes.

yumiova:
OK, this pull request is now integrated with `p_hsDecls`. I have rewritten the pull request description to reflect the new structure.

mrkkrp:
I think there should be a line before first declaration in class. What about instance declarations? I imagine there should be grouping there as well.

yumiova:
I have literally never seen anyone add a line before the first declaration, but fair enough. Should there be a line if there is only one declaration?

As for the instance declarations: I am thinking that can be done in another pull request.

mrkkrp:
> I am thinking that can be done in another pull request.

yumiova:
OK, I updated all the tests and now generate a newline before the first declaration.

yumiova:
In hindsight, maybe passing in some kind of style parameter to `p_hsDecl` instead is the right solution after all? Once grouping of instance declarations is handled, I'll presumably need to copy something similar to `p_associatedDecl` again. Perhaps it is a lot easier to just define handling of associated declarations together with normal declarations. I could use a trick similar to `p_hsDecls'` (i.e. only pass style into function called `p_hsDecl'`, where `p_hsDecl` is specialised to the usual style) in order to avoid changing downstream code.

We could either merge this now, and eventually make such a change later (*if it is needed*), or I could just do this now and be done in about an hour or so.

mrkkrp:
I'll return to this one later today or tomorrow. I merged the small ones first because it's easier.

PR #125, opened 2019-06-20
[#52] Implement printer for WarningD

ruhatch:


data/examples/declaration/warning/warning-multiline-out.hs
```diff
@@ -0,0 +1,9 @@
+{-# WARNING
+    test
+  , foo
+  [ "These are bad functions"
```
basile-henry: Shouldn't this list stay single line?

data/examples/declaration/warning/warning-single-line-out.hs
```diff
@@ -0,0 +1,4 @@
+{-# WARNING test "This is a warning" #-}
```
basile-henry: You transform singleton list into a string? Is this the concern of a formatter?

src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -39,7 +35,7 @@ p_hsModule loc@(L moduleSpan hsModule) = do
       Nothing -> pure ()
       Just hsmodName' -> line $ do
         located hsmodName' p_hsmodName
-        forM_ hsmodDeprecMessage (located' p_warningTxt)
+        forM_ hsmodDeprecMessage (located' (p_warningTxt ModuleWarning))
```
basile-henry: Since you seem to only be using `p_warningTxt` and `ModuleWarning` here (outside of the `Declaration.Warning` module), I'd be tempted to expose a single `p_warningModule` function instead. But it's not really an issue either way :sweat_smile:

data/examples/declaration/warning/warning-single-line-out.hs
mrkkrp: It appears that the inner representation of the two is the same.

src/Ormolu/Printer/Meat/Declaration/Warning.hs
```diff
@@ -0,0 +1,73 @@
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Ormolu.Printer.Meat.Declaration.Warning
+  ( p_warnDecls
+  , p_warningTxt
+  , WarningType(..)
```
mrkkrp: I think we should not leak this out of this module. Factually the outside world is interested in two functions: 1) `p_warnDecls` 2) `p_warningTxt ModuleText`. That's exactly what should be exported from this module.

I'll adjust it a bit later.

PR #131, opened 2019-06-26
Implement all facets of arrow notation

yumiova:
This pull request implements all facets of arrow notation (as far as I know), as specified in #47. Most syntactic constructs reuse pretty printing specified previously, such as arrow cases, if-expressions and lambdas. Relevant operators (such as arrow argument passing) and brackets are also formatted as operators and brackets typically are.

Note that supporting arrow notation necessitates a need to generalise `p_stmt` and `p_matchGroup`, since arrow notation *literally* reuses these syntactic constructs, except so-called "arrow commands" are used instead of expressions. In theory, all these functions could be duplicated and specialised for arrows, though that seems undesirable.

Relevant expressions (e.g. proc notation) will hang properly, in a similar way to case expressions and do notation, using the same logic.

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -13,7 +13,7 @@ where
 import Bag (bagToList)
 import BasicTypes
 import Control.Monad
-import Data.Data
+import Data.Data hiding (Infix, Prefix)
 import Data.List (intersperse, sortOn)
 import Data.Text (Text)
 import GHC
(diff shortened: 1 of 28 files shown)


src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -181,28 +209,101 @@ p_grhs style (GRHS NoExt guards body) =
...
+    unless (null cmds) $ do
+      breakpoint
+      inci (sequence_ (intersperse breakpoint (located' p_hsCmdTop <$> cmds)))
+  HsCmdArrForm NoExt form Infix _ [left, right] -> do
+    located left p_hsCmdTop
+    space
+    located form p_hsExpr
+    placeHanging (cmdTopPlacement (unLoc right)) $
+      located right p_hsCmdTop
+  HsCmdArrForm NoExt _ Infix _ _ -> notImplemented "HsCmdArrForm"
```
yumiova: `Infix` should never match if there are more or less than two arguments, hence the `notImplemented`.

mrkkrp:
Does this implement all remaining expression types or there is still something missing?

yumiova:
It *seems* like it, except for maybe the `Tick` family of expressions, which I have no idea what is.

yumiova:
Also, transform list comprehensions are not supported yet.

PR #140, opened 2019-06-29
Recognize gaps between comment blocks

### LINKED ISSUE #138: Undesirable re-alignment of comments
Input:

```haskell
class Foo a where

  -- | Foo.

  foo
    :: Int
    -> a -- ^ Something

  -- | Bar.

  bar :: a
```

output:

```haskell
class Foo a where

  -- | Foo.
  foo
    :: Int
    -> a -- ^ Something

    -- | Bar.

  bar :: a
```

Relatively easy to fix.

basile-henry:
Fixes #138

This PR adds a check for Haddock comments `-- |` so that they don't count as logically following.

Do you think I should add some checks for multiline comments starting with `{- |` as well?

### DIFF
--- a/data/examples/other/comment-alignment-out.hs
+++ b/data/examples/other/comment-alignment-out.hs
@@ -1,0 +1,9 @@
+class Foo a where
+
+  -- | Foo.
+  foo
+    :: Int
+    -> a -- ^ Something
+
+  -- | Bar.
+  bar :: a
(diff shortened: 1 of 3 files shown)


mrkkrp:
This is also what I thought of at the beginning. However, it might be a better idea to check for gaps (that is, newlines breaking a block of comments) and use that to tell when one block ends and another block starts:

```haskell
-- ^ One
-- block

-- | Another
-- block
```

Or maybe both. Today I'm not working on Ormolu, will return to this later.

PR #147, opened 2019-07-01
Sort language pragmas

### LINKED ISSUE #20: Implement automatic sorting of language pragmas
Not to forget. I'll skip this in current PR about comments, but eventually such a feature would be good.

waddlaw:
Close #20.

### DIFF
--- a/data/examples/declaration/class/associated-data-out.hs
+++ b/data/examples/declaration/class/associated-data-out.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE TypeFamilies #-}
+
 class Foo a where
 
   data FooBar a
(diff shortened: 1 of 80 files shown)


mrkkrp:
We're trying to figure out why CI fails for some users. Could you log in into CircleCI and try to click the "restart job" button there and see what happens?

waddlaw:
"restart job" button does not appear. "An unauthorized user error" has occurred.

ci

aspiwack:
Thanks for the screenshot: it's very helpful! We'll try to figure out how to fix this.

mrkkrp:
Apparently for some users CI works, see for example #150 which worked right away. We're puzzled by the behavior. To move forward with the PR, I'm going to restart the build myself.

mrkkrp:
I'm going to amend this work in order to implement everything mentioned in #20 because it makes sense to handle everything now instead of going back and forth with this.

mrkkrp:
I couldn't finish this this evening, but the idea here is to extract collection of extensions in a typed way and drop them from raw comments. Then we can just pretty-print the extensions so they have 1) fixed placement relative to other constructs such as module header 2) always totally normalized.

Will return to this when I have time.

PR #149, opened 2019-07-02
Update cabal file

mrkkrp:
One nice thing is that beginning from Cabal 2.4 we can use ** wildcard in paths, which is especially useful in our case because we have many examples in a big directory tree, so without ** listing all the directories correctly is a real pain.

All other changes are just to satisfy Cabal's nitpicks.

### DIFF
--- a/ormolu.cabal
+++ b/ormolu.cabal
@@ -15,12 +15,36 @@ extra-doc-files:      CONTRIBUTING.md
                     , CHANGELOG.md
                     , DESIGN.md
                     , README.md
-data-files:           data/examples/declaration/data/*.hs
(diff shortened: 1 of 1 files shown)


mrkkrp:
The failure appears to be a bug in nixpkgs.

mrkkrp:
Status of this one: pending a release with this fix https://github.com/haskell/cabal/issues/6125.

mrkkrp:
Since we cannot get new version of Cabal with the fix, we have to make sure that the package will work with older Cabal. Practically, this means that `data-files` should list all sub-directories because we cannot use `**` wildcard yet. The best way forward here is to list the directory tree (without listing individual files) and then insert that verbatim into the Cabal file.

PR #150, opened 2019-07-04
Allow reading from stdin

tuncer:


mrkkrp:
Now looking at this I'm somewhat reluctant because:

* Even if we merge it, I think the behavior should be "if no input file provided, read from stdin". The special handling of `-` filename seems ugly.
* Some modes of operation do not work with stdin. For example in place editing, as well as checking.
* I think Brittany does not support reading from stdin.

What do others think?

tuncer:
>Even if we merge it, I think the behavior should be "if no input file provided, read from stdin". The special handling of `-` filename seems ugly.

To not break the API, I avoided changing `ormoluFile`'s signature to `Maybe FilePath`, but I can do that if it's ok. In that case, what path should be reported in parse error messages when it's `Nothing`?

> Some modes of operation do not work with stdin. For example in place editing, as well as checking.

With stdin, there is no file to inplace-edit, and we're already inplace-editing stdin to stdout. Inplace and Check modes are undefined when stdin is used, so we're exiting with failure. The reason we don't print an error message is that not all formatter integrations are smart enough to only replace the original code when exit code is 0. Those that do it wrong will replace your original code with the error message. Therefore, I opted to exit without a message. While this omits potentially helpful error explanations, we can document incompatibilities in the usage string.

> I think Brittany does not support reading from stdin.

Brittany does read from stdin, or it wouldn't work with neoformat.vim and format-all.el. See https://github.com/lspitzner/brittany/blob/460bd4dd2b14c11b171de99e61e3d778bb523604/src-brittany/Main.hs#L252

mboes:
Supporting `stdin` makes Ormolu more composable. Think about how much more annoying `cat` would be to use if you had to always give it a filename for the input. Also, `-` as a special name meaning "read from stdin" is a standard UNIX convention.

PR #153, opened 2019-07-04
Keep space in lambdas with lazyness annotation

### LINKED ISSUE #144: Syntax error with lambdas with laziness annotation
Similar to https://github.com/lspitzner/brittany/issues/228:

```haskell
a = \ ~x -> x
```

gets reformatted into

```haskell
a = \~x -> x
```

Which is not valid Haskell. This is an easy one.

dkellner:
Fixes #144

It will probably take more time for you to review than doing it yourself, but I'd appreciate any feedback if this is not the fix you intended.


LeonardoFurtado:
When solutions are too simple to implement, they are sometimes marked Good-First-Issues as they attract new contributors to the project.

Issue #163, opened 2019-07-07
Remove language extension whitelist


mboes:
In an effort to resolve #62 and other issues, #157 introduced an explicit list of extensions to turn on by default. This list is hard-coded, and moreover can't be configured, since that's what #157 also does: remove configurability (and rightly so). But if we're going to remove configurability, we should be extremely permissive about the inputs we allow.

As argued in https://github.com/tweag/ormolu/issues/62#issuecomment-497339902, we shouldn't need to whitelist extensions. Even keyword-stealing extensions should be fine, because it's not like we intend to format AST's differently based on whether some identifier is actually a keyword or not. Or at least - I can't think of an extension for which that would be true.

mrkkrp:
We certainly can't enable all of them at once. Examples that I discovered today include:

* `ViewPatterns` and `PatternSynonyms` do conflict because the syntax for view patterns is in conflict with the syntax for unidirectional pattern synonyms.
* Magic hash in some cases has a weird effect on `{-# things like this #-}`.

There may be more. I'm not afraid of letting `proc` or something like this to be stolen really, but those conflicts above just break the tool.

mboes:
Yes but managing a small set of conflicting extensions is much easier than managing a set of those that don't. The GHC Proposals Committee is on record for being against conflicting extensions, so we likely won't be seing many more of those in the future.

`-XMagicHash` is one of those few ones that shouldn't be enabled by default, because the format for `f# 1` will be very different depending on whether that occurrence of `#` stands for an operator or not.

mrkkrp:
So to make it clear: you want a blacklist instead of whitelist?

mboes:
Yes. The vast majority of extensions can safely be turned on without conflict, vs a tiny subset that does conflict. So better to blacklist than to whitelist.

mboes:
BTW do you have an example of conflict between `-XPatternSynonyms` and `-XViewPatterns`? I could not reproduce.

mrkkrp:
I'll show you tomorrow.

mrkkrp:
Actually I started digging that pattern synonyms stuff and found it to be a bug in Ormolu. The fix is coming.

mrkkrp:
Note: when done with this, add something to the docs and readme about which extensions are enabled by default.

waddlaw:
How about adding an option to print out a list of language extensions that are enabled by default?

mrkkrp:
It looks like a good idea to me.

PR #164, opened 2019-07-08
Update readme

waddlaw:
- Removed description of yaml configuration file.
- Added a list of extensions enabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,18 +64,6 @@ formatted output.
 $ ormolu --mode inplace Module.hs
 ```
 
-A configuration file can specify the language extensions to use by default.
-
-```console
-$ cat ormolu.yaml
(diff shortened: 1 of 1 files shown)


waddlaw:
Is the context setting affecting the CI not working?
https://github.com/tweag/ormolu/blob/master/.circleci/config.yml#L44

- [Using Contexts](https://circleci.com/docs/2.0/contexts/)

mrkkrp:
Thanks!

Explicitly enumerating enabled extensions is not a good idea because the list will tend to go out of sync with reality. Also see the discussion in #163. Once that issue is fixed, I'll add something about the extensions enabled by default, until then let's not add anything to the docs.

I'll merge the other part of the PR though.

PR #208, opened 2019-07-11
Comma-separated SPECIALIZE types

### LINKED ISSUE #206: SPECIALIZE pragma is not lossless
Before `ormolu`:

```
{-# SPECIALIZE fits13Bits :: Int -> Bool, Integer -> Bool #-}
```

After `ormolu`:

```
{-# SPECIALIZE fits13Bits :: Int -> Bool #-}
```

This is responsible for one of the failures in #200

int-index:
Fix #206

### DIFF
--- a/data/examples/declaration/signature/specialize/specialize-out.hs
+++ b/data/examples/declaration/signature/specialize/specialize-out.hs
@@ -11,5 +11,14 @@ baz :: Num a => a -> a
 baz = id
 {-# SPECIALIZE [~2] baz
   :: Int
-  -> Int
+     -> Int
   #-}
+
+{-# SPECIALIZE fits13Bits :: Int -> Bool, Integer -> Bool #-}
+{-# SPECIALIZE fits13Bits
+  :: Int
+     -> Bool
+  , Integer -> Bool
+  #-}
+fits13Bits :: Integral a => a -> Bool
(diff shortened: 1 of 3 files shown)


data/examples/declaration/signature/specialize/specialize-out.hs
```diff
@@ -11,5 +11,9 @@ baz :: Num a => a -> a
 baz = id
 {-# SPECIALIZE [~2] baz
   :: Int
-  -> Int
+     -> Int
```
int-index: I'm not sure why this happened. Suggestions?

basile-henry: I think the interaction of `velt` and `withSep` will align `::` with the comma (whether they exist or not). Here is an example with multi-line with comma so you can see:

```haskell
       {-# SPECIALIZE fits13Bits
         :: Int -> Bool
         , Integer -> Bool
         #-}
```

I'm not sure there's much we can do about aligning with a single multi-line type (no comma). But when we have multi-line comma, maybe we could add an extra space after the comma?

mrkkrp: This happens because `velt` wraps every "item" is `sitcc`. This is OK.

PR #213, opened 2019-07-13
Run on Hackage

### LINKED ISSUE #160: Testing on code from Hackage
Devise a method to run Ormolu on all or most packages found on Hackage. For that:

1. Download Hackage index.
2. Go through packages.
3. Select just source files.
4. Use standalone C preprocessor command to drop any CPP.
5. Run every file through Ormolu automatically logging problems.

mrkkrp:
Close #160.

TODO:

- [x] Filter out the bash script and nix files from Ormoul sources because their editing causes re-compilation of Ormolu. That's annoying. Perhaps we should just use a plain regexp instead of that gitignore magic.
- [x] Better organize the new nix code.
- [x] Improve error reporting in Ormolu. Any error reported by Ormolu should at least say which file is problematic. Also, do not show unhelpful spans in diff reporting, that's of no use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,9 @@ $ nix-build -A hackage.<package>
 ```
 
 Then inspect `result/log.txt` for possible problems. The derivation will
-also contain formatted files for inspection.
+also contain formatted `.hs` files for inspection and original inputs with
+`.hs-original` extension (those are with CPP dropped, exactly what is fed
+into Ormolu).
 
 ## Contributing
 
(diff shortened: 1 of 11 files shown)


mrkkrp:
It appears that dropping CPP is not that easy. I tried two standalone preprocessors: `gpp` and `mcpp`. Both chokes on Haskell with CPP :-)

neongreen:
Try cpphs.

PR #223, opened 2019-07-14
Handle multi-line case patterns

### LINKED ISSUE #198: Broken layout for multiline patterns in case matches
This one is from pandoc (severely reduced).

Input:

```haskell
readerBench doc name =
  runPure $ case (getReader name, getWriter name) of
    (Right (TextReader r, rexts),
     Right (TextWriter w, wexts))  -> undefined
```

Output:

```haskell
readerBench doc name =
  runPure $ case (getReader name, getWriter name) of
    ( Right (TextReader r, rexts)
    , Right (TextWriter w, wexts)
    )  -> undefined
```

```
4:5: error: …
```

basile-henry:
This PR closes #198

It turns out there are some weird parsing rules for patterns in `case` statements specifically.
From what I could figure out, tuples, lists and even infix constructors like `:` needs to have a symbol/operator on each line of the multi-line pattern match (but this is only necessary at the top level pattern, nested patterns can do the same as anywhere else).
I am not too what what the rationals for GHC is to have a special case here. Maybe something to make patterns unambiguous? I'll have to ask around.
In the meantime, I believe this PR fixes all these case pattern issues while keeping a style more or less consistent with what we already have.

### DIFF
--- a/data/examples/declaration/value/function/pattern/multiline-case-pattern-out.hs
+++ b/data/examples/declaration/value/function/pattern/multiline-case-pattern-out.hs
@@ -1,0 +1,12 @@
+readerBench doc name =
+  runPure $ case (getReader name, getWriter name) of
+    ( Right (TextReader r, rexts),
+      Right (TextWriter w, wexts)) -> undefined
+
+f xs = case xs of
+  [ a,
+    b] -> a + b
+
+g xs = case xs of
+  ( a :
+      bs) -> a + b
--- a/data/examples/declaration/value/function/pattern/multiline-case-pattern.hs
+++ b/data/examples/declaration/value/function/pattern/multiline-case-pattern.hs
@@ -1,0 +1,12 @@
+readerBench doc name =
+  runPure $ case (getReader name, getWriter name) of
+    (Right (TextReader r, rexts),
+     Right (TextWriter w, wexts)) -> undefined
+
+f xs = case xs of
+  [    a,
+   b       ] -> a + b
+
+g xs = case xs of
+  (a:
+    bs) -> a + b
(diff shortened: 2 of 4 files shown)


basile-henry:
Ok, I think I got the reason the multi-line case pattern a bit wrong.

GHC can parse this:

```haskell
h xs =
  case xs of
    [ a
      , b
      ]
      -> a + b
```

But not this:
```haskell

h xs =
  case xs of
    [ a
    , b
    ]
    -> a + b
```


Do we prefer a solution more like this than what I did in this PR?
That might be a lighter solution. Tell me what you think!
I'll work on something implementing that when I can! :smile:

basile-henry:
This PR seems to also fix a related issue that I believe hasn't been reported yet with mult-line pattern matches outside function definition (like top level pattern matches):

```
coffee [nix-shell] ormolu git:(master) ✗
$ cat Test.hs

[ a
 , b
 , c
 ] = [1, 2, 3]

coffee [nix-shell] ormolu git:(master) ✗
$ cabal new-run -- ormolu Test.hs
Up to date
  Test.hs<rendered>:2:1


coffee [nix-shell] ormolu git:(master) ✗
$ cabal new-run -- ormolu Test.hs -u
Up to date
[ a
, b
, c
]  = [1, 2, 3]
```

mrkkrp:
Thank you for the investigation. This one is harder. I want to think about it a little.

basile-henry:
Funny thing about that last example. Even though this is not valid Haskell syntax:

```haskell
[ a
, b
, c
]  = [1, 2, 3]
```

This is (using curly braces around the declaration, even at the top level):

```haskell
{

[ a
, b
, c
]  = [1, 2, 3]

}
```

:sweat_smile:

*Note*: Same thing for multiline pattern matches in `let` and `case` using curly braces and semicolons.

mrkkrp:
Aha! So we have a solution now. Just put braces around everything!

PR #226, opened 2019-07-15
Improve nix-build instructions

mpickering:
This allows you to test ormolu without cloning the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ It's possible to try Ormolu on arbitrary packages from Hackage. For that
 execute:
 
 ```console
-$ nix-build -A hackage.<package>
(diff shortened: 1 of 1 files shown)


mrkkrp:
True, but if I have it locally, I could do without the URL. Maybe we could just mention the trick with URL somewhere?

mrkkrp:
The question here is perhaps *who is going to use this feature*?

1. If it's for the developers, they have the repo cloned, and maybe even with some modifications. So they should be using the command without URL.
2. If it's just for curious people, then maybe they could add the URL.

The thing is, I belong to (1) and I'm not sure that (2) happens often enough to justify the addition of the URL.

mpickering:
I added it because my office mate tried the instructions but they didn't work because he hasn't clone the repo. If you don't feel like it's worth it then I will just close this PR.

PR #228, opened 2019-07-15
Fix Record Pattern Synonyms

### LINKED ISSUE #191: Record field punning in pattern synonyms is broken
Input:

```haskell
{-# LANGUAGE PatternSynonyms #-}

pattern ReqNoProv{j} = Just j

main = print ReqNoProv{j = False}
```

Output:

```haskell
{-# LANGUAGE PatternSynonyms #-}

pattern ReqNoProv j = Just j

main = print ReqNoProv {j = False}
```

Compiling the output will fail with `Not in scope: ‘j’`.

mauriciofierrom:
This fixes #191 about curly braces being removed when using record pattern synonyms.

### DIFF
--- a/data/examples/declaration/value/pattern-synonyms/bidirectional-out.hs
+++ b/data/examples/declaration/value/pattern-synonyms/bidirectional-out.hs
@@ -1,9 +1,23 @@
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE PatternSynonyms #-}
 
 pattern Arrow t1 t2 = App "->" [t1, t2]
 
+pattern Arrow {t1, t2} = App "->" [t1, t2]
+
+pattern Arrow
+  { t1,
+    t2
+    } =
+  App "->" [t1, t2]
+
 pattern Int =
   App "Int" []
+
+pattern Maybe {t} =
+  App
+    "Maybe"
+    [t]
 
 pattern Maybe t =
   App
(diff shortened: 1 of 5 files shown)


mrkkrp:
Is this ready for review?

mauriciofierrom:
Not yet. The following example:

```Haskell
pattern FirstTwo'{x,y}
  <- x:(y:xs)
```

Produces:

```haskell
pattern FirstTwo' {x
                 , y } <-
  x : (y : xs)
```

I was considering either not supporting multiline layout for this particular case or asking which should the proper formatting be in this case. WDYT?

mrkkrp:
Ideally layout of pattern part in input should control layout of pattern part in output. That is:

```haskell
pattern FirstTwo'{x,y}
  <- x:(y:xs)
```

should result in:

```haskell
pattern FirstTwo' {x, y}
  <- x:(y:xs)
```

whereas e.g.

```haskell
pattern FirstTwo'{ x
                 ,y}
  <- x:(y:xs)
```

should format the names inside braces in multiline fashion. Putting a break between name of the pattern and `{ ... }` part may be a good idea as well.

src/Ormolu/Printer/Combinators.hs
```diff
@@ -245,6 +246,13 @@ braces m = sitcc $ do
   ospaces m
   txt "}"
 
+-- | Surround given entity by parallel curly braces @{:@ and @:}@.
+bracesPar :: R () -> R ()
```
mrkkrp: Why is this called `bracesPar` and says in the documentation that it surrounds `m` with `{:` and `:}` while it clearly just uses `{` and `}`. AFAIU that you also want normal braces for what you're trying to fix in this PR.

mauriciofierrom: That's an error. I'll fix it now. Thanks!

mauriciofierrom:
This is ready for review.

PR #231, opened 2019-07-20
[#184] Handle promoted types

### LINKED ISSUE #184: Spaces lost in '[ 'Foo, 'Bar ], leading to misparse as a character literal
Input:

```haskell
type Foo = Cluster '[ 'NodeCore, 'NodeRelay, 'NodeEdge ]
```

Output:

```haskell
type Foo = Cluster '['NodeCore, 'NodeRelay, 'NodeEdge]
```

```
<stdin><rendered>:1:20
parse error on input ‘'’
```

sphaso:
Fixes #184 

Thanks @gabrielelana !

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

### DIFF
--- a/src/Ormolu/Printer/Combinators.hs
+++ b/src/Ormolu/Printer/Combinators.hs
@@ -52,8 +52,8 @@ import Data.List (intersperse)
 import Data.Text (Text)
 import Ormolu.Printer.Comments
 import Ormolu.Printer.Internal
-import Ormolu.Utils (isModule)
-import Outputable (Outputable (..), showSDocUnsafe)
+import Ormolu.Utils (isModule, showOutputable)
+import Outputable (Outputable)
 import SrcLoc
 import qualified Data.Text as T
 
@@ -72,7 +72,7 @@ txt t = ensureIndent >> spit t
 -- does have an 'Outputable' instance.
 
 atom :: Outputable a => a -> R ()
-atom = txt . T.pack . showSDocUnsafe . ppr
+atom = txt . T.pack . showOutputable
 
 -- | Enter a 'Located' entity. This combinator handles outputting comments
 -- and sets layout (single-line vs multi-line) for the inner computation.
--- a/src/Ormolu/Printer/Meat/Type.hs
+++ b/src/Ormolu/Printer/Meat/Type.hs
@@ -37,6 +37,9 @@ p_hsType = \case
     case p of
       Promoted -> txt "'"
       NotPromoted -> return ()
+    case showOutputable (unLoc n) of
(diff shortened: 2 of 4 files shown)


src/Ormolu/Printer/Combinators.hs
```diff
@@ -67,12 +68,15 @@ import qualified Data.Text as T
 txt :: Text -> R ()
 txt t = ensureIndent >> spit t
 
+showMe :: Outputable a => a -> Text
+showMe = T.pack . showSDocUnsafe . ppr
```
mrkkrp: `showMe` is not a very good name for a function. In addition you don't add any documentation. Finally we have something like this in `Oromlu.Utils`:

```haskell
-- | Pretty-print an 'GHC.Outputable' thing.

showOutputable :: GHC.Outputable o => o -> String
showOutputable = GHC.showSDocUnsafe . GHC.ppr
```

mrkkrp:
I reformatted the code to be able to see what was changed easily. I think the solution may be a bit hacky. Need to think of something else. Will return to this tomorrow perhaps.

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -100,14 +101,28 @@ p_hsType = \case
...
   XHsType (NHsCoreTy t) -> atom t
 
+prefixForPromotedListTy :: [LHsType GhcPs] -> R ()
+prefixForPromotedListTy (l@(L _ _):_) = located l prefix
+  where
+    prefix (HsTyVar _ Promoted _) = space
+    prefix _                      = return ()
+prefixForPromotedListTy _ = return ()
+prefixForPromotedTy = located' (\name ->
+ if last (showOutputable name) == '\'' then
```
basile-henry: Looking at the `last` character is not going to be enough I don't think.
This is valid Haskell:

```haskell
{-# LANGUAGE DataKinds      #-}
{-# LANGUAGE KindSignatures #-}

data A = A'A

data B (a :: A) = B

b :: B ' A'A
b = B
```

basile-henry: Checking if the second character is a `'` might be good enough! :)

```haskell
case showOutputable name of
  _:'\'':_ -> space
  _ -> return ()
```

sphaso: Thanks!
I see why checking for the last character won't work, but what about:
```haskell
data A = VeryLong' Name' 
```
perhaps we should check if the first non-alphanumeric character is a `'`?

PR #233, opened 2019-07-20
Format unboxed sum pattern

### LINKED ISSUE #185: Unboxed sums in patterns are not supported
Input:

```haskell
{-# LANGUAGE UnboxedSums #-}
p = True
  where
    (# _x | #) = (# | True #)
```

Output:

```haskell
{-# LANGUAGE UnboxedSums #-}

p = True
  where
    _x = (# | True #)
```

gabrielelana:
Fixes #185

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

### DIFF
--- a/data/examples/declaration/value/function/pattern/unboxed-sum-pattern-out.hs
+++ b/data/examples/declaration/value/function/pattern/unboxed-sum-pattern-out.hs
@@ -1,0 +1,24 @@
+{-# LANGUAGE UnboxedSums #-}
+
+v = True
+  where
+    (# _x #) = (# True #)
+
+p = True
+  where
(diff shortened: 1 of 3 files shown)


mrkkrp:
This case is more complicated than it seems:

* Multiline formatting of this sort of patterns should also be supported.
* It's better to try to re-use the logic we already have for expressions involving unboxed sums, see `ExplicitSum` in `Value.hs`.
* Even then there is the problem of #198, which #223 tries to fix.

mrkkrp:
Now that #245 is merged, we're finally ready to solve this one properly.

PR #239, opened 2019-07-21
Adjust when we put parentheses around TH-quoted names

mrkkrp:


mrkkrp:
There are still problems with this: `''Semigroup.Option`. Need a less hacky criteria for deciding when to add the parens. Let's return to this tomorrow perhaps.

PR #245, opened 2019-07-30
Change parens style and how separated lists of items are printed

### LINKED ISSUE #182: Sanity check fails on a weird pattern binding
This is from cardano-sl.

```haskell
(       getNodeSettingsR
 :<|>   getNodeInfoR
 :<|>   getNextUpdateR
 :<|>   restartNodeR
 ):<|>( getUtxoR
 :<|>   getConfirmedProposalsR
 ) = client nodeV1Api
```

Ormolu produces:

```haskell
( getNodeSettingsR :<|>
  getNodeInfoR :<|>
  getNextUpdateR :<|>
  restartNodeR
) :<|>
  ( getUtxoR :<|>
    getConfirmedProposalsR
  ) = client nodeV1Api
```

```
<stdin><rendered>:5:1
```

### LINKED ISSUE #198: Broken layout for multiline patterns in case matches

### LINKED ISSUE #244: Parenthesized expression inside a do-block results in invalid Haskell
Here is an example I extracted from trying to format `hindent`:

```haskell
foo = do (bar
          baz)
```

this gets currently re-formatted as:

```haskell
foo = do
  ( bar
    baz
  )
```

which is not valid Haskell.

mrkkrp:
Close #244, close #182, close #198, close #223.

This is a work in progress. If successful it will:

1. Make the set of combinators clearer and indeed smaller.
2. Solve a number of issues, such as those about parse failures related to patterns.
3. Solve the bug from #244.

The idea is very simple, we stop doing this

```haskell
( foo
, bar
)
```

and start doing this

```haskell
(foo,
 bar)
```

There are three points here to note:

1. We switch to trailing commas which solves the indentation issues for patterns automatically.
2. We do not put closing punctuation on its own line. Which looks better when we have lots of nested punctuation and at the same time solves issues like #244.
3. The new general `sep` combinator finally is clear enough, and all the old zoo of `velt` and `velt'` and `sepWith`, etc. which was confusing and overlapping goes away.

Polishing this will take some time.


drathier:
Is the intention to move all commas to the end of each line, also for list literals? What about chained infix operators like `a <> b <> c` or `a <$> b <*> c`, where do they go when split into multiple lines? What about records? Etc.

mrkkrp:
Let's not get ahead of ourselves. Chaining of operators in its present form does not *break* anything and there is a separate issue about that. Records will be formatted like this:

```haskell
data MyData = MyData
  {myFoo :: Int,
   myBar :: Bar
  }
```

int-index:
Shouldn't it be

```
( foo,
  bar )

data MyData = MyData
  { myFoo :: Int,
    myBar :: Bar
  }
```

?

Then at any point your indentation is a multiple of 2.

mrkkrp:
Then #244 is not solved, or do you propose to have two types of parens?

int-index:
@mrkkrp I don't think we're on the same page, I took your example and added a space after the opening parenthesis and a space before the closing parenthesis:

```
-- your formatting
(foo,
 bar)

-- my formatting
( foo,
  bar )
```

I can't think of a reason this stylistic choice would have any effect on #244, and I do not propose two types of anything.

mrkkrp:
This takes some time to get right. TODO:

- [ ] Ensure that all examples have changed correctly (without introducing undesirable formatting features)
- [ ] Review the code itself and clean it up perhaps
- [ ] Add new examples making sure we also fixed the bugs we wanted to fix
- [ ] Try on some packages from Hackage to see how the formatter behaves after the change in general

This is not for today though.

PR #249, opened 2019-08-04
Avoid crashing on unboxed string literals

### LINKED ISSUE #187: Crashes on unboxed string literals
Input:

```haskell
{-# LANGUAGE MagicHash #-}
main = new "p"#
```

Output:

```
ormolu: Data.ByteString.ByteString.toConstr
CallStack (from HasCallStack):
  error, called at libraries/bytestring/Data/ByteString/Internal.hs:184:20 in bytestring-0.10.8.2:Data.ByteString.Internal
```

Surprisingly, the crash happens during sanity checking, not during formatting. `ormolu -u` works fine.

mrkkrp:
Close #187.

### DIFF
--- a/data/examples/declaration/value/function/unboxed-string-lit-out.hs
+++ b/data/examples/declaration/value/function/unboxed-string-lit-out.hs
@@ -1,0 +1,3 @@
+{-# LANGUAGE MagicHash #-}
+
+main = new "p"#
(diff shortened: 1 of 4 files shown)


mrkkrp:
With this I was able to format a code base as serious as that of `attoparsec`.

PR #255, opened 2019-08-06
Change formatting of the operators

### LINKED ISSUE #95: Improve the way operators are chained
There are few issues with current formatting of chains of operators. First of all, I think we should revert to this style with operator on the second line before the second argument:

```haskell
foo
  + bar
```

and in a sequence it should be:

```haskell
foo
  + bar
  + baz
```

This will allow us to have reasonable rendering of e.g. applicative notation.

`do`-notation should be treated specially:

```haskell
foo
  + bar
  + baz $ do
```

Also, things applied before `do` with `($)` should still make the whole construction hanging:

```haskell
foo = foo $ do
  ...
````

vs current:

```haskell
foo =
  foo $ do
    ...
```

The change described in #94 should then still allow users to have a newline before such constructions involving `do`.

### LINKED ISSUE #176: Ormolu produces unparseable code on a peculiar case expression
This has been distilled from wire-server.

```haskell
update = 
    do
      foobar
    `catch` \case
      a -> a
```

Ormolu produces (in unsafe mode):

```haskell
update =
  do
    foobar `catch` \case
    a -> a
```

### LINKED ISSUE #179: Unparseable code when formatting 'do' with operators
This compiles:

```haskell
foo = do
    1
    +
    2
```

Ormolu's result doesn't:

```haskell
foo =
  do
    1 +
    2
```

```
/Users/yom/tmp/do.hs:4:5: error: …
  |
```

### LINKED ISSUE #207: do-notation with operators fails the sanity check
Original code:

```
main =
  do stuff
   `finally` do
     recover
```

Ormolued code:

```
main =
  do
    stuff `finally` do
    recover
```

Now `recover` isn't a part of `finally`, silently changing the meaning of the program.

This is responsible for one of the failures in #200

### LINKED ISSUE #220: Common applicative chain application idiom is rendered quite sub-optimally
Input:

```haskell
polysemyStuff :: TcPluginM (PolysemyStuff 'Things)
polysemyStuff =
  let PolysemyStuff a b c d = polysemyStuffLocations
  in PolysemyStuff <$> doLookup a
                   <*> doLookup b
                   <*> doLookup c
                   <*> doLookup d
```

Output:

```haskell
polysemyStuff :: TcPluginM (PolysemyStuff 'Things)
polysemyStuff =
  let PolysemyStuff a b c d = polysemyStuffLocations
   in PolysemyStuff <$> doLookup a <*>
        doLookup b <*>
        doLookup c <*>
        doLookup d
```

utdemir:
Close #95, close #176, close #179 and close #207, close #220.

Currently, all tests are pasing except `declaration/value/function/parallel-comprehensions.hs`, where there is an issue about comment placement which makes the AST differ. The comment issue can be summarized as:

```haskell
-- before
main =
  bar $ -- bar
    baz -- baz

-- after
main =
  bar
    $ baz -- baz -- bar
```

Besides the comment issue, I also did not implement the special treatment of `do` notation yet, so currently the common idiom:

```haskell
foo $ do
  bar
  baz
```

now becomes:

```haskell
foo
  $ do
    bar
    baz
```

Which looks a bit unusual, but syntactically valid. Otherwise all looks reasonable to me. I tested a few major Hackage packages and couldn't find any introduced regressions and on the plus side now we can format `pandoc`.

I will work on fixing the `do` notation tomorrow, and take a look at the comment placement issue later on.

### DIFF
--- a/data/examples/declaration/value/function/arrow/proc-do-complex-out.hs
+++ b/data/examples/declaration/value/function/arrow/proc-do-complex-out.hs
@@ -31,8 +31,8 @@ foo
               w
               ) -> \u -> -- Procs can have lambdas
               let v =
-                    u ^ -- Actually never used
-                      2
+                    u -- Actually never used
+                      ^ 2
               in ( returnA -<
                      -- Just do the calculation
                      (x + y * z)
@@ -55,7 +55,7 @@ foo
           then ma -< ()
           else returnA -< ()
           returnA -<
-            ( i +
-                x *
-                y -- Just do the calculation
+            ( i
+                + x
+                * y -- Just do the calculation
               )
--- a/data/examples/declaration/value/function/arrow/proc-forms-out.hs
+++ b/data/examples/declaration/value/function/arrow/proc-forms-out.hs
@@ -9,15 +9,15 @@ foo1 f g h x =
 foo2 f g h x =
   proc (y, z) ->
     (| test
-         ( h f .
-             h g -<
-             y x .
-               y z
+         ( h f
+             . h g -<
+             y x
+               . y z
            )
-         ( h g .
-             h f -<
-             y z .
-               y x
+         ( h g
+             . h f -<
+             y z
+               . y x
            )
       |)
 
@@ -30,6 +30,6 @@ bar1 f g h x =
 bar2 f g h x =
   proc (y, z) ->
     (h f . h g) -<
-      (y x) . y z |||
-        (h g . h f) -<
+      (y x) . y z
+        ||| (h g . h f) -<
         y z . (y x)
--- a/data/examples/declaration/value/function/arrow/proc-ifs-out.hs
+++ b/data/examples/declaration/value/function/arrow/proc-ifs-out.hs
@@ -5,8 +5,8 @@ foo f = proc a -> if a then f -< 0 else f -< 1
 bar f g = proc a ->
   if f a
   then
-    f .
-      g -<
+    f
+      . g -<
       a
   else
     g -<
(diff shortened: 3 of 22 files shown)


utdemir:
I found one more example when a comment breaks the layout:

```
foo = bar ++
    {- some comment -}
    case foo of
      a -> a
```

After formatting:

```
foo = bar ++ {- some comment -}
case foo of
  a -> a
```

utdemir:
I believe I implemented all the changes mentioned in #95. There are still a few tests failing because of two issues regarding the comment placement. I mentioned both of the issues in this thread, one of them is about merging of multiple line comments to one, the other one is about multi-line comments introducing a newline without regarding indentation. I could not find a non-hacky way to fix those, I would appreciate any kind of help @mrkkrp .

Also, I am not 100% confident about the change on `exprPlacement` function; it seems to work fine, however I feel like it might look strange when the left hand side is longer than a few words. We can improve this later on.

utdemir:
I just noticed that it partially reverts #251 . I'm going to fix it by only hanging the second argument when both arguments are on the same line.

mrkkrp:
This example

```haskell
foo = bar ++
    {- some comment -}
    case foo of
      a -> a
```

seems to be working now. It produces:

```haskell
foo = bar
  ++ {- some comment -}
  case foo of
    a -> a
```

Which is a valid thing.

I'll see about merging of comments.

mrkkrp:
I fixed the issue with merging comments. That's an interesting corner case. I had to use some new things from #257 to make it work. After that I updated the remaining tests, added some more test cases to `operators.hs` ( verbatim taken from the issues we're trying to close). It all seems to work. I'm going to merge this now.

mrkkrp:
BTW, you can just create branches in this repo directly. No need to use a fork to submit changes.

PR #256, opened 2019-08-06
Fixing the problem with decorations in multi-equation function definitions

### LINKED ISSUE #193: Backticks or parentheses in function declarations are gained/lost
### Backticks are gained if the first equation has them

Input:

```haskell
x `op1` y = True
op1 x y = False
```

Output:

```haskell
x `op1` y = True
`op1` x y = False
```

### Backticks are lost if the first equation doesn't have them

Input:

```haskell
op2 x y = False
x `op2` y = True
```

Output:

```haskell
op2 x y = False
x op2 y = True
```

teofr:
Close #193.

Added 2 tests for this issue.
Changed the way the group of functions are matched, ideally this should be improved into nicer looking code in a future commit.

The main problem was that the annotations made by the GHC parser drops the information of all but the first definition of a multilime function definition. We need to treat these differently and go recover each of the definition's information from deep bottom (in the context).

More information on the `Note [m_ctxt in Match]` in here http://hackage.haskell.org/package/ghc-8.6.4/docs/src/HsExpr.html#Match.


data/examples/declaration/value/pattern-synonyms/explicitely-bidirectional-out.hs
```diff
@@ -13,5 +13,4 @@ pattern HeadC' x <-
 pattern Simple <-
   "Simple"
   where
-    Simple =
-      "Complicated"
+    Simple = "Complicated"
```
mrkkrp: Looks like we even fixed a minor other bug, too!

Issue #258, opened 2019-08-07
Investigate how hard would it be to support semicolons properly


mrkkrp:
By "support" I mean working with inputs that contain semicolons and outputting normal Ormolu output where semicolons are dropped.

There are some issues related to semicolons: https://github.com/tweag/ormolu/issues?q=is%3Aissue+is%3Aopen+label%3Asemicolons.

The aim of this issue is to investigate:

1. Where semicolons can appear in Haskell syntax in general, for that Haskell report may be useful.
2. Try to assess how hard would it be to support semicolons with what we currently have.
3. What sort of problems prevent us from supporting semicolons.

It may be that semicolons are too tricky to support properly. In that case we'll proceed with #253, otherwise we're going to implement a solution to support semicolons and fix the related bugs.

mrkkrp:
Some packages that currently fail because of semicolons (may be useful for checking real-world examples):

* `aeson`
* `lens`
* `aws`
* `parsec3`

utdemir:
The first thing I can think of is that, converting semicolon syntax to a multi line construct breaks idempotency in the current design. Consider this example

```
let a = a in do { foo; bar; }
```

On the first run, ormolu would think that it is a single line construct, and start formatting it in that way, however it would split that `do` notation to multiple lines when printing:

```
let a = a in do
  foo
  bar
```

But on the next run, it would notice that it's a multiline construct and format it accordingly:

```
let a = a
 in do
      foo
      bar
```

As you can see, somehow we have to detect that our layout is going to be a multiline layout before we even hit the `do`. To achieve this, we will probably have to have a preprocessing step where we convert semicolon-syntax to identation-syntax, and format the result. Or as a more hacky solution, we can simply format the code once more if we detect this issue.

utdemir:
As per Haskell 2010 report, we are be able to use semicolons everywhere layout matters, including the top-level. Below is a type-correct Haskell program with non exhaustive set of examples:

```haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TemplateHaskell #-}

module Semicolons where
{ import Prelude
; f0 = let { a = 1; b = 2 } in b
; f1 = let a = 1; b = 2
           c = 3
       in  b
; f2 = a where { a = 3 }
; f3 = do a <- [1]; let { b = a }; return b
; f4 = case True of { True -> 1; False -> 2 }
; f5 = case f2  of 1 -> 1; 2 -> 2
[...]

PR #263, opened 2019-08-08
Fix printing of TH brackets

### LINKED ISSUE #196: Declaration splices get broken sometimes
Input:

```haskell
{-# LANGUAGE TemplateHaskell #-}

$( do [d| baz = baz |] )
```

Output:

```haskell
{-# LANGUAGE TemplateHaskell #-}

$(do
  [d|baz = baz
  |])
```

```
5:3: error: …
  |
```

utdemir:
Close #196

The actual solution was simply indenting the closing `|]` one more level. However there were a few more issues around them, which led me to a slightly bigger refactor.

The main reason is that, all `p_*Decl` functions used to print a trailing newline. This makes sense for top-level constructs, however it was making printing something like `[d|data Foo = Foo|]` impossible.

In this PR, I removed trailing newlines from individual printers and gave that responsibility to `p_hsDecls`, and inserted an additional trailing newline when printing modules.

While doing that, I noticed a few bugs/inconsistencies, and I had to fix them in the process:

* Warning pragmas used to not print a trailing newline, so they were always attached to the next expression. I made it more like the other pragmas, where we attach it to a neighbour function if the name matches, otherwise we separate it with a newline.
* We used to print single line GADT's and single line `do` notations using multiple lines, which breaks idempotency on some cases (formatting an already formatted output can give a different result). I tweak them to prefer single line layout if possible (sometimes it is not possible because of the semicolon syntax).

### DIFF
--- a/data/examples/declaration/data/gadt-syntax-out.hs
+++ b/data/examples/declaration/data/gadt-syntax-out.hs
@@ -1,3 +1,2 @@
 {-# LANUGAGE GADTSyntax #-}
-data Foo where
-  MKFoo :: a -> (a -> Bool) -> Foo
+data Foo where MKFoo :: a -> (a -> Bool) -> Foo
--- a/data/examples/declaration/warning/warning-single-line-out.hs
+++ b/data/examples/declaration/warning/warning-single-line-out.hs
@@ -1,4 +1,5 @@
-{-# WARNING test "This is a warning" #-}
 {-# DEPRECATED test, foo "This is a deprecation" #-}
+
+{-# WARNING test "This is a warning" #-}
 test :: IO ()
 test = pure ()
--- a/data/examples/declaration/splice/bracket-out.hs
+++ b/data/examples/declaration/splice/bracket-out.hs
@@ -2,23 +2,22 @@
 
 foo =
   [e|
-  foo bar
-  |]
+    foo bar
+    |]
 
 foo =
   [e|
-  foo bar
-  |]
+    foo bar
+    |]
 
 foo = [t|Char|]
 
 foo =
   [d|
-  foo :: Int -> Char
+    foo :: Int -> Char
 
-  bar = 42
-
-  |]
+    bar = 42
+    |]
 
 foo =
   [||
--- a/data/examples/declaration/splice/splice-decl-out.hs
+++ b/data/examples/declaration/splice/splice-decl-out.hs
@@ -15,3 +15,21 @@ foo bar
 -- TemplateHaskell allows Q () at the top level
 do
   pure []
+
+$(do [d|baz = baz|])
+
+[d|data T a where Foo :: T ()|]
+
+[d|
+  data T = T
+    deriving (Eq, Ord, Enum, Bounded, Show)
+  |]
+
+$(singletons [d|data T = T deriving (Eq, Ord, Enum, Bounded, Show)|])
+
+$( singletons
+     [d|
+       data T = T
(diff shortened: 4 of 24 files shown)


data/examples/declaration/data/gadt-syntax-out.hs
```diff
@@ -1,3 +1,2 @@
 {-# LANUGAGE GADTSyntax #-}
-data Foo where
-  MKFoo :: a -> (a -> Bool) -> Foo
+data Foo where MKFoo :: a -> (a -> Bool) -> Foo
```
mrkkrp: I think we should keep the previous style because it's more readable (after all, that's the point of the tool!). Currently on `master` it works by re-formatting single version into the multi-line version without issues.

data/examples/declaration/warning/warning-single-line-out.hs
```diff
@@ -1,4 +1,5 @@
-{-# WARNING test "This is a warning" #-}
 {-# DEPRECATED test, foo "This is a deprecation" #-}
+
+{-# WARNING test "This is a warning" #-}
```
mrkkrp: What do you check with this change? If it's just the presence of newline after deprecation annotation, then wouldn't it be already printed even without warning annotation following it? (I.e. newline would be there between deprecation annotation and `test :: IO ()`.)

data/examples/declaration/data/gadt-syntax-out.hs
utdemir: In its current design, `ormolu` is not able to convert a single-line expression/statement to a multi-line one. The reason is that, since we format things differently on single line and multiple lines, if we re-format a single line-version to a multi-line one, there is a chance that we will format it differently on the next run, breaking the idempotency. As an example:


It works currently on `master` because we don't have a test where we have this problem in a nested scope. As an example, this wouldn't work on master (assuming we applied the fix for the closing bracket):

```
{-# LANGUAGE TemplateHaskell #-}

foo = [d|data Foo where MKFoo :: a -> (a -> Bool) -> Foo|]
```

Run 1:

```
{-# LANGUAGE TemplateHaskell #-}

foo = [d|data Foo where
    MKFoo :: a -> (a -> Bool) -> Foo
  |]
```

run 2:

```
{-# LANGUAGE TemplateHaskell #-}

foo = [d|
  data Foo where
    MKFoo :: a -> (a -> Bool) -> Foo
  |]
```

From ormolu's perspective, the `[d| |]` bracket in a single-line
layout on the "run 1", but on a multi-line layout on "run 2". 

There is another example of this issue in [this comment](https://github.com/tweag/ormolu/issues/258#issuecomment-519342454).

data/examples/declaration/warning/warning-single-line-out.hs
utdemir: I was testing the absence of newline after the `WARNING` annotation. After this PR it's only adjacent to the `test` function because of the [line 143 
 of Declaration.hs](https://github.com/tweag/ormolu/pull/263/files#diff-0ccd0f9b89649783f9d4f7e7a6dc9e5eR143)

data/examples/declaration/data/gadt-syntax-out.hs
mrkkrp: This makes sense to me, thanks for the explanation.

data/examples/declaration/warning/warning-single-line-out.hs
mrkkrp: I think we can keep it "as is".

PR #266, opened 2019-08-09
Re-order OPTIONS_GHC and OPTIONS_HADDOCK pragmas

### LINKED ISSUE #214: GHC and Haddock options should group together with language pragmas on top
`OPTIONS_GHC` and `OPTIONS_HADDOCK` things should group together with language pragmas on top of files.

utdemir:
Closes #214 .

This PR is based on the changes on #265 , so we should merge it first.

This change adds an ad-hoc parser for `OPTIONS_*` pragmas. I did not want to use an existing tokenizer, because I felt like tokenizing and pretty printing the GHC options are more prone to error without providing much benefit. Even if we choose to format the contents, I think we should find another parser (probably still from `ghc` library), since the `tokenizer` we have loses some information(`-O2` and `- O 2` returns exactly the same tokens).

So, the new behaviour is that, we parse the contents of the `LANGUAGE` pragmas and create an extension per line, however we don't do anything to `OPTIONS_GHC` and `OPTIONS_HADDOCK` pragmas other than re-ordering.


mrkkrp:
I rebased on current master. I'll get to reviewing it a bit later.

Issue #189, opened 2019-07-10
Let oneliner with semicolons is broken


neongreen:
Input:

```haskell
foo x = let {g :: Int -> Int; g = id} in ()
```

Output:

```haskell
foo x = let g :: Int -> Int
  g = id in ()
```

mrkkrp:
Semicolons are quite tricky to work with with our approach, yeah.

mrkkrp:
Resolved by #271.

Issue #152, opened 2019-07-04
Case oneliner with semicolons is broken


ethercrow:
Valid input:

```
module Case where

foo :: IO ()
foo = do
    let bar = case [1] of [_] -> "singleton"; _ -> "not singleton"
    print bar
```

Invalid output:

```
module Case where

foo :: IO ()
foo = do
  let bar = case [1] of [_] -> "singleton"
    _ -> "not singleton"
  print bar
```

neongreen:
Latest Ormolu breaks explicitly:

```
foo = case b of 1 -> 2; 2 -> 1
```

```
Parsing of formatted code failed:
<stdin><rendered>:1:7-22
Unexpected case expression in function application:
    case b of { 1 -> 2 }
You could write it with parentheses
Or perhaps you meant to enable BlockArguments?
Please, consider reporting the bug.
```

neongreen:
One more with braces:

```haskell
x <  y = case compare x y of { LT -> True;  EQ -> False; GT -> False }
```

```haskell
x < y = case compare x y of LT -> True
  EQ -> False
  GT -> False
```

```
Parsing of formatted code failed:
  <stdin><rendered>:2:6-7
  parse error on input ‘->’
Please, consider reporting the bug.
```

mrkkrp:
Braces do not work very well. I’m not sure yet what to do with them in general.

PR #268, opened 2019-08-09
Error out on semicolon uses

### LINKED ISSUE #253: Detect if semicolon syntax is used and error-out immediately
We may eventually support semicolon syntax, however not in the first release. I propose we error out if the syntax is even used. Reasons:

* It's better for the image of the project if Ormolu errors out clearly explaining why instead of just failing with non-obvious messages which could make users think that the tool is unreliable.
* It's easier to spot remaining issues if we filter out semicolon-induced ones for now, for example when trying the tool on packages from Hackage.

utdemir:
Closes #253 .

Since there are a few issues about the handling of semicolon-separated
layouts, we error-out on most uses of the semicolon for now.

As a result, #189 and #152 now shows a nice error message like:

```
Parsing of source code failed:
  file: <stdin>, lines: [5]
  Ormolu does not support all uses of semicolons yet.
```

However, since `ormolu` does support some uses of semicolons, we
lost a bit of functionality; now we can not format some packages
like `attoparsec` like we used to, since they contain some semicolon
usages. Maybe we should first try to format the code hoping for the
best, and only check for semicolon uses when our output fails to parse
or parses to a different AST.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,14 @@ also contain formatted `.hs` files for inspection and original inputs with
 `.hs-original` extension (those are with CPP dropped, exactly what is fed
 into Ormolu).
 
+## Current Limitations
+
+* Ormolu currently does not support all uses of semicolons to separate
+expressions, formatting modules with semicolons might fail.
+* CPP is not supported, and Ormolu will error out on CPP enabled
+modules. Use `--tolerate-cpp` flag if your module enables the CPP
+flag without actually using CPP.
+
 ## Contributing
 
 See [CONTRIBUTING.md](./CONTRIBUTING.md).
--- a/data/examples/declaration/value/function/do.hs
+++ b/data/examples/declaration/value/function/do.hs
@@ -1,9 +1,6 @@
 {-# LANGUAGE RecursiveDo #-}
 
-foo = do foo
-
-bar = do foo
-         bar
+bar = do { foo; bar }
 
 baz =
   mdo bar a
--- a/data/examples/declaration/value/function/do.hs
+++ b/data/examples/declaration/value/function/do.hs
@@ -1,6 +1,9 @@
 {-# LANGUAGE RecursiveDo #-}
 
-bar = do { foo; bar }
+foo = do foo
+
+bar = do foo
+         bar
 
 baz =
   mdo bar a
--- a/src/Ormolu.hs
+++ b/src/Ormolu.hs
@@ -24,6 +24,7 @@ import Ormolu.Diff
 import Ormolu.Exception
 import Ormolu.Parser
 import Ormolu.Parser.Result
+import Ormolu.Parser.Semicolon
 import Ormolu.Printer
 import Ormolu.Utils (showOutputable)
 import qualified CmdLineParser as GHC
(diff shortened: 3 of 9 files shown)


mrkkrp:
Actually I'm thinking that we could do a much simpler thing here. How about just adding a note to errors about incomplete support for semicolons?

* We are already pretty good at everything else, so chances are if users run into bugs, they will be related to semicolons and the note will help.
* Otherwise the note is just a note after all, a failure may be still unrelated to semicolons and that's fine.
* Cases when current Ormolu works for semicolons will still be supported.

README.md
```diff
@@ -78,6 +78,11 @@ also contain formatted `.hs` files for inspection and original inputs with
 `.hs-original` extension (those are with CPP dropped, exactly what is fed
 into Ormolu).
 
+## Current Limitations
+
+* Ormolu currently does not support using semicolons to separate
+expressions, therefore refuses to format files with semicolons.
+
```
mrkkrp: Could you also add another bullet point about CPP here?

utdemir:
That's a good idea @mrkkrp , I implemented the change, now if we get an error and there is semicolons, we show a message like:

```shell
$ echo 'do { foo; bar } `finally` cleanup' | ./result/bin/ormolu 
  at <stdin>:1:1-33
Note that not all usages of semicolons are supported at the moment.
  (semicolons found at lines: [1])
```

data/examples/declaration/value/function/do.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE RecursiveDo #-}
 
-bar = do { foo; bar }
```
mrkkrp: Should these changes be dropped now?

src/Ormolu.hs
```diff
@@ -122,3 +124,4 @@ showWarn (GHC.Warn reason l) = unlines
   [ showOutputable reason
   , showOutputable l
   ]
+
```
mrkkrp: A nitpick, but still: please make sure not to add redundant lines like this.

src/Ormolu/Exception.hs
```diff
@@ -35,17 +36,28 @@ instance Exception OrmoluException where
...
            [] -> ["in " ++ path]
-           xs -> xs) ++
-      [ "Please, consider reporting the bug."
-      ]
+           xs -> xs)
+      ++ maybe [] (pure . showSemicolonWarning) w
+      ++ [ "Please, consider reporting the bug."
+         ]
+
+data SemicolonWarning = SemicolonWarning [Int]
```
mrkkrp: Documentation would be good here. In general, even type and most functions can benefit from documentation.

src/Ormolu/Parser.hs
```diff
@@ -67,6 +68,7 @@ parseModule Config {..} path input' = liftIO $ do
             , prCommentStream = comments
             , prExtensions = exts
             }
+
```
mrkkrp: The same here.

src/Ormolu/Parser.hs
```diff
@@ -191,3 +193,5 @@ setDefaultExts flags = foldl' GHC.xopt_set flags autoExts
     allExts = [minBound..maxBound]
 
 deriving instance Bounded Extension
+
+-- Semicolon Warnings
```
mrkkrp: This seems to be just a "header" at the very end of the file? Should be dropped as well, right?

utdemir:
I applied the changes you suggested. Sorry I misunderstood what you've asked. If you think this changes are unnecessary, I would be happy to revert; less code is usually better. So, let me know if you want to just modify the warning messages to always mention the semicolon warning.

utdemir:
However, I think a better option is to just merge #271 and close this PR.

mrkkrp:
We decided to go with #271.

PR #271, opened 2019-08-12
Handle semicolons

### LINKED ISSUE #152: Case oneliner with semicolons is broken

utdemir:
Close #152, #189, #258.

This PR adds a `p_layout` combinator which uses semicolon-separated layout on single line constructs, and indentation based layout on multi line constructs. I implemented some common constructs like `do` and `let` using the new combinator, and they can now properly handle semicolons; however there is still some more work remaining to use it on all. In the end, we should be able to remove the `newline` combinator (should be either replaced by `breakpoint`, or by `p_layout`).

Also, currently this PR uses an inherently incorrect approach when deciding when to omit braces. So, we sometimes use braces unnecessarily (`do { foo }`), and there are probably some edge cases we will format incorrectly (I couldn't figure out an example).

Current test-cases are passing alongside with the issues #152 and #189, and we can now format `aeson` and almost `lens` (there is another issue regarding exporting pattern synonyms).

### DIFF
--- a/data/examples/declaration/class/associated-data-out.hs
+++ b/data/examples/declaration/class/associated-data-out.hs
@@ -1,8 +1,6 @@
 {-# LANGUAGE TypeFamilies #-}
 
-class Foo a where
-
-  data FooBar a
+class Foo a where data FooBar a
 
 -- | Something.
 class Bar a where
--- a/data/examples/declaration/class/functional-dependencies-out.hs
+++ b/data/examples/declaration/class/functional-dependencies-out.hs
@@ -3,9 +3,7 @@
 -- | Something.
 class Foo a b | a -> b
 
-class Bar a b | a -> b, b -> a where
-
-  bar :: a
+class Bar a b | a -> b, b -> a where bar :: a
 
 -- | Something else.
 class
--- a/data/examples/declaration/value/function/lambda-multi-line-out.hs
+++ b/data/examples/declaration/value/function/lambda-multi-line-out.hs
@@ -20,3 +20,8 @@ tricky2 =
   flip all (zip ws gs)
     $ \(wt, gt) ->
       canUnify poly_given_ok wt gt || go False wt gt
+
+foo =
+  prop "is inverse to closure" $ \(f :: StaticPtr (Int -> Int))
+  (x :: Int) ->
+      (unclosure . closure) f x == deRefStaticPtr f x
--- a/src/Ormolu/Printer/Combinators.hs
+++ b/src/Ormolu/Printer/Combinators.hs
@@ -24,6 +24,9 @@ module Ormolu.Printer.Combinators
   , breakpoint'
     -- ** Formatting lists
   , sep
+  , sepSemi
+  , useBraces
+  , dontUseBraces
     -- ** Wrapping
   , sitcc
   , line
@@ -42,6 +45,7 @@ module Ormolu.Printer.Combinators
   )
 where
 
+import Control.Monad
 import Data.Data (Data)
 import Data.List (intersperse)
 import Data.Text (Text)
@@ -155,6 +159,37 @@ sep
   -> [a]                        -- ^ Elements to render
   -> R ()
 sep s f xs = sequence_ (intersperse s (f <$> xs))
+
+-- | Render a collection of elements layout-sensitively using given printer,
+-- inserting semicolons if necessary and respecting 'useBraces' and
+-- 'dontUseBraces' combinators.
+--
+-- > useBraces $ sepSemi txt ["foo", "bar"]
+-- >   == vlayout (txt "{ foo; bar }") (txt "foo\nbar")
+--
+-- > dontUseBraces $ sepSemi txt ["foo", "bar"]
+-- >   == vlayout (txt "foo; bar") (txt "foo\nbar")
+
+sepSemi
+  :: (a -> R ())                -- ^ How to render an element
+  -> [a]                        -- ^ Elements to render
+  -> R ()
+sepSemi f xs = vlayout singleLine multiLine
+  where
+    singleLine = do
+      ub <- canUseBraces
+      case xs of
+        [] -> when ub $ txt "{}"
+        xs' ->
+          if ub
+          then do
+            txt "{ "
+            sep (txt "; ") (dontUseBraces . f) xs'
+            txt " }"
+          else
+            sep (txt "; ") f xs'
+    multiLine =
+      sep newline (dontUseBraces . f) xs
 
 ----------------------------------------------------------------------------
 -- Wrapping
--- a/data/examples/declaration/class/associated-type-defaults-out.hs
+++ b/data/examples/declaration/class/associated-type-defaults-out.hs
@@ -1,8 +1,6 @@
 {-# LANGUAGE TypeFamilies #-}
 
-class Foo a where
-
-  type FooBar a = Int
+class Foo a where type FooBar a = Int
 
 -- | Something.
 class Bar a where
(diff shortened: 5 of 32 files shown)


utdemir:
I implemented a nicer logic for braces, and added a few tests. This change touches a lot of things, and it is possible that there're a few hidden blugs. However, even if it is not complete, I think it is "good enough" for most cases. What do you think @mrkkrp ? How much time should we spend on this?

mrkkrp:
> and it is possible that there're a few hidden bugs

You can try the tool on various packages and inspect how it formats things to gain more confidence.

> How much more time we should spend on this?

If you feel like spending a bit more time on this will allow you to improve how things work, feel free. This feature is more important than all the others we have for 0.0.2.0.

utdemir:
I wrote a [small Nix expression](https://gist.github.com/a28a45739819a65a52983f7701a4554b) to compare different `ormolu` versions using a list of Hackage packages. 

What I found is that, for all of the packages I've listed, none of them started to return more errors and now we can format some common packages like `aeson`, `cassava`, `diagrams-core`, `distributed-process`, `idris`, `ShellCheck`, `leksah` and `intero` successfully.

data/examples/declaration/class/associated-data-out.hs
```diff
@@ -1,8 +1,6 @@
 {-# LANGUAGE TypeFamilies #-}
 
-class Foo a where
-
```
mrkkrp: Is it possible to keep the old behavior here? I think putting `class` and `data` actually helps readability, or are there any concerns about idempotence or something else that makes it problematic?

data/examples/declaration/class/functional-dependencies-out.hs
```diff
@@ -3,9 +3,7 @@
 -- | Something.
 class Foo a b | a -> b
 
-class Bar a b | a -> b, b -> a where
-
-  bar :: a
+class Bar a b | a -> b, b -> a where bar :: a
```
mrkkrp: Similar question here.

data/examples/declaration/value/function/lambda-multi-line-out.hs
```diff
@@ -20,3 +20,8 @@ tricky2 =
   flip all (zip ws gs)
     $ \(wt, gt) ->
       canUnify poly_given_ok wt gt || go False wt gt
+
+foo =
+  prop "is inverse to closure" $ \(f :: StaticPtr (Int -> Int))
```
mrkkrp: This rendering of lambda is quite ugly, maybe we should open an issue about it?

src/Ormolu/Printer/Combinators.hs
```diff
@@ -270,3 +277,28 @@ comma = txt ","
 
 space :: R ()
 space = txt " "
+
+----------------------------------------------------------------------------
+-- Layout
+
+data LayoutOptions
```
mrkkrp: Documentation is very important, if you look at other functions in this module, they are all documented very carefully because this module is important for building everything else. Good documentation also makes it easier for reviewers to understand how things work.

I like how you describe your PRs, surely the actual code which will be read perhaps many years later and not once deserves not less attention :)

src/Ormolu/Printer/Combinators.hs
```diff
@@ -270,3 +277,28 @@ comma = txt ","
...
+  = LayoutOptions
+      { loSit :: Bool
+      }
+
+defaultLayoutOptions :: LayoutOptions
+defaultLayoutOptions = LayoutOptions False
+
+p_layout :: LayoutOptions -> [R ()] -> R ()
```
mrkkrp: We should choose a different name for this function because the `p_` convention is reserved for functions printing various parts of AST, it's not for combinators. Perhaps I never documented this explicitly...

src/Ormolu/Printer/Combinators.hs
```diff
@@ -270,3 +277,28 @@ comma = txt ","
...
+
+----------------------------------------------------------------------------
+-- Layout
+
+data LayoutOptions
+  = LayoutOptions
+      { loSit :: Bool
+      }
+
+defaultLayoutOptions :: LayoutOptions
```
mrkkrp: I'm not sure if you win anything with the `LayoutOptions` here. It adds verbosity, but maybe it would be better to have a sum type isomorphic to `Bool` here just with more meaningful names for data constructors? Do you plan to extend `LayoutOptions` with other parameters?

utdemir: You are right. Initially, that `LayoutOptions` had two more fields, but I managed to remove them on later iterationsm. I'll make it a sum type type.

src/Ormolu/Printer/Combinators.hs
utdemir: You are right. I will change it to `layout`.

data/examples/declaration/class/associated-data-out.hs
utdemir: I also think it hurts readability to have single line data/class/instance declarations.

However, as you mentioned, there are idempotency concerns, ideally we should not convert any single line constructs to multi line ones. Here are some examples (hand-written, so  take them with a grain of salt):

```haskell
[d|class Foo a where data Foobar a|]
-- 1st pass:
[d|class Foo a where
  data Foobar a|]
-- 2nd pass:
[d| class Foo a where
  data Foobar a |]
```

```haskell
module M where { class Foo a where data Foobar a; foo = bar }
-- 1st pass
module M where { class Foo a where
  data Foobar a; foo = bar }
-- 2nd pass:
module M where

class Foo a  where
  data Foobar a

foo = bar
```

In short, since we change how we format things based on the fact that the enclosing SrcSpan is multi-line or not, we can not easily convert things to multi-line with the current design.

If we want to be able to do this kind of things, we can work-around this issue with some tweaks. For example, we can tweak the `located` combinator; so if we spit a newline inside the inner computation, we discard its effects and run it again as a mult-line layout (this should bubble up to enclosing `Located`'s as well). Actually, a similar fix can help us with unimplemented constructs as well; if we catch a "not implemented" error in a `Located` object, we can just use the original `SrcSpan` without formatting, and hope that it's still a valid syntax.

data/examples/declaration/class/functional-dependencies-out.hs
utdemir: See my answer to the previous comment.

data/examples/declaration/class/associated-data-out.hs
utdemir: Just to clarify; the reason we're formatting this in a single line is that the input is single line. We would format it as before if the input were multi-line.

utdemir:
Thank you for the review @mrkkrpv. I addressed your points. Please feel free to change the wording on the docs if they sound strange.

The reason I want to move away from the newline combinator is that; we do not support using it when formatting single-line constructs (because of the idempotency issue discussed in the comments above), and since `breakpoint` and `breakpoint'` combinators have the same behaviour as `newline` on multi-line constructs there is no situation where using `newline` is preferable.

mrkkrp:
@utdemir Please note the changes I performed in 4af0fb19d77eae7876f9bccb7d55c9cb15775300. Another question: if I replace `useBraces` with `dontUseBraces` everywhere, tests still pass. It means then, that either the tests do not cover something important or braces are not that necessary and the code can be further simplified.

What's the motivation for having braces?

PR #272, opened 2019-08-12
Fix mis-formatting of per-argument comments in certain cases

### LINKED ISSUE #260: Comments are output in a weird way for function arguments
Here is a very strange bug:

```haskell
runParserT
  :: Monad m
  => ParsecT e s m a -- ^ Parser to run
  -> String
  -> -- ^ Name of source file
  s
  -> -- ^ Input for parser
  m (Either (ParseErrorBundle s e) a)
```

Can be reproduced with current master by ormolizing `megaparsec`. The function is in `Text.Megaparsec`.

This bug is strange because in all other cases the output is perfectly fine. For example:

```haskell
runParser
  :: Parsec e s a -- ^ Parser to run
  -> String -- ^ Name of source file
  -> s -- ^ Input for parser
  -> Either (ParseErrorBundle s e) a
```

It would be interesting to figure this one out.

mrkkrp:
Close #260.

### DIFF
--- a/data/examples/declaration/instance/contexts-out.hs
+++ b/data/examples/declaration/instance/contexts-out.hs
@@ -25,14 +25,12 @@ instance
   ( Read a, -- Foo
     Read b,
     Read
-      ( c,
-        -- Bar
+      ( c, -- Bar
         d
         )
     )
   => Read
-       ( a,
-         -- Baz
+       ( a, -- Baz
          b,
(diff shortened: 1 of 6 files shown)


mrkkrp:
This one is good, however CI does not pass because of newly discovered #273 (which is present even in current master, it's just the tests do not reveal it).

PR #287, opened 2019-08-14
Fix formatting of infix type synonyms

### LINKED ISSUE #284: Infix type alias is formatted incorrectly
Found on `singletons` codebase.

Original:

```haskell
type a ~> b = TyFun a b -> Type
```

After `ormolu --unsafe`:

```haskell
type ~> a b = TyFun a b -> Type
```

Which is not a valid syntax:

```
  infile<rendered>:1:6-7
  Operator applied to too few arguments: ~>
```

utdemir:
Closes #284 .

This PR takes the fixity of the type alias into account when printing.

I am making an assumption that an infix type alias has exactly two type variables. I can not think of a counter-example.


aspiwack:
> I am making an assumption that an infix type alias has exactly two type variables. I can not think of a counter-example.

There aren't a lot of reason for this not to be the case. But it is not a property of the language. There can, in fact, be any number of type variables, from 0 (the most common non-2 value) to however large a number.

Here is an example with 0, which be turn in a example with 1 to 3 type variable as well (though, there would be little good reason, in this case, to have it be any other number than 0 or 2):

```haskell
data Sum f g a = Left (f a) | Right (g a)

type (:+:) = Sum
```

utdemir:
Thank you @aspiwack, that makes sense. It looks like your example (`type (:+:) = ...`) is parsed as a prefix synonym, but I was able to get to find something with more than 2 type varialbes:

```haskell
type ((((a :+: b) c) d) e) = ()
```

However, I couldn't figure a way to get a infix type synonym with 0 or 1 type variables.

I will fix this issue today.

PR #302, opened 2019-08-15
Dedent parallel list comprehensions by one step

### LINKED ISSUE #301: ParallelListComp
A parallel list comprehension is styled this way:

```haskell
{-# LANGUAGE ParallelListComp #-}

[ a
  | x <- xs,
    predicate x,
    let x' = f x
    | y <- ys,
      predicate y,
      let y' = f y
    | z <- zs,
      predicate z,
      let z' = f z
  ]
```

as if `y` and `z` are descendants of `x`. But no, they are parallel, and must look like siblings:

```haskell
[ a
  | x <- xs,
    predicate x,
    let x' = f x
  | y <- ys,
    predicate y,
    let y' = f y
  | z <- zs,
    predicate z,
    let z' = f z
  ]
```

Bug?

yumiova:
Close #301.

This pull request removes indentation of the '|' character in parallel list comprehensions by one layer, as specified in #301. Said indentation was caused by a use of `sitcc` which, when removed, aligns things properly.

A comment is also added to document this in case anyone later wonders why one expression uses `sitcc` and the other does not.

### DIFF
--- a/data/examples/declaration/value/function/parallel-comprehensions-out.hs
+++ b/data/examples/declaration/value/function/parallel-comprehensions-out.hs
@@ -21,19 +21,19 @@ baz x y z w =
       a
         `mod` b -- Value
         == 0
-      | c <- -- Baz 1
-          z
-            * z -- Baz 2
-            -- Baz 3
-      | d <- w -- Other
-      | e <- x * x -- Foo bar
-      | f <- -- Foo baz 1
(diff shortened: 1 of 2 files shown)


mrkkrp:
Let's not but the label "bug" on PRs that fix bugs, unless they also introduce some bugs :)

PR #305, opened 2019-08-16
Hang after block arguments

### LINKED ISSUE #241: BlockArguments gets formatted in an unpleasing way
```haskell
foo do
  bar
```
is currently formatted as
```haskell
foo
  do
    bar
```
Is that intentional? Thanks.

utdemir:
Closes #241 .

When the value of the function application has a hanging form (can happen after `-XBlockArguments`), hang the construct.

### DIFF
--- a/data/examples/declaration/value/function/block-arguments-out.hs
+++ b/data/examples/declaration/value/function/block-arguments-out.hs
@@ -1,0 +1,24 @@
+f1 = foo do bar
+
+f2 =
+  foo do
+    bar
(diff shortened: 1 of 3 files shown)


mitchellwrosen:
@utdemir I noticed your tests don't contain any do-lambda forms like

```haskell
foo \x -> y
```

(Not sure how we want to format that, or how it looks before/after your change, but it is something that's only possible with `BlockArguments`)

PR #306, opened 2019-08-16
Preserve the name of the expression bracket

### LINKED ISSUE #203: [|foo|] is turned into [e|foo|]
`[|foo|]` and `[e|foo|]` are equivalent, but if it's impossible to distinguish between them at the AST level, perhaps at least the `[|foo|]` form could be considered the standard?

Alternatively you can look at the source text to distinguish them, I guess.

utdemir:
Close #203 .

Both `[e|...|]` and `[||]` mean the same thing, and instead of sticking to a single style, we can just preserve what user wrote.

### DIFF
--- a/src/Ormolu/Printer/Internal.hs
+++ b/src/Ormolu/Printer/Internal.hs
@@ -76,7 +76,7 @@ data RC = RC
   , rcDebug :: Bool
     -- ^ Whether to print debugging info as we go
   , rcEnclosingSpans :: [RealSrcSpan]
-    -- ^ Span of enclosing element of AST
+    -- ^ Spans of enclosing elements of AST
   , rcAnns :: Anns
     -- ^ Collection of annotations
   , rcCanUseBraces :: Bool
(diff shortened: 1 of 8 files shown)


src/Ormolu/Printer/Internal.hs
```diff
@@ -326,12 +327,17 @@ setIndent i m' = do
...
--- | Get 'RealSrcSpan' of enclosing span for given referencne span.
+-- | Get 'RealSrcSpan' of enclosing span for given reference span.
 
 getEnclosingSpan :: RealSrcSpan -> R (Maybe RealSrcSpan)
 getEnclosingSpan r =
   listToMaybe . dropWhile (== r) <$> R (asks rcEnclosingSpans)
 
+-- | Get 'RealSrcSpan' of the current enclosing span.
+
+getEnclosingSpan' :: R (Maybe RealSrcSpan)
```
mrkkrp: When situations like this arise, it's often better to make the old function more powerful instead of adding more versions of it.

PR #307, opened 2019-08-16
Change MultiWayIf layout to keep consistent 2-space indent increase

### LINKED ISSUE #300: What is the policy for odd (2k+1) indentation?
MultiWayIf is formatted such way:
```haskell
if | condition -> _
   | condition -> _ -- 3 spaces
   | condition ->
     continuation -- 5 spaces
       and_so_on -- 7 spaces
```

Is it intended?

cblp:
Close #300

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -549,8 +549,9 @@ p_hsExpr = \case
       breakpoint
       inci (p_hsExpr x)
   HsMultiIf NoExt guards -> do
-    txt "if "
-    sitcc $ sep newline (located' (p_grhs RightArrow)) guards
+    txt "if"
+    breakpoint
+    inci . sitcc $ sep newline (located' (p_grhs RightArrow)) guards
   HsLet NoExt localBinds e -> do
     txt "let "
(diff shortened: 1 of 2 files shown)


src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -549,8 +549,9 @@ p_hsExpr = \case
       breakpoint
       inci (p_hsExpr x)
   HsMultiIf NoExt guards -> do
-    txt "if "
-    sitcc $ sep newline (located' (p_grhs RightArrow)) guards
+    txt "if"
```
mrkkrp: Here, notice we first output the literal `"if"`, then we need either single space or newline, this is called `breakpoint`. Then we need to make sure that the rest is more indented, so we use `inci`. It has no effect in single-line case, and bump indentation as you'd expect in multiline case.

cblp: Thanks! I will try to use them in future patches.

cblp:
@mrkkrp, you rewrote whole my change, and now I'm not its author.

mrkkrp:
You're the author and I'm the committer, technically.

PR #318, opened 2019-08-18
Use --check-idempotency flag on tests

utdemir:
Since #308, Ormolu is able to check if the formatting is idempotent. Using that, we can remove the idempotency check on the tests.

### DIFF
--- a/tests/Ormolu/PrinterSpec.hs
+++ b/tests/Ormolu/PrinterSpec.hs
@@ -27,18 +27,19 @@ checkExample srcPath' = it (fromRelFile srcPath' ++ " works") $ do
   expectedOutputPath <- deriveOutput srcPath
   -- 1. Given input snippet of source code parse it and pretty print it.
   -- 2. Parse the result of pretty-printing again and make sure that AST
-  -- is the same as AST of the original snippet. (This happens in
-  -- 'ormoluFile' automatically.)
-  formatted0 <- ormoluFile defaultConfig (fromRelFile srcPath)
-  -- 3. Check the output against expected output. Thus all tests should
-  -- include two files: input and expected output.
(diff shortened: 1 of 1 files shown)


mrkkrp:
The only problem here is that failure reporting is now not so nice (bringing back the idempotence test to the test suite):

```
  tests/Ormolu/PrinterSpec.hs:25:25: 
  1) Ormolu.Printer declaration/value/function/operators.hs works
       uncaught exception: OrmoluException
       OrmoluNonIdempotentOutput SrcLoc "data/examples/declaration/value/function/operators.hs<rendered>" 2 7 "ar\n    ++ {- some co" "ar\n    ++ case foo o"
```

Even if we use `withPrettyOrmoluExceptions` to wrap the test, we get

```
<...>
  declaration/value/function/prefix.hs works
Formatting is not idempotent:
  data/examples/declaration/value/function/operators.hs<rendered>:2:7
  before: "ar\n    ++ {- some co"
  after:  "ar\n    ++ case foo o"

  declaration/value/function/operators.hs works FAILED [1]
  declaration/value/function/static-pointers.hs works
<...>
```

in the middle of the list of passed/failed tests and then:

```
  tests/Ormolu/PrinterSpec.hs:25:25: 
  1) Ormolu.Printer declaration/value/function/operators.hs works
       uncaught exception: ExitCode
       ExitFailure 6

  To rerun use: --match "/Ormolu.Printer/declaration/value/function/operators.hs works/"
```

in the explanation section.

It may be against the DRY principles, but I'm actually in favor of keeping the extra check here and preserving the old reporting style, which with shorter examples should be even more user-friendly.

utdemir:
That makes sense, I'm closing the issue.

PR #319, opened 2019-08-19
Split large examples

### LINKED ISSUE #312: Split examples into smaller ones
There has been a lot of good intense work lately and as a result of that some examples have grown considerably. The problem is that we do not show diffs when something is not formatted as expected, we show entire "expected/got" files. It works well when files are small, but not so well where they are huge (some of our examples are well beyond 100 lines). It can be hard to understand where the problem is.

This task is about splitting bigger examples (longer than 20 lines I'd say) into smaller ones. It's a good idea to create new sub-directories and name each file in a meaningful way.

utdemir:
Closes #312 .

I splitted some large examples to multiple files. Currently, every example containing more than one declaration is at most 25 lines.

However I am not quite happy with this, see my comment on #312.


mrkkrp:
What happened to `data/examples/declaration/value/function/transform-comprehensions.hs`? It looks like it's just deleted?

utdemir:
> What happened to `data/examples/declaration/value/function/transform-comprehensions.hs`? 

They're renamed to `data/examples/declaration/value/function/comprehension/transform*.hs`.

PR #320, opened 2019-08-19
Conditionally add an extra newline on instances

### LINKED ISSUE #170: Newline after instance declaration
```
instance A.FromJSON Service where
  parseJSON =
    A.genericParseJSON $
      A.defaultOptions
        { A.fieldLabelModifier = A.camelTo2 '_' . drop 2
        }
```

gets formatted to

```
instance A.FromJSON Service where

  parseJSON =
    A.genericParseJSON $
      A.defaultOptions
        { A.fieldLabelModifier = A.camelTo2 '_' . drop 2
        }
```

Is this intentional? This is of course subjective, but the extra newline makes it look like `instance A.FromJSON Service where` is a separate thing from `parseJSON`, which it is not.

utdemir:
Closes #170 .

This implements @mrkkrp's suggestion on the ticket of only adding a trailing newline after the instance header when there are multiple declarations.

### DIFF
--- a/data/examples/declaration/instance/associated-data-out.hs
+++ b/data/examples/declaration/instance/associated-data-out.hs
@@ -3,7 +3,6 @@
 instance Foo Int where data Bar Int = IntBar Int Int
 
 instance Foo Double where
-
   newtype
     Bar
       Double
--- a/data/examples/declaration/instance/contexts-comments-out.hs
+++ b/data/examples/declaration/instance/contexts-comments-out.hs
@@ -14,5 +14,4 @@ instance
            )
          )
   where
-
   readsPrec = undefined
--- a/data/examples/declaration/instance/contexts-out.hs
+++ b/data/examples/declaration/instance/contexts-out.hs
@@ -6,7 +6,6 @@ instance
     )
   => Ord (a, b)
   where
-
   compare _ _ = GT
 
 instance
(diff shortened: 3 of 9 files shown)


mrkkrp:
I think my explanation in #170 was not clear enough, let me explain. I meant that the newline should be inserted only in cases when there are other newlines to be inserted between declarations. There are may be signatures that complicate the logic:

```haskell
instance Foo Int where
  foo x = x + 1

instance Foo Int where

  foo :: Int -> Int
  foo x = x + 1
```

This is the formatting with this PR. Should the newline in the second example be there? I'd guess not. The check I had in mind that we somehow should use `separatedDecls` to figure out if there will be any newlines separating groups of declarations and then only in that case we should insert the newline at the top.

utdemir:
You are right, I missed that part of your comment.

However, I think if there are multiple declarations, I'd prefer to have the newline after instance if there are multiple declarations, even if they are relevant to each other. I think I'd appreciate the newline for a case like:

```haskell
instance Foo Int where
  foo :: Int -> Int
  foo x = x + 1
  {-# INLINE foo #-}
```

I will still do the change to get `separatedDecls` into account, feel free to revert it if you want to preserve the newline.

utdemir:
This PR and #327 is conflicting, so we have to modify the either one depending on which one will be merged first.

mrkkrp:
Now we need the same sort of handling for type class definitions, I guess.

PR #321, opened 2019-08-19
Align lambda arguments

### LINKED ISSUE #282: Rendering of lambda with multi-line argument list is quite ugly
This example

```haskell
foo =
  prop "is inverse to closure" $ \(f :: StaticPtr (Int -> Int))
                                  (x :: Int) ->
    (unclosure . closure) f x == deRefStaticPtr f x
```

gets formatted as

```haskell
foo =
  prop "is inverse to closure" $ \(f :: StaticPtr (Int -> Int))
  (x :: Int) ->
      (unclosure . closure) f x == deRefStaticPtr f x
```

which is a bit confusing. I think indentation of the `(x :: Int)` bit should be bumped one level?

utdemir:
Closes #282.

### DIFF
--- a/data/examples/declaration/value/function/lambda-multi-line1-out.hs
+++ b/data/examples/declaration/value/function/lambda-multi-line1-out.hs
@@ -9,6 +9,7 @@ bar x = \y ->
     else 12
 
 foo =
-  prop "is inverse to closure" $ \(f :: StaticPtr (Int -> Int))
-  (x :: Int) ->
-      (unclosure . closure) f x == deRefStaticPtr f x
+  prop "is inverse to closure"
+    $ \(f :: StaticPtr (Int -> Int))
+       (x :: Int) ->
+        (unclosure . closure) f x == deRefStaticPtr f x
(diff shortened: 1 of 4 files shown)


data/examples/declaration/value/function/lambda-multi-line-out.hs
```diff
@@ -23,5 +23,5 @@ tricky2 =
 
 foo =
   prop "is inverse to closure" $ \(f :: StaticPtr (Int -> Int))
-  (x :: Int) ->
+                                  (x :: Int) ->
```
mrkkrp: It looks like this one is against "minimal diffs" strategy. Imagine that the prefix part

```
prop "is inverse to closure"
```

is changed so that its length has changed. In that case all lambda arguments will be shifted, i.e. we have a case of diff amplification. I'd imagine that we also should insert a newline after `$` and before `\(...)` part so they go on the next line and one indentation level to the right.

utdemir: That is a good point. I changed the logic to not hang lambdas if the argument list spans multiple lines. 

However it behaves slightly different than how you mentioned; the newline is before the `$` rather than after. Now we get:

```haskell
foo =
  prop "is inverse to closure"
    $ \(f :: StaticPtr (Int -> Int))
       (x :: Int) ->
        bar
```

I think this behaviour makes more sense since we usally place operators at the beginning of the line.

What do you think?

PR #325, opened 2019-08-22
Be more conservative when inserting braces

### LINKED ISSUE #323: Redundant braces around nested `where` definitions
Input and expected output:
```hs
a = b
  where
    c = d
      where
        e = f
```

Actual output:
```hs
a = b
  where
    c = d
      where
        { e = f }
```

utdemir:
Closes #323.

> When there is a single local bind on the next line after "where",
> the located combinator thinks we are in a single line layout, since
> the span on the AST does not cover the string "where". Instead, we
> inherit the parent layout while keeping the `located` combinator for
> comments etc.

I don't like exposing 'getLayout' combinator, but that was the simplest
way. I'm happy to change it if you have a better way for achieving that.


src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -231,7 +231,13 @@ p_match' placer pretty style isInfix strictness m_pats m_grhss = do
           breakpoint
           txt "where"
           unless whereIsEmpty breakpoint
-          inci $ located grhssLocalBinds p_hsLocalBinds
+          -- When there is a single local bind on the next line after "where",
```
mrkkrp: I do not understand this solution. The present behavior seems to be correct: if the "body" (that is, the expression) of `where` is single-line (which is true for the inner block in our case) the `located` combinator should set single-line layout, which it does. Artificially setting it to multi-line is not a good idea, even though on the surface it appears to solve the problem, now the single-line expression inside `where` has multi-line layout. In practice, so far it doesn't break anything because we use sub-spans to carefully control the inner layout, but in future it can cause bugs because developers will be thinking that when they have `x = y` the layout of the entire expression is single line, and their assumption might be broken.

There is a similar case for `let` bindings, I added an example for that too. Ideally both cases should have the same fix.

Looking at `sepSemi`, we can see that there are two conditions for using braces:

* The layout is single-line.
* The printing expression is inside of `useBraces`.

Now if you agree that `e = f` by itself, as a whole, is still a single-line construct the question is rather should it be inside of `useBraces` in this case and should the code controlling insertion of braces be more complex?

utdemir: I understand that `foo = bar` is single-line, however in a layout like below,

```
where
  foo = bar
```

I would say that it would be okay to assume that where bindings are multiline. I think it is fair to consider `where <bindings>` as a conceptual unit to decide on the layout rather than just the `<bindings`>. 

But. you are right that it's not a great fix; it just happens to solve this specific issue. I am updating the PR with an hopefully more general fix.

utdemir:
@mrkkrp I reverted the commit you've commented on. Instead I changed the code to be a bit more conservative when inserting braces.

I checked some packages on Hackage and it looks like it is still emitting syntactically correct code. 

When I think about it, I probably should've opened a new PR. Let me know if you prefer that.

src/Ormolu/Printer/Meat/Declaration/Value.hs
mrkkrp: I just want to clarify what I meant: the whole thing (with `where`) is multiline, but when we descend into the first equation, it is, by itself, is single line and should not be set to multiline artificially.

PR #326, opened 2019-08-22
Use build-tool-depends instead of build-tools

utdemir:
It looks like [using build-tools was deprecated](https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-build-tool-depends) since Cabal version [2.0](https://github.com/haskell/cabal/blob/1d2ff3450ee7fd99e89545e2468bf8f512b6e00a/Cabal/doc/file-format-changelog.rst#id55). This commit changes "build-tools" usage to "built-tool-depends" with the correct syntax.

However, there is an issue within "nix-shell" that `cabal new-test` can not resolve "hspec-discover". I couldn't pinpoint the exact issue, but it works outside of "nix-shell"; and old-style commands also keep working.

Before we merge this, I would like to figure out why it doesn't work in "nix-shell". Please let me know if anyone knows why.

### DIFF
--- a/ormolu.cabal
+++ b/ormolu.cabal
@@ -130,7 +130,7 @@ test-suite tests
                     , path           >= 0.6 && < 0.7
                     , path-io        >= 1.4.2 && < 2.0
                     , text           >= 0.2 && < 1.3
-  build-tools:        hspec-discover >= 2.0 && < 3.0
+  build-tool-depends: hspec-discover:hspec-discover >= 2.0 && < 3.0
   other-modules:
                       Ormolu.Printer.CombinatorsSpec
                     , Ormolu.Parser.PragmaSpec


luc-tielen:
Maybe version bounds for hspec-discover are not correct with what is available inside the nix store?

utdemir:
That would make sense, but I don't think it is the case. `ghc-pkg --list` returns me `(hspec-discover-2.6.1)`, which is within the version bounds on the cabal file (`>= 2.0 && < 3.0`).

While checking that, I noticed is that `hspec-discover` is shown differently on the `ghc-pkg` output:

```
$ ghc-pkg list|grep hspec-discover -A 3 -B 3
    hpc-0.6.0.3
    hspec-2.6.1
    hspec-core-2.6.1
    (hspec-discover-2.6.1)
    hspec-expectations-0.8.2
    integer-gmp-1.0.2.0
    integer-logarithms-1.0.2.2
```

Looking at `ghc-pkg`'s source, [it looks like](https://github.com/ghc/ghc/blob/24782b89907ab36fb5aef3a17584f4c10f1e2690/utils/ghc-pkg/Main.hs#L1562) it uses parentheses when the package is not broken or exposed.

I think what is happening is that, Nix understands that that tool is required and adds the binaries to the `$PATH` but somehow adds the package to the package-db differently; then `cabal` does not like that one for some reason and either tries to download build the package, or fails trying to figure out a build plan.

By the way, everything works fine in `nix-build`, only `cabal new-build --enable-tests` inside `nix-shell` is the problem.

mrkkrp:
What is the status of this one? FTR, both `cabal new-test` and `cabal new-build --enable-tests` works inside of `nix-shell` for me. Also note that since this feature was introduced in version 2.0 of cabal, you probably need to bump [this field](https://github.com/tweag/ormolu/blob/ud%2Fuse-build-tool-depends/ormolu.cabal#L3) because its purpose is to specify which version of the Cabal your configuration is compatible with.

Could it be that you keep 1.18 there and thus Cabal just emits a warning:

> build-tool-depends was added in Cabal 2.0, and it will be ignored (with a warning) with old versions of Cabal.

and ignores the field, thus you end up without `hspec-discover` and the tests cannot run?

Even if so, I do not know why it works for me but not for you.

mrkkrp:
I rebased on `master`.

mrkkrp:
Going to close for now. If you feel like we should re-open, please do.

PR #327, opened 2019-08-25
Group declarations referring to multiple names

### LINKED ISSUE #104: Multiple functions in single signature results in odd behaviour
I have two cases here, the first a bit less contrived than the second:
```
fst, snd :: Bool
(fst, snd) = (True, False)

fst, snd :: Bool
fst = True
snd = False
```
which gives:
```
fst, snd :: Bool

(fst, snd) = (True, False)

fst, snd :: Bool
fst = True

snd = False
```
The first should definitely not have the newline. The second is clearly more tricky, but I actually think we should keep them together if that's how people choose to write it.

utdemir:
Closes #104.

This PR extends the declaration grouping logic to handle declarations referring to multiple names. eg:

```haskell
fst, snd :: Bool
(fst, snd) = (True, False)

fst, snd :: Bool
fst = True
snd = False
```

It makes the logic a bit more complex since now we can not decide whether to separate two declarations solely by themselves. This new method tries to separate given declarations to groups of relevant declarations; and then prints the groups together. It simply:

1. Takes the first declaration and calls it a group "header"
2. Keeps adding the next declaration to the group if it is relevant to either to the group header or to the previous element.

The relevancy logic(previously `separatedDecls`, now `groupedDecls`) is pretty much the same, but now it handles declarations binding multiple names, and groups them together if they share at least one name.

### DIFF
--- a/data/examples/declaration/value/function/case-single-line-out.hs
+++ b/data/examples/declaration/value/function/case-single-line-out.hs
@@ -3,5 +3,4 @@ foo x = case x of x -> x
 
 foo :: IO ()
 foo = case [1] of [_] -> "singleton"; _ -> "not singleton"
-
 foo = case [1] of { [] -> foo; _ -> bar } `finally` baz
--- a/data/examples/declaration/signature/type/multi-value-out.hs
+++ b/data/examples/declaration/signature/type/multi-value-out.hs
@@ -1,6 +1,10 @@
 foo, bar :: Int
+foo = 1
+bar = 2
 
 foo,
   bar,
   baz
     :: Int
+bar = 2
+baz = 3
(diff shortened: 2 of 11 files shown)


data/examples/declaration/value/function/case-single-line-out.hs
```diff
@@ -3,5 +3,4 @@ foo x = case x of x -> x
 
 foo :: IO ()
 foo = case [1] of [_] -> "singleton"; _ -> "not singleton"
-
 foo = case [1] of { [] -> foo; _ -> bar } `finally` baz
```
utdemir: This change is caused by the fact that the second `foo` definition is grouped together with the type signature.

PR #332, opened 2019-08-26
Fix idempotence caused by wandering comments

### LINKED ISSUE #273: Idempotence is broken for an example involving operators
Discovered while fixing #260. Here is an example derived from existing tests:

```haskell
foo =
  bar
  ++
    {- some comment -}
    case foo of
      a -> a
```

this becomes after first run:

```haskell
foo =
  bar
    ++ {- some comment -}
    case foo of
      a -> a
```

if we run Ormolu once again however, we get:

```haskell
foo =
  bar
    ++ case foo of {- some comment -}
      a -> a
```

mrkkrp:
Close #273, close #324.

TODO:

- [x] Rebase on current `master`.
- [x] Make the updated/new primitives work correctly.
- [x] Document the new primitives/changes.
- [x] Make sure all tests pass/new tests are reasonable.
- [x] Go through all the code and make sure we take advantage of the improved approach to emitting `space`. This should simplify the code in quite a few places.
- [x] Explain the changes in commit message and here.


mrkkrp:
# Explanation

Previously, if an operator had preceding comments attached to its second argument, they would end up printed right after the operator:

```
a
  + -- b comment
    b
```

On second run however, the comment would be interpreted as attached to `(+)` and the result would be:

```
a
  + b -- b comment
```

Breaking the idempotence guarantees.

The solution that this commit implements includes several steps:

* Introduce the concept of “dirty line”. A line is dirty if it has something on it that can have a comment attached to it.

* `txt` is supposed to output fixed bits of syntax that cannot have comments attached to them (at least in Ormolu's model).

* `atom` on the other hand outputs things that mark the current line dirty.

* When we're to print preceding comments for the second argument we check if the current line is dirty. If it is, we output an extra newline to prevent the first comment from changing “hosts”.

* Now there is another problem with trailing whitespace after the operator in that case. We solve that by making spaces a bit “lazy”. When the `space` combinator is used (which is the recommeneded way to separate different constructs now) it just guarantees that the next thing we'll output on the  same line will be separated from previous output by a single space. So, using `space` twice results in single space in output still. This has the extra benefit of simplifying all the logic that made sure that we have only single space and not 0 or 2 spaces when spaces are inserted conditionally and independently.

PR #334, opened 2019-08-27
Assert that Ormolu is formatted with current version of Ormolu

### LINKED ISSUE #269: Format Ormolu's source code with Ormolu
We already can do it, but it's a bit early to do because the style is still unstable. However, at some point (I'm guessing before releasing version 0.0.2.0) we could do this. The style then should be enforced on CI level.

Also, there should be an easy way (a script?) to re-format the entire code base.

mrkkrp:
Close #269. I'm not ashamed of its output anymore.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
           name: Check formatting
           command: |
             ./format.sh
-            git diff --exit-code --color=always
+            nix-shell -p "(import ./nix/nixpkgs/default.nix).git" --command "git diff --exit-code --color=always"
       - run:
           name: Generate Haddocks
           command: nix-build --attr ormolu.doc
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,11 @@ jobs:
           name: Build the package (includes running the tests)
           command: nix-build --keep-going
       - run:
+          name: Check formatting
+          command: |
+            ./format.sh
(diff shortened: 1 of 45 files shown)


.circleci/config.yml
```diff
@@ -29,6 +29,12 @@ jobs:
       - run:
           name: Build the package (includes running the tests)
           command: nix-build
+      - run:
+          name: Check formatting
+          command: |
+            result/bin/ormolu --mode inplace $(find src -type f -name "*.hs")
+            result/bin/ormolu --mode inplace $(find tests -type f -name "*.hs")
```
utdemir: This would probably match test fixtures, which are not formatted. We should probably only apply it to `*-out.hs` files. Or we can just remove this line because tests themselves ensure that `-out.hs` files are formatted correctly.

mrkkrp: Fixtures are in `data/`.

utdemir: Oh, yes, my mistake, sorry.

mrkkrp:
I'm going to suspend this till version 0.0.2.0 because there is a place that suffers from idempotence issue and so it seems too early to require Ormolu be formatted with Ormolu on CI.

mrkkrp:
This one is good, but blocked on #336 in my opinion.

mrkkrp:
Not to forget: this also should format `*.hs-boot` files.

PR #341, opened 2019-08-28
Use GHC 8.6.5

mrkkrp:
As the first release is approaching, we better switch to the ~~latest~~ GHC 8.6.5.


mrkkrp:
Fails because `cachix` is broken. Will try to figure out a solution.

Issue #343, opened 2019-08-29
Support `EViewPat`


utdemir:
Example input:

```
foo = ('a' -> 'a')
```

Ormolu:

```
*** Exception: not implemented yet: EViewPat
```

I do not know what it means or which language extension enables that syntax. It is not `-XViewPatterns`. Also, it looks like that extension is stealing some syntax, because I got the error when parsing [a file](https://github.com/faylang/fay/blob/7f8c13bb20d70273f03d804939f16022d3afb39a/src/tests/Test/Convert.hs#L67) where `→` is used as an infix operator rather than as a pattern. Probably disabling that extension will fix the issue.

utdemir:
It looks like those constructs are a side effect of some ambiguities in Parser. As far as I understand, patterns are initially parsed as expressions, and those expressions are turned to patterns. However if you use a pattern construct as an expression, they just stay there until the renamer, then GHC rejects them.

And this issue is actually fixed on upstream with [this commit](https://gitlab.haskell.org/ghc/ghc/commit/52fc2719b93ab39be3e52eba531ee173b9134183). However GHC 8.6.5 does not have it.

So, the programs Ormolu fail are not syntactically correct. But it is unfortunate that users see an Ormolu error rather than a parse error. I think the best solution is to just replace our error message to `error "pattern expression in expression context"`. I will send a PR for that.

utdemir:
Another alternative would be to just pretty print them as patterns; and let GHC reject it afterwards.

PR #345, opened 2019-08-29
Remove UnicodeSyntax from default extensions

utdemir:
Without `-XUnicodeSyntax`, this is a working Haskell program:

```
main = print ('a' → 'a')
  where
    (→) = (,)
```

However with `-XUnicodeSyntax`, `→` is interpreted as function arrrow (`->`) and it causes a parse error.

Therefore, this PR removes `UnicodeSyntax` from default extensions since it makes Ormolu fail on valid Haskell programs.

We can successfully format `fay` with this PR.

### DIFF
--- a/data/examples/declaration/signature/type/unicode-out.hs
+++ b/data/examples/declaration/signature/type/unicode-out.hs
@@ -1,0 +1,4 @@
+{-# LANGUAGE UnicodeSyntax #-}
+
+foo :: forall a. Show a => a -> String
+foo = const ()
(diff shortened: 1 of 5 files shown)


mrkkrp:
Does it mean that this resolves #343?

utdemir:
@mrkkrp Just posted a comment on that issue. It does solve the issue on `fay`, but there is another issue uncovered there. I think we still need to apply a separate fix for #343.

PR #349, opened 2019-08-30
Automatically test some Hackage packages on CI

### LINKED ISSUE #303: Assert clean run on various packages on CI

utdemir:
closes #303 .

This PR introduces a `hackageTests` attribute, which, when evaluated runs `ormolize` on a few hand-picked Haskell packages and fails on errors.

While doing that, I also:

* Added a `doCheck` parameter to `ormolize` which causes the build to fail in case of issues.
* Passed `--keep-going` parameter to `nix-build` on CI, so that the first failed build won't immediately stop every other one (the whole command will still fail).
* Commented out some packages with known issues, so we can uncomment them once we fix the issues.

### DIFF
--- a/nix/ormolize/default.nix
+++ b/nix/ormolize/default.nix
@@ -1,15 +1,22 @@
-{ pkgs, haskellPackages }: _: p:
+{ pkgs, haskellPackages }: { package, doCheck ? false }:
   pkgs.stdenv.mkDerivation {
-    name = p.name + "-ormolized";
-    src = p.src;
+    name = package.name + "-ormolized";
+    src = package.src;
     buildInputs = [
       haskellPackages.cpphs
-      (pkgs.haskell.lib.dontCheck haskellPackages.ormolu)
+      (if doCheck
+        then haskellPackages.ormolu
+        else pkgs.haskell.lib.dontCheck haskellPackages.ormolu)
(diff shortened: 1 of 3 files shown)


nix/ormolize/default.nix
```diff
@@ -1,15 +1,20 @@
...
+{ pkgs, haskellPackages }: { package, doCheck ? false }:
   pkgs.stdenv.mkDerivation {
-    name = p.name + "-ormolized";
-    src = p.src;
+    name = package.name + "-ormolized";
+    src = package.src;
     buildInputs = [
       haskellPackages.cpphs
-      (pkgs.haskell.lib.dontCheck haskellPackages.ormolu)
+      haskellPackages.ormolu
```
utdemir: If we leave `dontCheck` here then CI will build ormolu twice; once with tests enabled, once with disabled.

mrkkrp: I think this is OK. The reason we have `donetCheck` here is that we can try how things work without running tests on every rebuild, this is very annoying.

PR #353, opened 2019-09-04
Operator fixity handling

### LINKED ISSUE #315: Figure out a way to deal with precedence of operators
To achieve good formatting of code with many operators with different precedence it looks like we'll need to know precedence of every operator.

So far the alternatives I know about are:

* Hard-coding knowledge of precedence of certain operators (Hindent).
* Making this info part of a config (Birttany). Will be difficult for us, as we don't have any config!
* Trying to learn from GHC actual info about operators. Probably complicated and won't allow us to format a file without knowing anything about its project and environment, which is a handy thing.
* Try to figure out precedence from original formatting. Needs investigation.

utdemir:
Close #315.

This PR is just a WIP to work on operator precedence handling.

It currently contains a function to apply known fixities to a tree of operators. For now, I am using hard-coded list of fixities borrowed from HSE and HLint's codebases. In the final version of this PR, we are supposed to have a way to automatically infer fixities from the code layout. Alternatively, we can use a mix of approaches where we use hard-coded precedences for Prelude functions and try to infer the ones coming from external libraries.

It introduces a data structure called `OpTree`, which represents a tree of expressions and operators. The reason for the data structure is to be able to handle expressions and types in the same way. The alternative is to accept the AST alongisde with a bunch of accessors; however I think a dedicated data structure is clearer.

As I said before, this PR is not ready yet, missing pieces are

- [x] Apply the same logic to type operators.
- [x] Decide & implement how we want to infer fixities from the code layout.
- [x] Prettify the code and add some unit tests for 'reassociate' function.

### DIFF
--- a/data/examples/declaration/type-synonyms/multi-line-out.hs
+++ b/data/examples/declaration/type-synonyms/multi-line-out.hs
@@ -14,8 +14,5 @@ type Foo
 
 type API
   = "route1" :> ApiRoute1
-      :<|> "route2"
-      :> ApiRoute2 -- comment here
-      :<|> OmitDocs
-      :> "i"
-      :> ASomething API
+      :<|> "route2" :> ApiRoute2 -- comment here
+      :<|> OmitDocs :> "i" :> ASomething API
--- a/data/examples/declaration/value/function/infix/lenses-out.hs
+++ b/data/examples/declaration/value/function/infix/lenses-out.hs
@@ -1,11 +1,20 @@
 lenses =
   Just $ M.fromList
-    $ "type"
-    .= ("user.connection" :: Text)
-    # "connection"
-    .= uc
-    # "user"
-    .= case name of
-      Just n -> Just $ object ["name" .= n]
-      Nothing -> Nothing
-    # []
+    $ "type" .= ("user.connection" :: Text)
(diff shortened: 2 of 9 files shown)


utdemir:
I think this is ready for review.

I implemented a custom logic where we assign a score to every occurance of an operator based on their location, and the average of that score determine the fixity of the operator.

As you can imagine, the solution is a bit brittle; and it is easy to mislead it if you knowingly craft an input, but it gave acceptable results for every code snippet I found online. And since it returns the same AST no matter how we infer the fixities, it is not the end of the world if we infer something incorrectly.

Interestingly, almost every example I found online was quite simple, so I didn't add many test cases. If you have a specific test case on your mind I'm happy to add it to the PR.

You can compare it to the version with hard-coded fixities you can revert my second commit.

The code is not really optimised, and I think it has quadratic time complexity. Notably, we use `opTreeLoc` function quite often and it traverses the whole tree every time. Memoizing that on the `OpBranch` constructor would make formatting files with reeeally long operator chains a lot faster. We can do this once we decide to optimize for speed.

mrkkrp:
@neongreen, @int-index could you try to run Ormolu from this branch on your stuff and comment here if you find something wrong with it?

PR #356, opened 2019-09-04
Refactor comment rendering

### LINKED ISSUE #337: Idempotence bug caused by merging of two comments
Consider the following example (taken from `yesod-core`):

```haskell
foo = langs'
  where
    -- The language preferences are prioritized as follows:
    langs' = catMaybes [ lookup langKey gets -- Query _LANG
                       , lookup langKey cookies     -- Cookie _LANG
                       , lookupText langKey session -- Session _LANG
                       ] ++ langs                    -- Accept-Language(s)

    -- Github issue #195. We want to add an extra two-letter version of any
    -- language in the list.
    langs'' = addTwoLetters (id, Set.empty) langs'
```

if we run Ormolu once, we get:

```haskell
foo = langs'
  where
    -- The language preferences are prioritized as follows:
    langs' =
      catMaybes
        [ lookup langKey gets, -- Query _LANG
          lookup langKey cookies, -- Cookie _LANG
          lookupText langKey session -- Session _LANG
          ]
        ++ langs -- Accept-Language(s)
    -- Github issue #195. We want to add an extra two-letter version of any
    -- language in the list.
    langs'' = addTwoLetters (id, Set.empty) langs'
```

however on the second run the `-- Accept-Language(s)` comment merged with the comment about GitHub and this causes the later to be considered as the contiuation of the first. It is thus aligned accordingly:

```haskell
foo = langs'
  where
    -- The language preferences are prioritized as follows:
    langs' =
      catMaybes
        [ lookup langKey gets, -- Query _LANG
          lookup langKey cookies, -- Cookie _LANG
          lookupText langKey session -- Session _LANG
          ]
        ++ langs -- Accept-Language(s)
        -- Github issue #195. We want to add an extra two-letter version of any
        -- language in the list.
    langs'' = addTwoLetters (id, Set.empty) langs'
```

The best solution here seems to be keeping the blank line between the two definitions, however I did not think of how to detect when we should do it.

mrkkrp:
Close #337.

* This removes (or rather puts it to a lower level) logic around “modifying newline” because it was very hard to reason about and almost blocked my work on fixing issue #337.

* I also dropped debugging output because it's too verbose and I'm not using it anyway.

* As part of these changes I also changed now the `newline` combinator works. Now, similar to `space`, the second `newline` in a row just tells the rendering engine to prefix next thing with a newline, using the `newline` combinator more than twice in a row has no effect.

* To take full advantage of the new feature I also went through the code and
simplified some logic around outputting exact amount of newlines because now
it's harder to get things wrong, so we can be less careful with counting
newlines.


mrkkrp:
Remains:

- [ ] I found a snippet which makes Ormolu hung. It should be fixed.
- [ ] Go through the source code and find/simplify places where the new smarter newlines allow us to be less careful.
- [ ] Add good description and squash.

Here is the problematic snippet:

```haskell
newNames :: [(String, String)]
newNames = let (*) = flip (,) in
    ["Control" * "Monad"

    -- Foo

    -- Bar
    ]
```

mrkkrp:
Other than that, this one allows us to enable 4 more packages in the tests, including Ormolu itself, which also unblocks #334.

PR #359, opened 2019-09-04
Remove extra space before type variables with apostrophe

### LINKED ISSUE #354: Leading whitespace in kind signature
I'm formatting the following:

```haskell
{-# LANGUAGE KindSignatures #-}

data Format (k :: *) (k' :: *) (k'' :: *)
```

And the resulting file is:

```haskell
{-# LANGUAGE KindSignatures #-}

data Format (k :: *) ( k' :: *) ( k'' :: *)
```

I'm surprised by the leading whitespace in the second and third kind signature (e.g. `( k' :: *)`). Note that it does not appear in the first one.

I'm using ormolu 10232b9c86f1974a8bf9234076a0f4e0fc0845af which is `master` at the time of redacting this issue.

utdemir:
Closes #354.

We were putting a space before type variables if they contain the symbol `'`, but I think it is only necessary if the type is promoted.

I couldn't think of a case where this breaks; tests seem to pass, including `hackageTests`.

### DIFF
--- a/data/examples/declaration/data/kind-annotations-out.hs
+++ b/data/examples/declaration/data/kind-annotations-out.hs
@@ -1,1 +1,3 @@
 data Something (n :: Nat) = Something
+
+data Format (k :: *) (k' :: *) (k'' :: *)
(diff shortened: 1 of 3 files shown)


src/Ormolu/Exception.hs.bac
```diff
@@ -0,0 +1,96 @@
+{-# LANGUAGE LambdaCase #-}
```
mrkkrp: Is this a backup file?

utdemir: Yep, my mistake.

Issue #355, opened 2019-09-04
Infix pattern synonym: ghc panic


guibou:
The following:

```haskell
{-# LANGUAGE PatternSynonyms #-}

pattern a :< b = (a, b)
```

Leads to ormolu crash:

```
ormolu: panic! (the 'impossible' happened)
  (GHC version 8.6.5 for x86_64-unknown-linux):
	not implemented yet: InfixCon
CallStack (from HasCallStack):
  error, called at src/Ormolu/Utils.hs:31:22 in ormolu-0.0.1.0-KdSq2D84oCzHMfc2HHokyf:Ormolu.Utils

Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
```

I'm using ormolu 10232b9 which is master at the time of redacting this issue.

utdemir:
Besides supporting infix pattern synonyms, another issue is reporting that as a GHC bug. I thought we had some piece of code to prevent that, I will try to figure out why it didn't work.

mrkkrp:
Why should it be reported as a GHC bug though? It’s from working valid core AFAIK.

utdemir:
@mrkkrp Turns out it was just a laziness issue causing the exception to be thrown inside GHC's parser when we check whether the AST differs after formatting.

PR #360, opened 2019-09-05
Don't wrap Ormolu exceptions as GHC errors

utdemir:
Related: #355.

Turns out when `--unsafe` is not set, exceptions thrown in formatting code are catched in the subsequent parsing check, which wraps them as GHC errors adding a text like:

```
Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
```

Which is misleading since the error is not caused by GHC.

This PR avoids it by strictly binding the result of `printModule`.

### DIFF
--- a/src/Ormolu.hs
+++ b/src/Ormolu.hs
@@ -1,5 +1,6 @@
 -- | A formatter for Haskell source code.
 
+{-# LANGUAGE BangPatterns    #-}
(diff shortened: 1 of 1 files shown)


mrkkrp:
It looks like only one bang pattern is necessary here, I dropped the other one. I also added explanation in the code.

PR #363, opened 2019-09-06
Enable checking the ‘tls’ package on CI

mrkkrp:
I noticed that `tls` fails the idempotence check, even though the example I extracted is terribly similar to the problem I just fixed in #356.


mrkkrp:
Not to forget: check if the first idempotence failure in `intero` gets fixed as well.

PR #382, opened 2019-09-15
Introduce a breakpoint after ‘forall’ in type signatures

### LINKED ISSUE #148: Type synonym is reformatted into a less readable form
Before Ormolu:

```
type CoerceLocalSig m m' =
  forall r a.
  LocalSig m  r a ->
  LocalSig m' r a
```

After Ormolu:

```
type CoerceLocalSig m m'
  = forall r a. LocalSig m r a
  -> LocalSig m' r a
```

That's clearly worse because now `LocalSig` isn't aligned. Now, I'm not sure what the style guide Ormolu follows, but if my hand-written version is incompatible with it (because of the trailing `->`), at least keep the `forall` on its own line:

```
type CoerceLocalSig m m'
  = forall r a.
     LocalSig m r a
  -> LocalSig m' r a
```

mrkkrp:
Close #148.

@utdemir I remember you tried to fix this one as well and found it hard to fix properly. I understand that my fix is a bit hacky because of the `txt "   "` bit which I imagine may look very strange in certain situations, but I fail to come up with any such realistic examples. Please take a look at this and tell me what you think.

### DIFF
--- a/data/examples/declaration/rewrite-rule/type-signature-out.hs
+++ b/data/examples/declaration/rewrite-rule/type-signature-out.hs
@@ -7,7 +7,8 @@
   k
   z
   ( g
-      :: forall b. (a -> b -> b)
+      :: forall b.
+         (a -> b -> b)
       -> b
       -> b
     ).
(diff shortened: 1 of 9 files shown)


mrkkrp:
I'm going to merge this for now, but we decided we'll go for a refactoring of type signatures in general.

PR #392, opened 2019-09-17
Change the style of type signatures

### LINKED ISSUE #336: Detect headers in comments and add a newline after them
This one is more experimental. Sometime there are headers like this:

```haskell
-- Helpers

foo :: Int -> Int
```

The blank line between the header and `foo` is currently erased. It would be good to somehow detect the header and make sure there is a newline after it so it doesn't get glued together with the next definition.

### LINKED ISSUE #365: Unpleasant results with an overly indented type signature
Here is a bit from `hakyll` (which should be enabled on CI once this is fixed):

```haskell
tagCloudField :: String
               -- ^ Destination key
               -> Double
               -- ^ Smallest font size, in percent
               -> Double
               -- ^ Biggest font size, in percent
               -> Tags
               -- ^ Input tags
               -> Context a
               -- ^ Context
```

On the first pass we get:

```haskell
tagCloudField
  :: String
  -- ^ Destination key
  -> Double
  -> -- ^ Smallest font size, in percent
  Double
  -> -- ^ Biggest font size, in percent
  Tags
  -> -- ^ Input tags
  Context a
  -- ^ Context
```

Which is not only ugly, but also not idempotent:

```haskell
tagCloudField
  :: String
  -- ^ Destination key
  -> Double
  -- ^ Smallest font size, in percent
  -> Double
  -- ^ Biggest font size, in percent
  -> Tags
  -- ^ Input tags
  -> Context a
  -- ^ Context
```

The solution should ensure that we search the final fix point immediately.

### LINKED ISSUE #389: Change the style of type signatures
It was discussed off-line and we found that we're going to use this style:

```haskell
foo ::
  Foo ->
  Bar ->
  Baz
```

Because it plays nicer with advanced features of GHC Haskell and `forall`s. We'll also need to find a way to re-format `-- ^` Haddocks into `-- |` style and put them before arguments, not after.

mrkkrp:
Close #389, close #365, close #336, close #375.


bjornbm:
Should the example in `DESIGN.md` be updated as well?

mrkkrp:
@bjornbm Yes, we can do that once the PR is finished.

bjornbm:
@mrkkrp Ah, yes, of course, as you prefer. (I assumed you would nominally include the documentation update in this PR.)

mrkkrp:
I will, what I meant is that the change is not finished yet.

mrkkrp:
Not to forget: this change might have some potential for fixing #336 as well.

mrkkrp:
- [ ] Fix the remaining two test suite failures.
- [ ] Implement `IEDocNamed` and whatever is necessary to make all package that we check on CI to pass. Add relevant tests to the test suite.
- [ ] #365 should just work at this point, add a test for it and enable `hakyll`.
- [ ] Inspect how various packages are formatted to check that no funny business is going on.
- [ ] Review the PR and make sure it's in a good shape.
- [ ] #336 should be easy to make work (not strictly necessary though): Haddocks should stick to whatever they are attached to, everything else should be separated by a blank line when it's on the top level.

mrkkrp:
Now, that was a big one.

int-index:
This looks lovely. Thanks!

PR #397, opened 2019-10-02
Keep the ($) operator on the same line in certain cases

### LINKED ISSUE #390: Leave application operator ($) on the first line in multi-line expressions
Currently a multi-line expression involving `($)` is reformatted like this:

```haskell
Foo $
  x + y

-- becomes:

Foo
  $ x + y
```

Which is arguably harder to read. It could be nicer to special-case `($)` and leave it on the first line.

mrkkrp:
Close #390.

### DIFF
--- a/data/examples/declaration/value/function/infix/applicative.hs
+++ b/data/examples/declaration/value/function/infix/applicative.hs
@@ -1,6 +1,4 @@
-
 f =
   Foo <$> bar
       <*> baz
       <*> baz'
-
--- a/data/examples/declaration/value/function/infix/comments-out.hs
+++ b/data/examples/declaration/value/function/infix/comments-out.hs
@@ -1,10 +1,8 @@
 main =
-  bar
-    $ baz -- bar
-    -- baz
+  bar $ -- bar
+    baz -- baz
 
-bar
-  $
+bar $
   {- foo
    -}
   bar
--- a/data/examples/declaration/value/function/infix/dollar-chains-out.hs
+++ b/data/examples/declaration/value/function/infix/dollar-chains-out.hs
@@ -1,0 +1,18 @@
+module Main where
+
+foo =
+  fmap escapeLeadingDollar
+    . dropPaddingSpace
+    . dropWhileEnd T.null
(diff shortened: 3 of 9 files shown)


aspiwack:
May I enquire in which certain cases will the `($)` be hanging?

mrkkrp:
When it has a an expression which has "hanging form" (`case`, `do`-block, etc.) on its right-hand side. `($)` will be placed on next line when it's part of a multi-line construction such as this:

```haskell
  where
    r =
      fmap escapeLeadingDollar
        . dropPaddingSpace
        . dropWhileEnd T.null
        . fmap (T.stripEnd . T.pack)
        . lines
        $ unpackHDS docStr
```

In other cases it'll be kept on the same line and breakpoint will be placed after the `($)` operator:

```haskell
  when (GHC.xopt Cpp dynFlags && not cfgTolerateCpp) $
    throwIO (OrmoluCppEnabled path)
```

instead of (current behavior):

```haskell
  when (GHC.xopt Cpp dynFlags && not cfgTolerateCpp)
    $ throwIO (OrmoluCppEnabled path)
```

aspiwack:
Did I understand correctly?

The `($)` will be hanging if:

- Its right-hand side starts with a hanging form
- Or, its left hand side is single-line

Can it happen that a sequence with multiple `($)` will have some hanging `($)` and some line-starting `($)`? Can we come up with an example to add to the test suite?

mrkkrp:
This:

```haskell
foo =
  bar $
  baz $
  quux
```

gets reformatted into this:

```haskell
foo =
  bar $
    baz
    $ quux
```

What happens here is that `bar` is single line but `bar $\n baz` is multi-line.

aspiwack:
I suggest having this somewhere in the test suite, so that we are conscious of this behaviour.

This particular does look a tad problematic to me, to be honest. But what matters most, is that the behaviour is recorded and tested for.

mrkkrp:
I think there should be another condition so we get:

```haskell
foo =
  bar
    $ baz
    $ quux
```

in that case. I'll take a look this evening.

PR #402, opened 2019-10-03
Update the list of goals in the readme

mrkkrp:
As discussed, we found another design goal. I added it to the readme. Also, keeping `DESIGN.md` in sync with the development is not worth the effort. The formatter now is fully implemented (modulo bugs fixes and keeping it up to date with new GHC versions), so arguably there is no longer need for a
design document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,20 +11,19 @@ the following goals in mind:
 
 * Using GHC's own parser to avoid parsing problems caused by
   [`haskell-src-exts`][haskell-src-exts].
-* Let some whitespace be programmable. The layout of the input
-  influence the layout choices in the output. This means that the
-  choices between single-line/multi-line layouts in each particular
-  situation are made by the user, not by an algorithm. This makes the
-  implementation simpler and leaves some control to the user while
-  still guaranteeing that the formatted code is stylistically
-  consistent.
-* Writing code in such a way so it's easy to modify and maintain. Roughly,
-  it means that the project follows the path of [`hindent`][hindent] and is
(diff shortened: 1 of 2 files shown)


curiousleo:
As someone just browsing through this repo, I thought that the design document was an excellent place to start with to understand the motivation of the project, the "prior art" and big design decisions. Seeing it added a lot of credibility to the project: the document made it very obvious that plenty of thought and research had gone into making `ormolu`.

I see that it is annoying to keep it up to date. I want to suggest two alternatives to just deleting the design document:

- Delete it, but keep a (reasonably prominent) link to `DESIGN.md` as it was just before the deletion commit in `README.md`, making it clear that it is the "original" or "no longer updated" design doc.
- Keep `DESIGN.md` but add a big bold caveat at the top saying "This document is no longer maintained".

README.md
```diff
@@ -27,6 +27,10 @@ the following goals in mind:
 * Idempotency: formatting already formatted code doesn't change it.
 * Be well-tested and robust to the point that it can be used in large
   projects without exposing unfortunate, disappointing bugs here and there.
+* Make stylistic choices according to the direction of evolution of the
```
mboes: We can state the same thing more succinctly:

> Choose a style compatible with modern dialects of Haskell. As new Haskell extensions enter broad use, we may change the style to accomodate them.

mrkkrp:
Let's just drop the second sentence so for the third goal we'll have just:

> Writing code in such a way so it's easy to modify and maintain.

mboes:
SGTM

PR #423, opened 2019-10-27
Migrate to ghc-lib-parser

### LINKED ISSUE #344: Switch to GHC 8.8.1
New version has been released. Once `nixpkgs` come with this version of GHC as the default and LTS switches to it we can switch as well.

### LINKED ISSUE #422: Use ghc-lib-parser instead of GHC
This would let us have features from GHC 8.8.1 while still using GHC 8.6.5 from Nixpkgs.

I can make a patch.

neongreen:
Fixes #422, fixes #344.

## Note to the reviewer

This can be reviewed commit-by-commit.

## TODO

- [x] Fix the way I work with `XPat`.
- [x] Run nixpkgs tests.
- [x] Check that the executable is buildable.
- [x] Fix warnings.
- [x] Format with Ormolu.
- [x] Add a test for:

  ```
  type family G a b where
    forall x y. G [x] (Proxy y) = Double
    forall z.   G z   z         = Bool
  ```

- [x] Implement support for the point above.

- [x] Investigate:

   ```
   {-# LANGUAGE TypeFamilies, TypeOperators, NoStarIsType #-}

   instance PNum Nat where
     type a * b = ()
   ```

- [x] Go through singletons and look for more regressions.

- [x] Add a test and implementation for:

   ```
   {-# RULES "example" forall a. forall (x :: a). id x = x #-}
   ```

- [x] Add tests for:

   ```
   type family a ! b
   type family a . b
   ```

- [x] Add paren tests:

   ```
   class a + b
   data D1 = forall a b. (a + b) => D1 a b
   data D2 = forall a b.  a + b  => D2 a b -- now allowed

   data T = MkT1 { a :: {-# UNPACK #-} (Maybe Int && Bool) }
          | MkT2 { a :: {-# UNPACK #-}  Maybe Int && Bool  } -- now allowed

   data G where
     MkG1 :: {-# UNPACK #-} (Maybe Int && Bool) -> G
     MkG2 :: {-# UNPACK #-}  Maybe Int && Bool  -> G  -- now allowed

   Proxy '(a :: A, b :: B)
   ```

- [x] Try out a rule with two levels of multiline foralls.

- [x] Fix regression: the forall should not be broken across several lines:

   ```haskell
   sEither ::
     forall a b c (l :: a ~> c) (r :: b ~> c) (e :: Either a b).
     (forall n. Proxy l -> Sing n -> Sing (Apply l n)) ->
     (forall n. Proxy r -> Sing n -> Sing (Apply r n)) ->
     Sing e ->
     Sing (Either_ l r e)
   sEither l _ (SLeft x) = l Proxy x
   sEither _ r (SRight x) = r Proxy x
   ```

- [x] Add a test for an (accidental?) improvement: the `:<>:` declaration should remain on the same line:

   ```haskell
   data ErrorMessage' s
     = -- | Show the text as is.
       Text s
     | -- | Pretty print the type.
       -- @ShowType :: k -> ErrorMessage@
       forall t. ShowType t
     | -- | Put two pieces of error message next
       -- to each other.
       ErrorMessage' s :<>: ErrorMessage' s
     | -- | Stack two pieces of error message on top
       -- of each other.
       ErrorMessage' s :$$: ErrorMessage' s
   ```


neongreen:
Tests pass.

TODO:

- [ ] Fix the way I work with `XPat`.
- [ ] Run nixpkgs tests.
- [ ] Check that the executable is buildable.
- [ ] Fix warnings.

src/Ormolu/Printer/Combinators.hs
```diff
@@ -108,6 +109,27 @@ located' ::
...
+-- | A version of 'located' that works on 'Pat'.
+--
+-- Starting from GHC 8.8, 'LPat' == 'Pat'. Located 'Pat's are always
+-- constructed with the 'XPat' constructor, containing a @Located Pat@.
+--
+-- Most of the time, we can just use 'p_pat' directly, because it handles
+-- located 'Pat's. However, sometimes we want to use the location to render
+-- something other than the given 'Pat'.
+--
+-- If given 'Pat' does not contain a location, we error out.
```
neongreen: I'm rather disappointed with GHC here.

Of course, an alternative option is swallowing the bug and doing `p -> ($ p)` if we get something other than an `XPat`. I'm not sure what is the best course of action here.

src/Ormolu/Printer/Meat/Declaration/Class.hs
```diff
@@ -103,8 +103,9 @@ p_funDep (before, after) = do
 defltEqnToInstDecl :: TyFamDefltEqn GhcPs -> TyFamInstDecl GhcPs
 defltEqnToInstDecl FamEqn {..} = TyFamInstDecl {..}
   where
-    eqn = FamEqn {feqn_pats = tyVarsToTypes feqn_pats, ..}
+    eqn = FamEqn {feqn_pats = map HsValArg (tyVarsToTypes feqn_pats), ..}
```
neongreen: This is questionable. I don't have enough experience with GHC internals to say whether it's the right thing to do.

src/Ormolu/Printer/Meat/Declaration/Rule.hs
```diff
@@ -24,7 +24,7 @@ p_ruleDecls = \case
 
 p_ruleDecl :: RuleDecl GhcPs -> R ()
 p_ruleDecl = \case
-  HsRule NoExt ruleName activation ruleBndrs lhs rhs -> do
+  HsRule NoExt ruleName activation _ ruleBndrs lhs rhs -> do
```
neongreen: Here we are skipping `rd_tyvs :: Maybe [LHsTyVarBndr (NoGhcTc pass)]`. I will try to find an example where `rd_tyvs` is non-empty.

src/Ormolu/Utils.hs
```diff
@@ -72,3 +74,9 @@ splitDocString docStr =
            in if leadingSpace x
                 then dropSpace <$> xs
                 else xs
+
+typeArgToType :: LHsTypeArg p -> LHsType p
+typeArgToType = \case
+  HsValArg tm -> tm
+  HsTypeArg _ ty -> ty
+  HsArgPar _ -> error "typeArgToType: HsArgPar"
```
neongreen:
```
{-
Note [HsArgPar]
A HsArgPar indicates that everything to the left of this in the argument list is
enclosed in parentheses together with the function itself. It is necessary so
that we can recreate the parenthesis structure in the original source after
typechecking the arguments.

The SrcSpan is the span of the original HsPar

((f arg1) arg2 arg3) results in an input argument list of
[HsValArg arg1, HsArgPar span1, HsValArg arg2, HsValArg arg3, HsArgPar span2]

-}
```

This is weird. And we probably don't need to handle it because currying?

neongreen:
```
ormolu --version

ormolu 0.0.1.0 ghc-lib-parser 0d041b3638f30d8b9efba0708e6d2cd321b2a0de
using ghc-lib-parser 8.8.1
```

Not sure if this is the output we want (why `ghc-lib-parser 0d041b3638f30d8b9efba0708e6d2cd321b2a0de`?).

src/Ormolu/Utils.hs
neongreen: > Outdated

Not outdated.

src/Ormolu/Printer/Meat/Declaration/Rule.hs
neongreen: Probably this:

```
{-# RULES "example" forall a. forall (x :: a). id x = x #-}
```

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -54,6 +53,7 @@ p_hsType' multilineArgs = \case
     located f p_hsType
     breakpoint
     inci (located x p_hsType)
+  HsAppKindTy _ _ _ -> notImplemented "HsAppKindTy"
```
neongreen: Can we trigger this one?

src/Ormolu/Utils.hs
```diff
@@ -72,3 +73,9 @@ splitDocString docStr =
            in if leadingSpace x
                 then dropSpace <$> xs
                 else xs
+
+typeArgToType :: LHsTypeArg p -> LHsType p
+typeArgToType = \case
+  HsValArg tm -> tm
+  HsTypeArg _ ty -> ty
+  HsArgPar _ -> notImplemented "HsArgPar"
```
neongreen:
```
{-
Note [HsArgPar]
A HsArgPar indicates that everything to the left of this in the argument list is
enclosed in parentheses together with the function itself. It is necessary so
that we can recreate the parenthesis structure in the original source after
typechecking the arguments.

The SrcSpan is the span of the original HsPar

((f arg1) arg2 arg3) results in an input argument list of
[HsValArg arg1, HsArgPar span1, HsValArg arg2, HsValArg arg3, HsArgPar span2]

-}
```

This is weird. And we probably don't need to handle it because currying?

neongreen:
https://github.com/neongreen/ormolu/blob/ghc-lib-parser/src/Ormolu/Printer/Meat/Type.hs#L84

@mrkkrp do you remember in what cases parentheses "are not always encoded with 'HsParTy'"? I'm asking because GHC 8.8 doesn't require parentheses anymore, but Ormolu still adds them – I'd rather it didn't do it where it's not necessary.

neongreen:
Oh alright, I found an example: `data Something (n :: Nat) = Something`.

neongreen:
All tests pass and it's ready to be reviewed.

neongreen:
I don't know why CI fails, might be a timeout.

data/examples/other/haddock-sections-out.hs
```diff
@@ -0,0 +1,7 @@
+-- $weird #anchor#
```
neongreen: This is a regression – before this commit it was rendering it as `-- $weird#anchor#`. I'm not sure how it got broken.

default.nix
```diff
@@ -15,6 +15,11 @@ let
   };
   ormoluOverlay = self: super: {
     "ormolu" = super.callCabal2nix "ormolu" source { };
+    # Nixpkgs provides ghc-lib-parser-8.8.0.20190424, but we want
+    # ghc-lib-parser-8.8.1. We disable Haddock generation because it's way
```
mrkkrp: Why can't we use `ghc-lib-parser-8.8.0.20190424`?

data/examples/other/haddock-sections-out.hs
mrkkrp: It looks like named sections and [anchors](https://haskell-haddock.readthedocs.io/en/latest/markup.html#anchors) have nothing in common. Why do you think that it is a regression? Is this example from a real code base where such a combination serves some purpose?

src/Ormolu/Printer/Meat/Declaration/Rule.hs
```diff
@@ -43,16 +47,6 @@ p_ruleDecl = \case
 p_ruleName :: (SourceText, RuleName) -> R ()
 p_ruleName (_, name) = atom $ HsString NoSourceText name
 
-p_ruleBndrs :: [LRuleBndr GhcPs] -> R ()
```
mrkkrp: Good job at reducing the duplication!

src/Ormolu/Utils.hs
mrkkrp: I do not know. I'm OK with this as long as everything works in practice.

mrkkrp:
This is good to go but is blocked on the CI issues.

PR #453, opened 2019-11-07
Fix GHC warnings

cblp:


mrkkrp:
Are these GHC warnings or HLint warnings?

cblp:
GHC warnings, a typo fix, and a bit of HLint warnings, sorry. Should I remove them?

mrkkrp:
No, I just asked.

mrkkrp:
Thank you for opening the PR. What is disturbing is that those thing were not caught by CI. In fact this was because I forgot to enable the `dev` flag in Nix. I added the change to avoid the problem in the future. Going to merge when it's green.

PR #455, opened 2019-11-07
Fix HLint warnings

cblp:


src/Ormolu/Exception.hs
```diff
@@ -84,7 +86,7 @@ withPrettyOrmoluExceptions m = m `catch` h
           OrmoluParsingFailed _ _ -> 3
           OrmoluOutputParsingFailed _ _ -> 4
           OrmoluASTDiffers _ _ -> 5
-          OrmoluNonIdempotentOutput _ _ _ -> 6
```
neongreen: Can you change the other branches as well?

src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -36,9 +36,7 @@ p_hsmodImport ImportDecl {..} = do
   space
   when ideclQualified (txt "qualified")
   space
-  case ideclPkgQual of
-    Nothing -> return ()
-    Just slit -> atom slit
+  forM_ ideclPkgQual atom
```
neongreen: Meh. I'd argue this one should just be disabled in `.hlint.yaml`.

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -145,7 +146,7 @@ p_hsType' multilineArgs = \case
   where
     isPromoted = \case
       HsTyVar _ IsPromoted _ -> True
-      HsExplicitListTy _ _ _ -> True
+      HsExplicitListTy {} -> True
       HsExplicitTupleTy _ _ -> True
```
neongreen: `HsExplicitTupleTy` should also be changed then.

src/Ormolu/Exception.hs
cblp: Done

src/Ormolu/Printer/Meat/ImportExport.hs
cblp: Reverted.

Should `.hlint.yaml` be added?

src/Ormolu/Printer/Meat/Type.hs
cblp: Done

src/Ormolu/Printer/Meat/ImportExport.hs
neongreen: I would add it, yeah. Once I start using HLint, I usually end up needing a `.hlint.yaml` almost immediately.

cblp: If this should be done, this should be done a separate PR.

mrkkrp: `.hlint.yaml` should be added if we decide to run Hlint on CI. Otherwise no.

Issue #299, opened 2019-08-15
What is the policy for `if-then-else` clause?


cblp:
This looks ugly:

```haskell
do
  action before
  if condition
  then action 1
  else action 2
  action after
```

Is it intended?

cblp:
In real code, I've seen two styles:
```haskell
do
  action before
  if condition
    then action 1
    else action 2
  action after
```
and
```haskell
do
  action before
  if condition then
    action 1
  else
    action 2
  action after

```

mrkkrp:
I myself align `then` and `else` one level to the right, however I think @mboes and @aspiwack had some argumentation for the current design, so I invite them to recite it here.

mrkkrp:
I'd like to change it to give a bump to `then` and `else` myself, because I agree that the current formatting is ugly.

aspiwack:
I don't use if-then-else in do-blocks, personally (I use `case` systematically). So, no opinion in that context.

Outside of do-block, this problem doesn't occur, so an extra indentation of `then` and `else` is purely gratuitous, and I don't think that we should have gratuitous indentations.

I believe that was all the argument I have. Or had.

mrkkrp:
I think we should do whatever makes it easier to read and be unconditional about formatting inside/outside so-blocks.

cblp:
@aspiwack in this context, is "gratuitous" bad or good? Should we avoid or pursue it?

cblp:
Variant B has some advantages:
1. It works as well in the pure expression syntax context.
2. Looks like if-statement in other languages.
3. Offers exactly 1 indentation level for 1 logical level.

So, we can have one style for all purposes.

PR #457, opened 2019-11-08
Decrease syntax noise by changing if-then-else 2-level layout to 1-level

cblp:
Motivation
----------

Before:

```hs
if condition
  then do
    2nd level of indentation
  else do
    2nd level of indentation
-- 5 lines, 2 of them are purely syntactical, with no semantics
```

After:

```hs
if condition then do
  1st level of indentation
else do
  1st level of indentation
-- 4 lines
```

1. Exactly 1 indentation for 1 code block => Lesser code depth. Currently, the number of indentation levels is twice greater than the number of semantic layers.
2. Closer to if-statement layout in other languages. => More friendly to newcomers with experience in other languages.

Drawbacks (not problems in my opinion)
---------------------------------------

1.  In Haskell98 mode, usage in do-blocks needs NonDecreasingIndentation extension.
    Usage in pure expressions works out of the box.
    In Haskell2010 mode, it works out of the box.

2.  Longer lines in some cases due to ` then` ending.

3.  More lines in short cases:

	```hs
	-- before: 3 lines
	if condition
	  then x
	  else y
	```

    vs.

	```hs
	-- after: 4 lines
	if condition then
	  x
	else
	  y
	```

4.  Reformatting code formatted with ormolu 0.0.1.0. This is not a problem since 0.0.1.0 doesn't commit being stable.

    Compensated by the decrease in longer cases, see example in Motivation section.

### DIFF
--- a/data/examples/declaration/value/function/arrow/proc-do-complex-out.hs
+++ b/data/examples/declaration/value/function/arrow/proc-do-complex-out.hs
@@ -23,8 +23,8 @@ foo
           ( b + 1, -- Funnel into arrow
             d * b
           )
-      if x `mod` y == 0 -- Basic condition
-        then case e of -- Only left case is relevant
+      if x `mod` y == 0 then -- Basic condition
+        case e of -- Only left case is relevant
           Left
             ( z,
               w
@@ -36,23 +36,24 @@ foo
                       -- Just do the calculation
                       (x + y * z)
                   )
-        else do
-          let u = x -- Let bindings bind expressions, not commands
-            -- Could pattern match directly on x
-          i <- case u of
-            0 -> (g . h -< u)
-            n ->
-              ( ( h . g -<
-                    y -- First actual use of y
-                )
+      else do
+        let u = x -- Let bindings bind expressions, not commands
+          -- Could pattern match directly on x
+        i <- case u of
+          0 -> (g . h -< u)
+          n ->
+            ( ( h . g -<
+                  y -- First actual use of y
               )
-          returnA -< ()
-          -- Sometimes execute effects
-          if i > 0
-            then ma -< ()
-            else returnA -< ()
-          returnA -<
-            ( i
-                + x
-                * y -- Just do the calculation
             )
+        returnA -< ()
+        -- Sometimes execute effects
+        if i > 0 then
+          ma -< ()
+        else
+          returnA -< ()
+        returnA -<
+          ( i
+              + x
+              * y -- Just do the calculation
+          )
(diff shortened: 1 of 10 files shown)


cblp:
Hm, CI found a bug in case

```hs
  -- comment
  else do
```

but this mug is in master, is not introduced by my patch.

mrkkrp:
Another subjective issue. Personally, I'd be very uncomfortable with this. I don't think we need to go out of our way to save a level of indentation. Also, I have never seen this formatting in Haskell code. Maybe in JavaScript or C++.

With current layout you have a condition, then you have clearly two branches formatted quite symmetrically which makes sense to me.

In general now that we released the initial version we're going to need very good reasons to change any part of formatting. Those should better be technical and non-subjective.

mrkkrp:
@cblp Please make sure to first open an issue for changes like this so we could discuss before you spend time implementing something.

cblp:
The issue is #299.

cblp:
> I don't think we need to go out of our way...

What is your way?

cblp:
> I have never seen this formatting in Haskell code

This style is so common, that an extension was created for it, and is has been made default in Haskell2010.

cblp:
The technical reason is decreasing syntactical noise.

mrkkrp:
> The issue is #299.

That issue resulted in bumping indentation of `then` and `else`. I don't think we agreed on going with the version B there and there has been no new comments for a while.

> What is your way?

"Go out of your way" is an [idiom](https://idioms.thefreedictionary.com/go+out+of+way).

> The technical reason is decreasing syntactical noise.

Syntactical noise is hard to define precisely, so it's hardly a technical reason. I perceive your layout as more noisy, and this is just my opinion. Given two conflicting opinions on a subjective issue such as this I see no reason to change the status quo.

I do not want to discourage you from contributing, but if we start accepting changes that tweak existing formatting just because of taste preferences, the style will never get fixed as there are so many opinions as there are people. It's not a very productive use of our time, too.

cblp:
> just because of taste preferences

No, I really hope this can help people.

When I look to

```hs
if...
  then
    a
  else
    b
```

I see 5 syntax nodes: a fork, two branch selectors, two branch bodies.

And in

```hs
if...
  a
else
  b
```

I see only 3 syntax nodes: a fork and two branch bodies.

Well, it comes I'm the only one.

mrkkrp:
I see 4 syntax nodes: a fork, one branch selector, and two branch nodes.

cblp:
Well, it seems to be purely subjective.

cblp:
The only metric applicable is the indentation depth but is it good or bad is unclear.

mrkkrp:
As discussed, it is not obvious that this change is in fact an improvement. The current Ormolu policy regarding formatting is that we're going to keep the style we have right now unless there is a good objective reason to change it, so that the style doesn't change a lot between releases for people who already started using the tool.

Issue #451, opened 2019-11-04
Idempotence issue with operators


judah:
The following code leads to an idempotence error:

```
a =
  b & c .~ d
    & e %~ f
            g
```

Result:
```
Formatting is not idempotent:
  before: "  & e\n      %~ f\n   "
  after:  "  & e\n    %~ f\n     "
```

Additionally, indenting the `%~` to a different line than the `&` seems unnecessarily aggressive.  Furthermore, it results in `g` having less indentation than `f` in the output, when before it was indented further than `f`:

```
a =
  b & c .~ d
    & e
      %~ f
        g
```

utdemir:
Thanks! I fixed the idempotency issue (see the PR).

As about the styling issues, Ormolu thinks `e %~ f g` is a multiline expression, and it formats it as one, putting the operator on a newline. Technically, we can have a special case when the LHS is single line and RHS is multi line; but it would cause us to have two different formatting styles and I am not sure if it is worth it.

You are right on `g` having less indentation than `f` is not ideal. We used to always indent the arguments more than the function itself, but later we changed that behavior; however I don't remember why. @mrkkrp Do you remember the reason of that change? To me it makes sense to use `sitcc` on the arguments.

neongreen:
The reason is that if you indent g further than f, you will have indentation depending on the length (of the operator), which Ormolu generally tries to avoid.

mrkkrp:
I'm inclined to close this one now because:

* The transformation is idempotent now.
* All other points are really artifacts of Oromlu's general formatting style which emerged mostly as a combination of heuristics and compromises. It is not clear that we should change those at this point.

PR #459, opened 2019-11-17
Use mode instead of mean when calculating operator fixities

utdemir:
Related: #451.

When deciding on the fixity of an operator, we assign a score
based on some heuristics. When an operator is used multiple times
in an operator chain, we used to calculate the mean of its scores
to decide on the final fixity. However, this was prone to outliers.

In an example like this:

```haskell
b & c .~ d
  & e
    %~ f
        g
```

The first use of `&` is surrounded by other values, so we assign it
a higher score (binds tighter). However the second occurence is at
the beginning of a line, and this has a lower score. In the end, when
we average those two values the result binds tighter than `%~`, which
is not what we want.

This PR replaces the `average` method with a `mode` operation instead,
so it is less prone to outliers.

This PR fixes this specific issue (#451), and does not cause any change
on other test cases, but it might cause some diffs in the wild.


neongreen:
Can you show how operators-out.hs would have looked before this PR?

utdemir:
@neongreen `operators-out.hs` would look the same before this PR, but formatting it again would cause an idempotence issue by trying to reformat it as:

```haskell
a =
  b & c .~ d
    & e
    %~ f
      g
```

mrkkrp:
Could you please rebase on `master`? Does this PR in fact fix #451 completely so that it can be closed when we merge this?

utdemir:
Thanks Mark, I rebased the PR. It fixes the idempotency issue on #451, but does not address the stylistic issues also mentioned there.

mrkkrp:
@utdemir Please include your descriptions in the commit message next time, I think they deserve to stay in commit history, not just on GitHub :-)

PR #461, opened 2019-11-18
-m 'inplace': recurses into directories

benjaminweb:


mrkkrp:
My opinion is that this should be done externally, e.g. by calling the formatter from `find`. This solution adds code (and recursive traversal is a subtle topic to get right as you start thinking about things like following symlinks) and adds dependencies which we try to keep minimal. I do not remember seeing this sort of functionality in other formatters.

Can you explain why do you want to add this feature?

mrkkrp:
Remember that Ormolu already supports formatting several files in one go, so here what I'd do:

```
$ ormolu -m inplace $(find my-dir -type -f -name "*.hs")
```

This works very fast in my experience.

benjaminweb:
> Can you explain why do you want to add this feature?

Seemless user experience of tooling.

It's the default behaviour of `cargo fmt` in rust.
`hlint` is doing it similarly, including its `--refactor` mode.

Correct, symlinks do face a problem that can lead to non-termination.

basile-henry:
> It's the default behaviour of `cargo fmt` in rust.

As far as I know it is not a feature of `rustfmt` that `cargo fmt` uses under the hood. So the parallel would be for a command like `cabal format` to use `ormolu` on all the project files.

benjaminweb:
How would that need to look like for cabal?

mrkkrp:
I imagine there could be `cabal format` command and Cabal probably already knows paths of all source files in a project, so it would just call a tool like Ormolu with the paths (it can also pass flags to enable extensions, etc.).

benjaminweb:
Where in the [cabal code base](https://github.com/haskell/cabal) would that implementation need to happen? Unfortunately, I can't find it at this moment.
Who of the cabal devs could hint us to the right place?

mrkkrp:
I'm not familiar with Cabal's source code, so I can't say.

benjaminweb:
Is https://github.com/haskell/cabal/blob/eb2f7642f0f5db8103942496676349deffdd45a4/cabal-install/main/Main.hs#L301 already what we want or is this a false-positive?

mrkkrp:
I didn't know there is a format command in Cabal. In fact when I do `cabal --help` I can't spot it. Can you?

benjaminweb:
Nope, which makes sense as it is `hiddenCmd` on line 301. But what's the context it ended up there? commit log is not revealing for me. To you?

basile-henry:
I think that's to format the `.cabal` file itself.

What I was talking about doesn't exist in cabal (yet). I think the way cargo does this formatting is via a plug-in system. I think `cargo fmt` isn't something that's implemented directly by cargo but instead via the rustfmt component that one can install via `rustup component add rustfmt`.
It would be really nice to have plug-ins in cabal as well.

mrkkrp:
So the conclusion here seems to be that this should be implemented elsewhere.

benjaminweb:
@phadej proposes in https://github.com/haskell/cabal/issues/5741#issuecomment-555536972:

> My opinion is that ormolu could depend on Cabal the library to parse the .cabal files, and then identify the sources (like sdist command does, all of that is few dozens lines of code).

He adds:

> Or to say it directly: we won't hard code any formatters, why ormolu and not say brittany?

benjaminweb:
It may be this part, @phadej is referring to:
https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/SrcDist.hs#L112-L148

mrkkrp:
I answered in the thread.

robx:
> > Can you explain why do you want to add this feature?
> 
> Seemless user experience of tooling.
> 
> It's the default behaviour of `cargo fmt` in rust.
> `hlint` is doing it similarly, including its `--refactor` mode.
> 
> Correct, symlinks do face a problem that can lead to non-termination.

Another point in support is Go's `gofmt`. I routinely run `gofmt -w .` to format all .go files in a project. It's very convenient.

PR #464, opened 2019-11-19
Always place do-blocks in hanging position

### LINKED ISSUE #417: Expression-style "do" leads to strange output
It looks as though Ormolu prefers the trailing-style `do` instead of the expression-style `do`. That is,
```
main = do
  putStrLn "Hi!"
  putStrLn "I am a program!"
```
instead of
```
main =
  do putStrLn "Hi!"
     putStrLn "I am a program!"
```

However, formatting the latter results in the following:
```
main =
  do
    putStrLn "Hi!"
    putStrLn "I am a program!"
```
which seems to necessitate a manual pass to move all `do` keywords up to the previous line before automatically formatting. Perhaps an exemption from "respect the user's newlines" is in order in this particular case?

mrkkrp:
Close #417.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@
   [440](https://github.com/tweag/ormolu/issues/440) and
   [448](https://github.com/tweag/ormolu/issues/448).
 
+* `do`-expressions written in non-hanging form are now converted to hanging
+  form. See [issue 417](https://github.com/tweag/ormolu/issues/417).
+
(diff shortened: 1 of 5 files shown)


neongreen:
Can you add a test case?

mrkkrp:
This is a work in progress.

mrkkrp:
I got it to work as I want, but there is a complication because of how comments interact with this feature. I'll return to this later.

mrkkrp:
Closing for now.

PR #471, opened 2019-11-29
Take non-commutative nature of language pragmas into account

### LINKED ISSUE #404: Sorting language pragmas should be more intelligent
This was mentioned in #216, but it is a separate issue.

>The problem is that a list of language extensions is not always a list of commutative elements, however it often is a list of commutative elements and users will likely expect the alphabetized behavior in the latter case.

> If one were to take care to note which language extensions were implied by which other ones, we can determine which elements in the list of language extensions could not be commuted "past" one another when alphabetizing. If this bookkeepping was performed, we could indempotently output a mostly alphabetized list of language extensions when one or more elements cannot commute "past" each other or a completely alphabetized list when all elements can be commuted, preserving the module's semantics in either case.

> While this approach would require much more work to implement and maintain, it is the most correct option for an opinionated code styler.

IOW, we should take into account that some pragmas that disable extensions enabled as side-effect of enabling other extensions should be put after those extensions because sometimes the order of extension pragmas matters.

mrkkrp:
Close #404.

The order in which language pragmas are put in the input sometimes matters.
This is because some language extensions can enable other extensions, yet
the extensions coming later in the list have the ability to change it. So
here we classify all extensions by assigning one of the four groups to them.
Then we only sort inside of the groups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@
 * Implemented correct handling of stack headers. [Issue
   393](https://github.com/tweag/ormolu/issues/393).
 
+* Sorting language pragmas cannot not change meaning of the input program
+  anymore. [Issue 404](https://github.com/tweag/ormolu/issues/404).
+
 ## Ormolu 0.0.1.0
 
 * Initial release.
--- a/src/Ormolu/Printer/Meat/Pragma.hs
+++ b/src/Ormolu/Printer/Meat/Pragma.hs
@@ -8,18 +8,46 @@ module Ormolu.Printer.Meat.Pragma
   )
 where
 
+import Data.Char (isUpper)
+import Data.Maybe (listToMaybe)
 import qualified Data.Set as S
 import qualified Data.Text as T
 import Ormolu.Parser.Pragma (Pragma (..))
 import Ormolu.Printer.Combinators
 
-data PragmaTy = Language | OptionsGHC | OptionsHaddock
+-- | Pragma classification.
+data PragmaTy
+  = Language LanguagePragmaClass
+  | OptionsGHC
+  | OptionsHaddock
+  deriving (Eq, Ord)
+
+-- | Language pragma classification.
+--
+-- The order in which language pragmas are put in the input sometimes
+-- matters. This is because some language extensions can enable other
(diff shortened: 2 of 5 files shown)


mrkkrp:
@neongreen Please see if this reflects your ideas of how this should be fixed.

CHANGELOG.md
```diff
@@ -35,8 +35,8 @@
 * Implemented correct handling of shebangs. [Issue
   377](https://github.com/tweag/ormolu/issues/377).
 
-* Implemented correct handling of stack headers. [Issue
```
neongreen: Shouldn't the previous entry also stay in the changelog?

src/Ormolu/Printer/Meat/Pragma.hs
```diff
@@ -28,10 +53,19 @@ p_pragma :: PragmaTy -> String -> R ()
...
   txt " #-}"
   newline
+
+-- | Classify a 'LanguagePragma'.
+classifyLanguagePragma :: String -> LanguagePragmaClass
+classifyLanguagePragma = \case
+  "ImplicitPrelude" -> ImplicitPrelude
+  "CUSKs" -> CUSKs
+  str
+    | "No" `isPrefixOf` str -> Disabling
```
neongreen: `NondecreasingIndentation` also starts with "No", you might have to break the string into case-separated chunks first.

src/Ormolu/Printer/Meat/Pragma.hs
```diff
@@ -8,18 +8,43 @@ module Ormolu.Printer.Meat.Pragma
...
+  | OptionsHaddock
+  deriving (Eq, Ord)
+
+-- | Language pragma classification.
+--
+-- The order in which language pragmas are put in the input sometimes
+-- matters. This is because some language extensions can enable other
+-- extensions, yet the extensions coming later in the list have the ability
+-- to change it. So here we classify all extensions by assigning one of the
+-- four groups to them. Then we only sort inside of the groups.
```
neongreen: It would be nice to specify that the Ord instance specifies the order of the pragmas.

It would also be nice to link to the issue tracker, because singling out CUSKs and ImplicitPrelude is definitely a head-scratcher, even despite the comment above.

PR #477, opened 2019-12-03
Implement correct moving of per-import comments

### LINKED ISSUE #469: formatting does not preserve comments inside explicit imports
formatting
```
import MyModule ({- instance MyClass -})
```
will produce
```
import MyModule ()

{- instance MyClass -}
```
which is not quite what I would expect but in multiple imports this is still clear that `MyClass` belongs to `MyModule` unfortunately the following is not so clear anymore

```
import           MyReallyLongModuleNameModuleB ({- instance AnotherClassB -})
import           MyReallyLongModuleNameA       ({- instance MyReallyLongClassNameA -})

import qualified MyModuleC                     as C
import qualified MyModuleD                     as D
```
will be formatted as
```
{- instance AnotherClassB -}
{- instance MyReallyLongClassNameA -}

import qualified MyModuleC as C
import qualified MyModuleD as D
import MyReallyLongModuleNameA ()
import MyReallyLongModuleNameModuleB ()
```

my assumption is that this is due to the length of class+module name

mrkkrp:
Close #469.

This is a work in progress.

TODO:

- [ ] Fix the last failure in the test suite (should also make the remaining packages to pass).
- [ ] Clean up and rewrite in a nice way.
- [ ] Add a changelog entry.


mrkkrp:
After two days of fighting this, the idea behind the change doesn't seem to work very well. We rely a lot on relative positions of comments and input syntactic elements. When we move imports around and try to move comments with them, all sorts of corner cases and weird situations arise. The only thing that would work is moving comments strictly on the same line as an import, for example:

```haskell
import Foo -- foo
-- import Baz
import Bar -- bar
```

Here the comments for `Foo` and `Bar` work fine, but `-- import Baz` is more tricky because it's now a "normal" comment and it is not clear where to put it. It could be as well continuation of `-- foo`. If we start to deal with continuations like this it's not clear when they should stop. Would one blank line be enough to delimit continuation of last import from "just a comment" which should not be affected by reordering.

I tried to keep a special map for comments but it doesn't work very well. Currently thinking if there is a way to handle all this nicely after all. Our comment placing code is tuned to place comments when nothing is moved, here I just often see things I do not understand.

neongreen:
> It could be as well continuation of -- foo

That would seem pretty weird. I know that currently Ormolu produces those kinds of mis-aligned comments, but I have never seen them elsewhere.

Perhaps we could assume that if a comment starts on the same column as an import, it belongs to that import.

mrkkrp:
> Perhaps we could assume that if a comment starts on the same column as an import, it belongs to that import.

We could assume that, but it may be not a super-reliable criterion. Just matching the column somehow looks like a weak heuristic. I might as well write my comments like this:

```haskell
import Foo -- this is foo,
  -- I like it import it everywhere I can
import Bar
```

This is especially true if the import itself is already pretty long. You probably won't be aligning all the comments to the right only to get some 20 or so usable columns before you run over 80-column limit.

neongreen:
It is not super reliable, but at least then I can write a multiline comment above an import and it will stick.

The requirement to keep my comments on the same line as the import would discourage me from writing multi-line comments.

Based on currently open issues, it seems like people will have to adjust to Ormolu anyway (don't use Haddock inappropriately or it will disappear, etc), so I don't mind it if Ormolu does not handle all possible cases. But between style A and style B, I would definitely choose B.

```haskell
import Data.Letter.Latin.A -- I like A
  -- and A likes me
```

```haskell
-- I like B
-- and B likes me
import Data.Letter.Latin.B
```

mrkkrp:
The preceding comment might as well be completely unrelated:

```haskell
import Foo
-- import Bar
import Baz
```

neongreen:
Also true, but I don't think it's a big deal Ormolu doesn't guess that `-- import Foo` is different from `-- I like Bar`.

It all depends on whether you agree that supporting extensive comments (i.e. more than one line) for imports/pragmas is a must-have.

mrkkrp:
Note to self: when a comment gets assigned to an import which after sorting becomes the last element of the import section it won't be considered as part of the import section on next parse:

```haskell
import Foo -- foo

-- here we go
import Bar
import Baz
```

Gets transformed into:

```haskell
import Bar
import Baz
import Foo -- foo

-- here we go
```

Where `-- here we go` is no longer considered part of the comments that belong to the import section.

The same does not hold for preceding comments because we never attach comments to the import that follows them.

The logic for assigning comments between imports is: assign everything to the current import until we move past next import. The logic for picking up trailing comments after the last import is: include everything before first blank line. Apparently the blank line here is the problem.

mrkkrp:
I've put so much effort into this one and nothing decent comes out of it. Closing.

PR #479, opened 2019-12-06
Preserve newlines in class instance declarations

### LINKED ISSUE #431: Instance declarations are double-spaced.
This: https://code.world/haskell#P1YYfE2tmTPYIMJBKh7rHlw
becomes this: https://code.world/haskell#P71SxA1-aRo3lISoy75cv1w

The result is many times less readable.  This isn't even a question of inconsistent spacing.  The original made a perfectly consistent choice.

srid:
Preserve user-added newlines with the following normalizations:

* No newlines below `where`
* Consecutive newlines are compressed into a single newline

Add a new test fixture, in addition to modifying the existing ones.

Fixes #431 

## Todo

- [x] Preserve newlines in instance declarations
- [ ] Preserve newlines in class declarations as well
- [ ] Fix: Do _not_ preserve newline _after_ the comment and above the declaration (possibly via using `groupDecls`)

## Open questions

- [ ] Preserve newlines _within_ a _group_ of declarations (`groupedDecls`)?

### DIFF
--- a/data/examples/declaration/instance/associated-data-out.hs
+++ b/data/examples/declaration/instance/associated-data-out.hs
@@ -11,9 +11,7 @@ instance Foo Double where
         Double
 
 instance Foo [a] where
-
   data Bar [a]
     = ListBar [Bar a]
-
   data Baz [a]
     = ListBaz
--- a/data/examples/declaration/instance/associated-data.hs
+++ b/data/examples/declaration/instance/associated-data.hs
@@ -13,6 +13,7 @@ instance Foo Double where
 
 instance Foo [a]
   where
+
     data Bar [a] =
             ListBar [Bar a]
     data Baz [a] =
--- a/data/examples/declaration/instance/associated-types-out.hs
+++ b/data/examples/declaration/instance/associated-types-out.hs
@@ -3,7 +3,6 @@
 instance Foo Int where type Bar Int = Double
 
 instance Foo Double where
-
   type
     Bar
       Double =
--- a/data/examples/declaration/instance/associated-types.hs
+++ b/data/examples/declaration/instance/associated-types.hs
@@ -9,5 +9,7 @@ instance Foo Double
         Double
         =
           [Double]
+
+
     type instance Baz  Double
       = [Double]
--- a/data/examples/declaration/instance/instance-sigs-multiple-out.hs
+++ b/data/examples/declaration/instance/instance-sigs-multiple-out.hs
@@ -1,12 +1,10 @@
 {-# LANGUAGE InstanceSigs #-}
 
 instance Applicative [] where
-
   pure ::
     a ->
     [a]
   pure a = [a]
-
   (<*>) ::
     [a] -> [a] -> [a]
   (<*>) _ _ = []
--- a/data/examples/declaration/instance/multi-parameter-out.hs
+++ b/data/examples/declaration/instance/multi-parameter-out.hs
@@ -1,7 +1,5 @@
 instance MonadReader a ((->) a) where ask = id
 
 instance MonadState s (State s) where
-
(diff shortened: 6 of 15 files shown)


srid:
I tried to refactor `p_hsDecls` into doing what `p_hsDeclsPreserveNl` currently does (so as to obviate the need for a new function), but the grouping behaviour kept coming in the way, so I left it as is.

@mrkkrp CI failed due to authorization issues.

mrkkrp:
I added you as a collaborator, so your commits should trigger CI now. I also restarted the build.

* Should the same principle apply to class declarations themselves?
* Can you add a changelog entry for this change?

srid:
Alright, so I extended the change to cover class decls as well, but there is one more issue left to fix. 

For the following code,

```haskell
class MyClass a where
  foo :: a -> Int
  -- Some comment
  bar :: a -> Int
```

The PR will insert a newline above the `bar` function, only because of that non-Haddock comment. If the comment is removed, the expected behaviour (no newline inserted) is seen. 

Somehow non-Haddock comments are not playing well with the `groupDecls` group of functions, I think. Will need further investigation.

mrkkrp:
@srid I think if definitions have Haddocks associated with them, there should be a newline because otherwise it's a wall of text which is not nice.

srid:
@mrkkrp Okay. Is there a reason why we wouldn't want to extend that behaviour to non-Haddock comments too? You can see two examples of those here (see the "Need to remove" PR comments): https://github.com/srid/ormolu/pull/1

srid:
@mrkkrp This seems really tricky to implement, so I could use some guidance in the code design.

My use of `locsWithBlanks` is problematic, because it doesn't take into account (non-Haddock) comments at all, which ormolu handles _separately_ from the parser AST. Specifically, a comment stream is separately defined and ormolu handles in the `located` function adapter.

I tried extending the `located` function to return a `Maybe RealSrcSpan` which would be the "last" comment processed by it. I was hoping to use it in `p_hsDecls` to better determine if a newline should be inserted or not. However that doesn't seem like it would work, as by the time `located` is evaluated, the parsed chunks would have been spit out. Unless the newline can be inserted retroactively? 

What's the best way to do this?

Maybe I'm not correctly approaching the general problem of spitting out newlinews in the special context of comment stream.

mrkkrp:
There are a few things to know about newlines and comments:

* Newlines and spaces are inserted in a special way. If you call `space` or `newline` once then you'll get one newline/space. If you repeat it you'll get two newlines (but still only one space in case of  `space`). It's not possible though to insert more newlines than two: further invocations have no effect. We use this to make sure normal comments are separated from haddocks and everything else by outputting more newlines (sometimes) than strictly necessary, if I remember correctly. I think we should preserve this behavior: if there is a comment between class methods or other definitions it should be surrounded by blank lines.

* Last inserted comment's span can be obtained via `getLastCommentSpan`, although I do not think you'll need it.

So that said, given that Haddocks are handled together with other declarations by `p_hsDecl` and other comments are automatically handled by the `located` combinators, I'd concentrate on getting one specific case to work: when there are methods/associated data types without a blank line between them. I'd try to get that placement to be preserved. Otherwise if there is a Haddock, there should be always a blank line before Haddock, and if there is a "normal" comment, there should be blank lines around it. If this proves to be problematic or overly complicated, maybe we could just drop the leading blank line in the class definitions to match the class instances style and call it a day.

mrkkrp:
What is the status of this? @srid

srid:
Artyom has expressed interest in resuming work on my branch. @neongreen - any updates?

neongreen:
I am working on it

neongreen:
Superseded by #495

Issue #484, opened 2019-12-26
--ghc-opt always erroneously reports unrecognized options


Avi-D-coder:
This is related to #437 & #438.

```haskell
$ echo 'idString = id @String' | ormolu -o TypeApplications
```
```
The following GHC options were not recognized:
  TypeApplications
```

Removing `src/Ormolu/Parser.hs:208` gets rid of the error message, but ignores `TypeApplications`. Output: `idString = id@String`
```haskell
    case NE.nonEmpty leftovers of
      Nothing -> return ()
      Just unrecognizedOpts ->
        throwIO (OrmoluUnrecognizedOpts (unLoc <$> unrecognizedOpts))
```

mrkkrp:
As you seem to have discovered by now, the problem is that GHC options for language extensions begin with the `-X` prefix, which is how you should provide them.

PR #485, opened 2019-12-26
Except ghc opts without -X prefix

Avi-D-coder:
closes: #484

### DIFF
--- a/src/Ormolu/Parser.hs
+++ b/src/Ormolu/Parser.hs
@@ -57,7 +57,9 @@ parseModule Config {..} path input' = liftIO $ do
         GHC.setGeneralFlag'
           GHC.Opt_Haddock
           (setDefaultExts baseDynFlags)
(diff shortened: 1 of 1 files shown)


mrkkrp:
I do not see why we would need to introduce this. We state clearly that we accept GHC options, which are not limited to the flags that control language extensions. In GHC, the language extension flags are prefixed by `-X` and that's just how you should pass them. Appending the `-X` prefix to any flag that does not have it already would just cause problems and prevent passing other types of GHC flags to the parser.

Avi-D-coder:
Fair enough, but a note should be added to help.

Out of curiosity what non extension GHC options can have a effect?

mrkkrp:
> but a note should be added to help

PRs adding the note are welcome.

> what non extension GHC options can have a effect

I don't remember any, but the important thing is that there may be some. Sorry, I'm on vacation and typing from phone, so won't try to find examples.

Issue #74, opened 2019-06-07
Respect more newlines?


ruhatch:
This is really a question about how `ormolu` wants to treat newlines in general. I like to be liberal with whitespace to increase readability. For example I:

- Add newlines between groups of imports that I want to remain separate
- Add multiple newlines between sections within a single file
- Add newlines between expressions in a `do` block when each is a little complex
- Add newlines between bindings in a `where` block

It seems that `ormolu` will strip most of these out right now and I'm wondering how this should be treated in the future. One of the most interesting features of `ormolu`, for me at least, is the ability for the programmer to influence the output by, for example, adding line-breaks in type signatures. Newlines seem like another place where we can transfer some control to the programmer.

mboes:
> Add newlines between groups of imports that I want to remain separate

We used to do this a lot on projects, but found it was counter productive. Managing the groups and moving them around meant bigger diffs and more cognitive overhead when adding imports (not to mention, harder for automatic tooling to add new ones at the right place) for little to no gain made this not worth it. So we have a monolithic set of imports that we sort alphabetically (like `haskell-mode` in Emacs does) and that's it.

> Add multiple newlines between sections within a single file

We normalize those, following long established precedent from Tibbe's style guide (also the basis of [Tweag's style guide](https://github.com/tweag/guides/blob/master/style/Haskell.md)). You can add comments in the middle if you really want more vertical space there.

> Add newlines between expressions in a do block when each is a little complex

Grouping statements sounds fair. Their order isn't arbitrary but dictated by the semantics you want.

> Add newlines between bindings in a where block

Also fair, but within a where block of bindings, they should either all have an intercalated newline or none of them.

ruhatch:
> We used to do this a lot on projects, but found it was counter productive.

Fair enough, and I suppose one of the `ormolu` principles is to normalise that kind of stuff

> Also fair, but within a where block of bindings, they should either all have an intercalated newline or none of them.

Yeah, that seems reasonable to me. Maybe this can be similar to type signature, where any blank line between two definitions causes this intercalation behaviour?

A couple of other potential whitespace spots:
- Between language pragma list and module name (I find it quite unreadable having no space between those)?
- Between non-haddock comment and next top-level definition (it seems reasonable to close the gap between a `-- |` comment and a top-level def, but a `--` or `{- -}` block might not be related to that next def)?

mboes:
> > within a where block of bindings, they should either all have an intercalated newline or none of them.
> 
[...]

Issue #463, opened 2019-11-19
Space inserted between INLINE pragma and declaration only if it has Haddock


michaelpj:
This gets left as it is:
```
{-# INLINE foo #-}
foo :: Int
foo = 1
```

But this gets reformatted:
```
{-# INLINE foo #-}
-- | This is foo
foo :: Int
foo = 1
```
to
```
{-# INLINE foo #-}

-- | This is foo
foo :: Int
foo = 1
```

It's also left alone if I change the Haddock comment into a non-Haddock comment.

I think the pragma belongs "with" the declaration. This seems to happen in the cases where there isn't a Haddock comment, and I don't think having one should change it.

mrkkrp:
This is the first time I see pragmas placed like this. To me this is very low priority, you probably should just put the pragmas in more conventional locations, that is, any location without putting Haddocks between the pragma and the declaration.

michaelpj:
I can of course put them somewhere else, but I don't think that it's an especially weird place to put them. They're comments that talk about the function, there's no obvious reason either they or the Haddocks should come first.

(And not to snark, but a small amount of searching turned up https://github.com/tweag/linear-types/blob/8c3f95fc9fe8f843b71f96b4c8d6e8a2bb5fecfa/Examples/src/Cursors/PureStorable.hs#L59 amongst others :) )

mrkkrp:
Yeah, all I say is that it's relatively low priority.

PR #495, opened 2020-01-10
Respect grouping in classes and instances

### LINKED ISSUE #431: Instance declarations are double-spaced.

neongreen:
Closes #431. Related to #74.

Default signatures are silly and I have no idea what's the right treatment for them re/ newlines, but whatever.

Declarations with Haddocks are always given newlines from both sides. I'm not sure it's the best thing, but it's really easy to get rid of in the future, so also whatever.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@
 * Fixed indentation in presence of type applications. [Issue
   493](https://github.com/tweag/ormolu/issues/493).
 
+* Class and instance declarations now do not have a blank line after
+  `where`. Grouping of methods from the original input is also preserved
+  with some normalizations. [Issue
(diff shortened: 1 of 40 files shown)


neongreen:
It always adds newlines above `--` comments, but I think it's fine.

neongreen:
https://github.com/tweag/ormolu/issues/463 seems to be another point in favor of "don't be smart about blank lines above Haddocks" 🤔

PR #527, opened 2020-03-26
Mention a second vim integration: neoformat

smelc:
ormolu has been integrated in [neoformat](https://github.com/sbdchd/neoformat) 3 months ago and it simply works: call `:Neoformat` to format the current buffer in place using `ormolu`.


mrkkrp:
Thanks!

Issue #530, opened 2020-03-28
Support -XStandaloneKindSignatures


utdemir:
GHC 8.10 [supports](https://downloads.haskell.org/ghc/8.10.1/docs/html/users_guide/glasgow_exts.html#extension-StandaloneKindSignatures) `-XStandaloneKindSignatures` extension:

```
type T :: (k -> Type) -> k -> Type
```

This is exposed as a new `KindSigD` constructor of `HsDecl`. It looks like it doesn't steal a keyword, so we should also be able to add it to default extensions.

Issue #526, opened 2020-03-26
Support the new qualified import syntax


mrkkrp:
As of GHC 8.10 it appears that a new form of import syntax is possible:

```haskell
import X qualified
```

Enabled by `ImportQualifiedPost`.

There are three options how to deal with it:

* Convert everything to the old style and delete the pragma. This follows the idea of making the output uniformly formatted.
* Convert everything to the new style. Not an option, because only the latest version of the compiler supports it and we'd need to add a pragma per file, which is probably not what is expected from a formatter.
* Format depending of whether the `ImportQualifiedPost` pragma is enabled. More work, but this should please everyone. So I'd say let's go with this one.

utdemir:
Just to make the options complete; I guess another option would be to leaving it to the user. Looks like ghc-lib-parser exposes the required information. And one disadvantage of your preferred approach is that adding that language pragma will amplify the diffs by changing every qualified import. But I do not see a better solution, so I think we should still go ahead with your suggestion.

I am planning to implement this over the weekend, if no one else starts before me. I'm planning to implement it with two PR's where the first one updates ghc-lib-parser and the Nix code for GHC 8.10 and the second one actually implementing this change.

mrkkrp:
> And one disadvantage of your preferred approach is that adding that language pragma will amplify the diffs by changing every qualified import.

That should not happen. If for a particular module the pragma is enabled, then all qualified imports in that module should be formatted in postfix fashion already. Neither "adding of language pragma" is necessary. All we have to do is to check if it's enabled and then format accordingly.

PR #531, opened 2020-03-28
Update ghc-lib-parser to 8.10.1

utdemir:
GHC 8.10.1 comes with some changes to the AST, which works great for Ormolu, but causes this diff to be a bit large:

* Trees That Grow extension points for new constructors are now statically proven to be uninhabited, via `noExtCon :: NoExtCon -> a`. Thanks to this change I got rid of many `notImplemented` calls.
* `LPat` constructor is now a lot more usable, so we don't need to use the `locatedPat` combinator and can remove some boilerplate code.

Also it comes with `ImportQualifiedPost` and `StandaloneKindSignatures` we should support. I did not implement them in the scope of this PR, since they have separate issues #526 and #530 respectively.

It causes one behaviour change, where the ordering of qualified and non-qualified imports of the same module is changed. This is due to our usage of `gcompare` resulting a different ordering because of the AST change caused by the `ImportQualifiedPost` extension. I think this is acceptable and we shouldn't try to keep backwards compatibility there.

Another behaviour change is that; previously HsExpr had a few extra constructors for arrows and patterns used in expression context. Those programs were syntactically incorrect, but refused on a later stage. But we nonetheless formatted those constructs so Ormolu didn't fail there while keeping the source code intact. However, now those constructors are removed, so Ormolu fails with a parse error in this case (same as GHC). I also removed some tests exhibiting this behaviour.


data/examples/declaration/value/function/arrow/proc-forms2.hs
```diff
@@ -4,16 +4,16 @@ bar0 f g h x =
   proc (y, z) ->
     (|   test (h f.(h g)  -<  (y x).y z)((h g)  .  h f-<y z  .  (y x))   |)
 
-bar1 f g x y = proc _ -> f -< x&&&g -< y
+bar1 f g x y = proc _ -> (f -< x)&&&(g -< y)
```
mrkkrp, line 7: What is the reason for adding parentheses?

src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -37,6 +37,9 @@ import Ormolu.Utils
 import RdrName (RdrName (..), rdrNameOcc)
 import SrcLoc (combineSrcSpans, isOneLineSpan)
 
+-- import Control.Exception (throw)
```
mrkkrp, line 40: Should these lines be removed?

src/GHC/DynFlags.hs
```diff
@@ -8,31 +10,36 @@ where
...
+    { sGhcNameVersion =
+        GhcNameVersion
+          { ghcNameVersion_programName = "ghc",
+            ghcNameVersion_projectVersion = cProjectVersion
+          },
+      sFileSettings = FileSettings {},
+      sTargetPlatform =
+        Platform
+          { platformWordSize = PW8,
+            platformMini = PlatformMini {platformMini_arch = ArchUnknown, platformMini_os = OSUnknown},
```
mrkkrp, line 28: Let's break this line so its length doesn't exceed 80 characters.

mrkkrp:
Can you also please squash everything together and add copy your explanations from the PR into the commit message?

data/examples/declaration/value/function/arrow/proc-forms2.hs
utdemir: Otherwise it is not a valid syntax. Previously it used to work because that error was thrown after the parser; now the parsing fails.

utdemir:
> Can you also please squash everything together and add copy your explanations from the PR into the commit message?

@mrkkrp Done.

PR #532, opened 2020-03-29
Place "qualified" after the module name when ImportQualifiedPost is enabled

### LINKED ISSUE #526: Support the new qualified import syntax

utdemir:
Closes #526.

This PR checks if `ImportQualifiedPost` language extension is enabled, and places the "qualified" keyword accordingly.

Since it causes a change on the AST, it also modifies Ormolu's AST diff check to ignore this change.

This PR is based on top of #531, so it should be rebased (only the relevant commits, including and after 8c0655e) after that is merged to master.

### DIFF
--- a/data/examples/import/qualified-post-out.hs
+++ b/data/examples/import/qualified-post-out.hs
@@ -1,0 +1,5 @@
+{-# LANGUAGE ImportQualifiedPost #-}
+
+import Data.Text qualified as T
+import Data.Text qualified (a, b, c)
+import Data.Text qualified hiding (a, b, c)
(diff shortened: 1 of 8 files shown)


mrkkrp:
@utdemir I made some edits before merging, see explanation in the commit message.

PR #535, opened 2020-04-08
Compile the executable with `-rtsopts`

ethercrow:


utdemir:
I'm just curious, is there a specific reason for this change or do you think it's just nice to give the users the ability?

ethercrow:
I was just using GHCRTS env variable to set options in a program I was developing, used ormolu in the same shell and noticed that it refuses to run at all in such conditions.

PR #540, opened 2020-04-10
Format records with a single data constructor more compactly

### LINKED ISSUE #425: Minimize vertical space for single-constructor records
The current style (A) for records is:

```haskell
data Foo
  = Foo
      { bar :: Bar,
        baz :: Baz
      }
```

I propose the following style (B) for single-constructor records:

```haskell
data Foo = Foo
  { bar :: Bar,
    baz :: Baz
  }
```

## Pros

* The most common style, according to Hackage grep.

* One-constructor records seem to stay one-constructor records so git diffs don't get worse. (Not sure about this one. Feedback welcome.)

* Slightly better use of horizontal space.

* Much better use of vertical space for `data`-heavy modules – important for people who split their screens horizontally instead of vertically.

## Cons

* Still, some diffs are inevitable. Especially annoying when the fields were heavily commented and you have to rewrap the comments.

## Unresolved questions

* Should (A) be reformatted into (B)? "Yes" seems like a good answer, but it would be slightly unfortunate if it was impossible to break a `data ReallyReallyReallyReallyLongName = ReallyReallyReallyReallyLongName ...` even when requested by the programmerr.

mrkkrp:
Close #425.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
   cause the data constructor be unconditionally rendered in multiline layout
   [Issue 427](https://github.com/tweag/ormolu/issues/427).
 
+* Records with a single data constructor are now formatted more compactly.
+  [Issue 425](https://github.com/tweag/ormolu/issues/425).
+
 * Implemented support for the new language extension `ImportQualifiedPost`.
 
 * Implemented support for the new language extension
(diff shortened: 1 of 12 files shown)


poscat0x04:
This is quite ugly IMO

PR #547, opened 2020-04-14
Compare failing logs with exact expected result

mrkkrp:
This will allow us to test on more code and catch changes in the logs as
they happen.


mrkkrp:
Will investigate it tomorrow. Running it on the packages from Hackage seems to take a while now too (about 6 minutes), which is a shame because it makes CI slower. Maybe there is a way to refactor it to invoke ormolu once per package...

PR #551, opened 2020-04-16
Preserve grouping of do statements

### LINKED ISSUE #74: Respect more newlines?

mrkkrp:
Close #74, close #536. This is an experiment to find an alternative to #536.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+* Grouping of statements in `do`-blocks is now respected. [Issue
+  74](https://github.com/tweag/ormolu/issues/74).
+
 ## Ormolu 0.0.4.0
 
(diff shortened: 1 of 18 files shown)


mrkkrp:
@utdemir I think this solution is slightly clearer.

PR #563, opened 2020-04-24
Update design notes for CPP

facundominguez:


README.md
```diff
@@ -116,8 +116,9 @@ independent top-level definitions.
 
 * CPP support is experimental. CPP is virtually impossible to handle
   correctly, so we process them as a sort of unchangeable snippets. This
-  works only in simple cases when CPP conditionals are self-contained. Use
-  Ormolu with CPP at your own risk.
+  works only in simple cases when CPP conditionals surround top-level
+  declarations. See the [CPP][design-cpp] section in the design notes for
+  a discussion of the dangers.
```
facundominguez, line 121: I made an attempt to be more precise here. Maybe ormolu could have some check of CPP directives really surrounding top-level declarations. I'm rather thinking aloud.

DESIGN.md
```diff
@@ -24,7 +24,7 @@ It also includes recommendations for future implementers.
 
 We set for the following goals (mostly taken from
 [brittany](https://github.com/lspitzner/brittany)):
-* Preserve the meaning of the formatted functions;
+* Preserve the meaning of the formatted functions when no CPP is used;
```
facundominguez, line 27: There is a discrepancy in the list that appear here and in the readme, maybe they should be unified somehow.

README.md
mrkkrp: Many times it also works in the situations when CPP is in depth of a definitions, for example her it worked correctly: https://github.com/mrkkrp/zip/blob/master/Codec/Archive/Zip/Internal.hs#L566-L596. Of course it's just luck.

PR #570, opened 2020-04-27
Replace “idempotency” with “idempotence”

mrkkrp:


mheinzel:
Would it maybe be possible to accept both versions for the next release and only then remove the old one?

mrkkrp:
We are iterating fast and Ormolu is still unstable. No stability guarantees are given so that we can keep things simple.

mrkkrp:
You can use `-c` (the short form) which works with both versions.

mheinzel:
Cool, no problem.

PR #574, opened 2020-04-30
Prevent incorrect detection of continuations of comment series

### LINKED ISSUE #449: Idempotence bug with comments in data declaration
Input:
```
data A =
     B  -- C

   -- D
   | E
```

Output:

```
Formatting is not idempotent:
  before: " B -- C\n  | -- D\n   "
  after:  " B -- C\n    -- D\n  |"
```

Investigating further, it formats it as:

```
data A
  = B -- C
  | -- D
    E
```

which then gets reformatted as:

```
data A
  = B -- C
    -- D
  | E
```

mrkkrp:
Close #449.


mrkkrp:
One possible solution was to try to eliminate the concept of "comment continuation", which I explored today. That did not work. A safer, albeit also less elegant, solution is to attach to every comment an indication whether or not it is preceded by some non-whitespace content and then use that to detect comment continuations better. I'll try that on Monday.

Issue #414, opened 2019-10-21
Remove duplication in imports


tseenshe:
It would be very useful if ormolu were to de-dupe import sections such as

```haskell
import Data.List (nub)
import Data.List (nubBy)
```

into

```haskell
import Data.List (nub, nubBy)
```

This would reduce the burden on tools that add imports such as [`haskell-tng-import-symbol-at-point`](https://gitlab.com/tseenshe/haskell-tng.el/blob/4e2cf006eb8c544ace8aafd2e8375dcb31dd8c8e/haskell-tng-util.el#L73-89)

mrkkrp:
I don’t have anything against it, but realistically it’s going to be a pretty low-priority target.

PR #597, opened 2020-05-25
Remove duplicate LIE's from import lists

jonascarpay:
Sanitizes import lists
```haskell
import Data.Text (a, a)
import Data.Text hiding (a, a)
```
to
```haskell
import Data.Text (a)
import Data.Text hiding (a)
```
This seems sensible to me, although `data/examples/import/misc-out.hs` suggests that the original behavior is as intended?

### DIFF
--- a/src/Ormolu/Imports.hs
+++ b/src/Ormolu/Imports.hs
@@ -11,7 +11,7 @@ where
 import Data.Bifunctor
 import Data.Function (on)
 import Data.Generics (gcompare)
-import Data.List (sortBy)
+import Data.List (nubBy, sortBy)
 import GHC hiding (GhcPs, IE)
 import GHC.Hs.Extension
 import GHC.Hs.ImpExp (IE (..))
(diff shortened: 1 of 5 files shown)


mrkkrp:
I'd say it is closely related to #414. We try not to transform AST much so far.

mrkkrp:
Oh sorry, I thought it is an issue, not a PR.

src/Ormolu/Imports.hs
```diff
@@ -44,9 +44,11 @@ compareIdecl (L _ m0) (L _ m1) =
     n1 = unLoc (ideclName m1)
     isPrelude = (== "Prelude") . moduleNameString
 
--- | Sort located import or export.
+-- | Sort located import or export, dropping duplicates.
 sortLies :: [LIE GhcPs] -> [LIE GhcPs]
-sortLies = sortBy (compareIE `on` unLoc) . fmap (fmap sortThings)
+sortLies = nubSortBy (compareIE `on` unLoc) . fmap (fmap sortThings)
+  where
+    nubSortBy f = (>>= take 1) . groupBy (\a b -> f a b == EQ) . sortBy f
```
mrkkrp, line 51: This doesn't read very nice to me. Why does `nubSortBy` take an argument when it's only used once? Furthermore, what's wrong with doing `sortBy` and then `nubBy` (available in `base`)?

jonascarpay: The argument is used twice, or do I misunderstand?

`nubBy` is O(n^2), whereas this should fuse away. I suppose it shouldn't matter for the sizes we're dealing with though, would you rather use `nubBy`?

mrkkrp: Performance hardly matters here at all, while readability does. The bottleneck of Ormolu is parsing, it is incomparably slower.

jonascarpay: Makes sense. How's this?

PR #611, opened 2020-06-12
Preserve blank lines in ‘let’ and ‘where’ bindings

### LINKED ISSUE #554: Respect newlines between 'where' bindings
Thanks for merging the PR to preserve the grouping within `do`-blocks! The main thing of #74 that I'm missing now is about newlines between `where` bindings.

When defining multiple helper functions in a `where` clause, Ormolu will turn them into a single block that for me is hard to parse into individual bindings. Others seem to have similar issues and in some cases even added [empty comments](https://github.com/wireapp/wire-server/blob/e8f7c1019f5bbb36034f2f44d3acc3a01c4ae61f/services/galley/src/Galley/API/Teams.hs#L478) to visually separate the bindings.

There where two approaches already described in #74, I'm fine with either one:

1. Newline between all bindings or none, depending on existing newlines (similar to type signatures)

   > within a where block of bindings, they should either all have an intercalated newline or none of them. ([source](https://github.com/tweag/ormolu/issues/74#issuecomment-499855052))

   > Maybe this can be similar to type signature, where any blank line between two definitions causes this intercalation behaviour? ([source](https://github.com/tweag/ormolu/issues/74#issuecomment-499936288))

   Note that this should always remove newlines between multiple clauses of the *same* definition, as it does for top-level definitions.

2. Individually allow single or no newline between bindings, depending on existing newlines (similar to `do`-blocks as)

   > I'd argue that one might group let bindings by semantics as well [...] and putting newlines everywhere or nowhere might be too extreme? ([source](https://github.com/tweag/ormolu/issues/74#issuecomment-509564884))

See above.

**Additional context**

An example for the two suggestions:

```haskell
x = y
  where
    y = triangle 10
    triangle n = triangle' n 0

    triangle' 0 acc = acc
    triangle' n acc = triangle' (n - 1) (acc + n)
```

currently gets turned into

```haskell
x = y
  where
    y = triangle 10
    triangle n = triangle' n 0
    triangle' 0 acc = acc
    triangle' n acc = triangle' (n - 1) (acc + n)
```

with proposal 1 will be turned into

```haskell
x = y
  where
    y = triangle 10

    triangle n = triangle' n 0

    triangle' 0 acc = acc
    triangle' n acc = triangle' (n - 1) (acc + n)
```

and with proposal 2 will stay as it was.

mrkkrp:
Close #554.


mheinzel:
Would it make sense to also have a test case where there are multiple consecutive blank lines to show that they get collapsed into a single one? At least that's how I think it would behave.

mrkkrp:
Yeah, although the way Ormolu printer works it is simply impossible to insert more than one consecutive blank line :)

PR #615, opened 2020-06-14
[wip] Sort constraints

414owen:
Want to get some initial feedback on this.

Sometimes, you just need a lot of constraints, so having them sorted improves the reader's constraint lookup time.

The constraints are sorted by their embedded identifiers, case insensitively, and excluding built-in syntax like `(,)`, `[,]`, `forall`.

There's currently two [tests](https://github.com/tweag/ormolu/blob/033bb16032f1a056652883a67a2666daf745c2b0/data/examples/declaration/class/super-classes-out.hs) [failing](https://github.com/tweag/ormolu/blob/033bb16032f1a056652883a67a2666daf745c2b0/data/examples/declaration/class/dependency-super-classes-out.hs), where the formatter isn't reordering comments placed after constraints (it kind of screws up the formatting):

```haskell
( -- Foo?
  Bar a, -- Bar?
  Baz a, -- Baz
  Foo a
) =>
```

### DIFF
--- a/data/examples/declaration/class/dependency-super-classes-out.hs
+++ b/data/examples/declaration/class/dependency-super-classes-out.hs
@@ -9,8 +9,8 @@ class (MonadReader r s, MonadWriter w m) => MonadState s m | m -> s where
 
 -- | 'MonadParsec'
 class
-  ( Stream s, -- Token streams
-    MonadPlus m -- Potential for failure
+  ( MonadPlus m, -- Potential for failure
+    Stream s -- Token streams
   ) =>
   MonadParsec e s m
     | m -> e s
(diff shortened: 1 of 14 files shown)


mrkkrp:
> Want to get some initial feedback on this.

This is an extremely bad idea. I wish people opened issues to discuss things like this before investing their time.

* Tests are failing and will be very hard to fix. I know, because I [tried](https://github.com/tweag/ormolu/pull/477) to do something similar for the special case with imports (which we do sort). Comments are hard. It is already hard to print them as they were without re-ordering.
* We should be careful with excessive normalization. I can understand removing or adding parentheses for consistency, but reordering constraints? There may be a reason for the order in which constraints are written.

All in all, what you are trying to do is very hard to get right. If you manage to get it right it'll be most certainly at the expense of code clarity/simplicity. Finally, the end goal itself is not without controversy (to me at least).

414owen:
Totally fair.
I still find this useful, so I'll use my fork, but I see why this isn't a one-size-fits-all.

PR #648, opened 2020-09-04
add workflow for release binaries

### LINKED ISSUE #543: Possible to have pre-built binaries
# The situation

Building `ormolu` from source takes a long time (primarily due to `ghc-lib-parser`). In an unconstrained environment, this is an annoyance. In a constrained environment, this becomes an issue.

One example is if `ormolu` is being built on CI. It's possible to timeout building just `ormolu` (depending on the CI service). If there were binaries available for download, that ought to make using (and adopting) `ormolu` considerably easier. That's not the only example, but it's one that I run into often.

Some inspiration can be found with `hlint`. There are binaries to [pull down `hlint` directly](https://github.com/ndmitchell/hlint/releases/tag/v2.2.11), or a [script to run `hlint` automatically](https://github.com/ndmitchell/hlint#running-with-continuous-integration) (if you trust scripts off the internet). Not asking for this level of commitment. Pointing out some prior art.

# The ask

Is there any possibility we could get pre-built binaries available for download?

I'm not asking for a specific location on the internet, automation with CI, or any of that right now. The only thing I'm really looking for is at least one pre-built binary available for download (could be manually added to a release for all it matters to me). It would be ideal to have binaries for Linux, macOS, and Windows, but even just having one of the three would be a huge improvement or having to build from source.

If there are pre-built binaries available somewhere, could the documentation be a bit more explicit on where to find them? I didn't see anything that immediately popped out in the readme.

amesgen:
Closes #543.

This PR will add binaries for Linux, macOS and Windows when a new release (corresponding to a tag) is created. See [here](https://github.com/amesgen/ormolu/releases/tag/test-release) for an example ([workflow run](https://github.com/amesgen/ormolu/actions/runs/239377467)).

On Linux, the excellent [ghc-musl](https://github.com/utdemir/ghc-musl) by @utdemir is used.

GHC 8.10.1 is used due to this bug on Windows on 8.10.2: https://gitlab.haskell.org/ghc/ghc/-/issues/18550

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,0 +1,39 @@
+name: Build binaries
+on:
+  release:
+    types: [published]
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-haskell@v1
+        if: runner.os != 'Linux'
+        with:
+          ghc-version: '8.10.1'
+          cabal-version: '3.2'
+      - name: Build Linux binary
+        if: runner.os == 'Linux'
+        run: |
+          docker run \
+            -v $(pwd):/mnt -w /mnt \
+            utdemir/ghc-musl:v14-ghc8101 \
+            bash .github/workflows/build.sh
+      - name: Build ${{ runner.os }} binary
+        if: runner.os != 'Linux'
+        shell: bash
+        run: |
+          cabal build exe:ormolu
+          cp $(find dist-newstyle \( -name ormolu -o -name ormolu.exe \) -type f) ormolu
+      - uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: ormolu
+          asset_name: ormolu-${{ runner.os }}
+          asset_content_type: application/octet-stream
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -1,0 +1,7 @@
+set -e
+
+cabal update
+cabal build exe:ormolu --enable-executable-static --ghc-options="-split-sections"
+
+cp $(find dist-newstyle -name ormolu -type f) .
+strip -s ormolu


.github/workflows/binaries.yml
```diff
@@ -0,0 +1,39 @@
+name: Build binaries
+on:
+  release:
+    types: [created]
```
mrkkrp, line 4: Should this be "published" instead?

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          ghc-version: '8.10.1'
+          cabal-version: '3.2'
+      - name: Build Linux binary
+        if: runner.os == 'Linux'
+        run: |
+          docker run \
+            -v $(pwd):/mnt -w /mnt \
+            utdemir/ghc-musl:v14-ghc8101 \
+            bash .github/workflows/build.sh
```
mrkkrp, line 25: This uses a different build method and in the end a different set of dependencies (Cabal-managed vs Nix-managed). Such executable is thus not tested by CI.

What I think we could do here is switching completely to GitHub actions + Cabal and pinning dependencies with Cabal freeze file. One thing that prevents this is that we now use Nix for testing extensively and a similar setup will be hard to achieve without Nix.

As an intermediate solution we could just merge this one for now.

.github/workflows/binaries.yml
amesgen: Yes, will change. I took it from [here](https://github.com/haskell/haskell-language-server/blob/e8bdd77921576f857368848480f131f38286c81c/.github/workflows/build.yml#L9), but e.g [this SO answer](https://stackoverflow.com/a/61066906) suggests to use `published`, but I can't find official documentation for this.

.github/workflows/binaries.yml
amesgen: I don't have any experience with nix, so I can't help with that. The Windows binaries can't use Nix of course.

mrkkrp (approved):
Thanks! Can you please squash commits and then we'll merge.

mrkkrp:
Should I create a release now to trigger the workflow?

amesgen:
Yeah, let's test it for e.g. the latest release!

amesgen:
Hmm, it doesn't seem to work for old tags.

mrkkrp:
I see, well there are two choices: create an ad-hoc tag/release just to test this or wait till a new version is published. I guess you'd prefer the first option?

amesgen:
Yes, and you can even delete the tag/release after this test.

PR #649, opened 2020-09-06
 Avoid modifying correctly-formatted files.

SamirTalwar:
Currently, when running with `--mode=inplace`, Ormolu always overwrites the files, even when nothing has changed, updating the modified timestamp of the file. This encourages Stack to rebuild everything when only one file has changed.

A quick check to see if the file needs rewriting solves this problem.

I would prefer the API allowed me to only read the file once, but as the `--mode=check` logic also reads the file twice, I decided this wasn't the biggest sin for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+* Ormolu no longer overwrites already formatted files. [PR
+  649](https://github.com/tweag/ormolu/pull/649).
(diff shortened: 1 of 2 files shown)


mrkkrp:
Thanks!

SamirTalwar:
No problem. Looking forward to this hitting nixpkgs-unstable. 🙂

PR #653, opened 2020-09-17
Clarify enabling/disabling

tomjaguarpaw:
I found the original paragraph a bit hard to understand, partly because it used the word "should" which is ambiguous in English. It can be interpreted as "really ought to but sometimes doesn't" or "is morally compelled to". "Must" is less ambiguous.


mrkkrp (approved):

mrkkrp:
Thanks!

PR #661, opened 2020-10-08
Allow path 0.8

felixonmars:
Builds fine and all tests pass here.


mrkkrp:
Thanks! I also released a corresponding revision on Hackage.

PR #666, opened 2020-11-10
Support MonadComprehensions.

### LINKED ISSUE #665: MonadComprehensions not supported
**Describe the bug**
If `MonadComprensions` is used, modules that use list comprehensions now error, where Ormolu previously formatted them successfully.

**To Reproduce**
Add `{-# language MonadComprehensions #-}` to a module containing list comprehensions that Ormolu normally passes.

**Expected behavior**
I _think_ that `MonadComprehensions` doesn't change the syntax provided by list comprehensions at all, it only weakens the type constraint internally, so code should format identically regardless of that extension. So I think the fix is to replace https://github.com/tweag/ormolu/blob/64735598c78ce9e3ab45c100ee03f4a81599da1c/src/Ormolu/Printer/Meat/Declaration/Value.hs#L705 with
```haskell
      MonadComp -> compBody
```

**Environment**
 - OS name + version: NixOS 20.03
 - Version of the code: 0.1.3.1

sellout:
Fixes #665.

I tested this with just the new example files, and it failed. Adding the change to Value.hs made it succeed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+* Added support for monad comprehensions. [Issue
+  665](https://github.com/tweag/ormolu/issues/658).
+
 ## Ormolu 0.1.3.1
 
 * Fixed a problem with multiline record updates using the record dot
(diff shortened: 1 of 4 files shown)


mrkkrp:
Can you please update the changelog accordingly?

PR #672, opened 2020-12-13
update CI for release binaries

amesgen:
[`actions/setup-haskell`](https://github.com/actions/setup-haskell) is deprecated, but [`haskell/actions/setup`](https://github.com/haskell/actions/tree/main/setup) is a drop-in replacement.

Also, updating to 8.10.2.

See [here](https://github.com/amesgen/ormolu/actions/runs/418860529) for a run in my fork.


mrkkrp:
Thanks!

Issue #691, opened 2021-03-13
Recursively format directory


brandon-leapyear:
**Is your feature request related to a problem? Please describe.**
I would like to be able to format all Haskell files in a given directory, e.g. run on all files in my project like `ormolu .`

**Describe the solution you'd like**
Ormolu should accept both files and directories. If a directory, recursively search for all `.hs` files, automatically excluding certain directories like `.stack-work` (which contains autogenerated haskell files), and (optionally) excluding additional files/directories passed in the command line.

**Describe alternatives you've considered**
Currently, this can be done with
```bash
# all files in project
ormolu $(git ls-files '*.hs')

# all files in directory
ormolu $(git ls-files 'foo/**/*.hs')
```
but it would be nice to have this built-in, like `hlint`

**Additional context**

mrkkrp:
Closing, see #697.

PR #697, opened 2021-04-11
Recursively format directory

goolord:
related to https://github.com/tweag/ormolu/issues/691

if the `FilePath` passed to `formatOne` is a file, format it like normal. 
if the `FilePath` passed to `formatOne` is a directory, walk that directory for Haskell source files, calling `formatOne` on each file.

not yet implemented: excluding directories. 
the first argument of `walkDir` allows you to exclude directories from being walked, but I'm not sure what the best way to do that is.
for instance, a user might want to exclude files/dirs in their gitignore or exclude directories via a glob

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -8,6 +8,7 @@ module Main (main) where
 
 import Control.Monad
 import Data.Bool (bool)
+import Data.Foldable (Foldable (foldl'))
 import Data.List (intercalate, sort)
 import Data.Maybe (mapMaybe)
 import qualified Data.Text.IO as TIO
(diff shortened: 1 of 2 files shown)


mrkkrp:
To be honest, I’m slightly opposed to this change because there are too many subtle details about walking directories (should we follow symlinks for example?) that I believe do not belong in Ormolu.

goolord:
yeah, this was pretty quick to sketch out but I quickly realized any more featureful handling of directories than this would quickly become very hairy. the feature doesn't really seem necessary in the presence of the bash trick mentioned in the issue, anyways

```bash
# all files in project
ormolu $(git ls-files '*.hs')

# all files in directory
ormolu $(git ls-files 'foo/**/*.hs')
```

mrkkrp:
I mentioned the `git ls-files` trick in the readme! I think it is even nicer than using `find`.

Issue #705, opened 2021-05-10
RDP record updates are converted to regular record updates


davidmrt98:
### Bug description
When the RecordDotPreprecessor plugin is enabled, record updates which conform to the RDP syntax (no space **before** the opening brace), but have one or many spaces **after** the opening brace, are converted to regular record updates when invoking Ormolu.

I consider this bug the "first half" of the idempotency bug presented in #701, since it can be observed in the first described invocation of Ormolu, whereas the second invocation uncovers another (somewhat unrelated) problem, which can be dealt with separately.

### Steps To Reproduce
- create a file with the following RDP style record update:
```hs
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

def{ g = 1}
```
- invoke Ormolu on this file
- the result looks like this (note the inserted space before the opening brace):
```hs
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

def {g = 1}
```
According to the last test case in `data/examples/declaration/value/function/record/dot-singleline.hs` and its `-out.hs` counterpart, this is intended behaviour, but I guess that is just an oversight.
### Expected behavior
As RDP record updates are semantically different from regular ones, Ormolu should always preserve the absence of spaces before the opening brace (when RDP is enabled). Thus, the formatting result should look like this:
```hs
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

def{g = 1}
```
### Environment
 - Ubuntu 18.0.4
 - ormolu-0.1.4.1

### The cause of the bug
Ormolu determines if a record update conforms to RDP syntax by comparing the source span of the expression (`def`) and the source span of the first updated field (`g`). Unfortunately, the position of the opening brace is independent of these source spans, so the current method for differentiating between regular and RDP record updates does not work correctly in the general case.

One possible way of fixing this is storing the location information of all the `AnnKeywordId`s in the Ormolu-specific `Anns` type so that the the exact location of the opening brace can be looked up, when needed. But considering that the actual RecordDotSyntax extension is around the corner, the added complexity of such a big change might not be worth it.

PR #706, opened 2021-05-10
Fix rendering of regular record updates with RDP enabled

### LINKED ISSUE #701: Record updates with RDP are not idempotent
```haskell
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

a =
  def{ N.k = g
     }
```

Invoking ormolu-0.1.4.1 results in:

```haskell
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

a =
  def
    { N.k = g
    }
```

Invoking ormolu-0.1.4.1 again results in:

```haskell
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

a =
  def{N.k = g
     }
```

davidmrt98:
Closes #701.

This pull request solves the "second half" of the idempotency bug presented in #701, but there is a another bug (the "first half") for which I do not see an obvious solution. Since the other bug is not directly related to this, I've opened a separate Issue where I explain the problem in detail, see #705.

When Ormolu checks if a record update conforms to the RDP syntax (no space **before** the opening brace), it compares the column numbers of the relevant source spans but not the line numbers. This is why a regular record update of the form
```hs
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

def 
  { g = 1 
  }
``` 
where the start / end columns of the expression `def`  and the field `g` are aligned in a specific way, "tricks" Ormolu into thinking that this is an RDP record update, even when it is clearly not. The formatting result looks like this:
```hs
{-# OPTIONS_GHC -fplugin=RecordDotPreprocessor #-}

def{g = 1
   }
```
For the same reason the second code snippet from #701 gets formatted the way it is. Fortunately, the problem is simply solved by checking if the relevant source spans are on the same line.

### DIFF
--- a/data/examples/declaration/value/function/record/dot-multiline-out.hs
+++ b/data/examples/declaration/value/function/record/dot-multiline-out.hs
@@ -8,3 +8,8 @@ fooplus'''' f n =
   f{foo = n,
     bar = n
    }
+
+fooplus''''' f n =
+  f
+    { foo = n
+    }
--- a/data/examples/declaration/value/function/record/dot-multiline.hs
+++ b/data/examples/declaration/value/function/record/dot-multiline.hs
@@ -5,3 +5,6 @@ bar' = (Foo 1){bar = 2
 fooplus'''' f n = f{foo = n,
                     bar = n
                    }
+
+fooplus''''' f n = f
+                   { foo = n }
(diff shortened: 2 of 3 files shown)


mrkkrp:
Thanks!

PR #711, opened 2021-06-13
Simplify p_conDecl GADT rendering by using p_hsType

expipiplus1:
Before this there was some logic replicated in p_hsType, what this
change does is instead of printing directly it constructs the
appropriate type and spans to pass to p_hsType


mrkkrp:
Excellent, thanks!

expipiplus1:
Thanks!

Issue #601, opened 2020-06-01
Improve disabling of Ormolu via magic comments


mrkkrp:
It should be possible to use the magic comments for enabling/disabling Ormolu in indented positions:

```haskell
foo =
    {- ORMOLU_DISABLE -}
    testCase "Foo" testFoo :
    testCase "Bar" testBar :
    testCase "Baz" testBaz :
    {- ORMOLU_ENABLE -}
    []
```

PR #714, opened 2021-06-17
Preserve formatting of disabled regions

### LINKED ISSUE #673: DISABLED ormolu remove empty lines
**Describe the bug**
I expect that ORMOLU_DISABLE can fully disable ormolu, but it continues removing empty lines.

**To Reproduce**
Input file:
```haskell
{- ORMOLU_DISABLE -}

module Foo (foo,bar) where



foo :: Int -> Int
foo = (+5)



bar :: Bool -> Bool
bar True  = True
bar False = True
```

Output file:
```haskell
{- ORMOLU_DISABLE -}

module Foo (foo,bar) where

foo :: Int -> Int
foo = (+5)

bar :: Bool -> Bool
bar True  = True
bar False = True
```

**Expected behavior**
Output source code should be identical to the input if I put `{- ORMOLU_DISABLE -}` at a file start.

```haskell
{- ORMOLU_DISABLE -}

module Foo (foo,bar) where



foo :: Int -> Int
foo = (+5)



bar :: Bool -> Bool
bar True  = True
bar False = True
```

**Environment**
 - OS name + version: MacOS X 11.0.1
 - Version of the code: ormolu 0.1.4.1

### LINKED ISSUE #708: Error on "-}" in regions with disabled formatting
**Describe the bug**
Parsing fails if the sequence of characters `-}` appears (in context other than a multiline comment) in a region where Ormolu is disabled using the magic comment.

**To Reproduce**
Run `ormolu` on the following input:

```
{- ORMOLU_DISABLE -}
a::String
a="-}"
```

This fails with:

```
<stdin>:3:7
  lexical error in string/character literal at character '\n'
```

**Expected behavior**
Unchanged data should be returned.

**Environment**
 - OS name + version: Linux 4.15.0-143-generic x86-64
 - Version of the code: 3adef733a99018c5e4437faaa0cc53b503c5d576 (ormolu 0.1.4.1 UNKNOWN UNKNOWN using ghc-lib-parser 8.10.2.20200916)

**Additional context**
My guess is that this follows from the way "disabled" regions are masked using a multi-line comment `{- ORMOLU_DISABLE_START ... ORMOLU_DISABLE_END -}` (see `Ormolu.Processing.Preprocess`).

If I am right, then it can perhaps be fixed by commenting out the "disabled" region by adding a prefix `-- ORMOLU_DISABLED` to each line in the region, instead of using the multiline comment. I think that with this approach #673 can also be easily fixed by modifying the following line to avoid removal of trailing spaces in "disabled" regions (and we'd avoid the removal of blank lines for free): https://github.com/tweag/ormolu/blob/3adef733a99018c5e4437faaa0cc53b503c5d576/src/Ormolu/Parser/CommentStream.hs#L107

kukimik:
This is meant to close #673 and close #708.

The idea is to cut out the disabled regions in the preprocessing, and paste them back in the postprocessing, instead of commenting them out.

When cutting out the disabled regions we reduce the indentation in each of them. The disabling and enabling markers bounding the region are not cut out (they are not treated as a part of the disabled region), but only normalized. Ormolu then finds the right indentation level for the disabling marker, treating it like a normal comment. In the postprocessing we paste the disabled regions back between the markers. The indentation level of a region is adjusted basing on the indentation level of the disabling marker.

I admit that the whole solution is somewhat hacky (but so is the current one). The pasting will go wrong if a new `{- ORMOLU_DISABLE -}` line somehow appears in the code as a result of formatting, or if an existing one disappears. I have not found a case when this may happen, but perhaps there should be an explicit `error` in postprocessing when the number of cut-out disabled regions does not match the number of `{- ORMOLU_DISABLE -}` markers.

The readability of the cutting/pasting code can probably be improved, and there are possibly some other details to be tweaked, but I'd love to get some feedback first.


src/Ormolu/Processing/Preprocess.hs
```diff
@@ -29,79 +31,89 @@ preprocess ::
...
+  -- | Regions where formatting is disabled, all remaining lines
+  (DisabledRegions, [String])
+cutDisabledRegions = go (id, id)
+  where
+    cut xs = (prefix, disabledRegionLines, suffix)
+      where
+        (prefix, rest) = break isOrmoluDisable xs
+        (disabledRegionLines, suffix) = break isOrmoluEnable rest
+    removeIndentation xs = drop (getIndentationLevel 0 xs) <$> xs
+    go (regionsSoFar, linesSoFar) (cut -> (pre, [], _)) =
```
mrkkrp, line 77: The problem with this is that GHC will see an input with a different number of lines in total and therefore its error messages (when the parser fails) will be shifted and inaccurate. Consider this:

```haskell
module Main (main) where

{- ORMOLU_DISABLE -}
foo Int
foo Int
foo Int
foo Int
foo Int
foo Int
foo Int
foo 5 EU>>>>

{- ORMOLU_ENABLE -}

main :: IO ()
main = return <<<
```

The last line fails to parse. On `master` it report the problematic location correctly:

```
disabled-test.hs:17:1
```

Note line 17.

With your change the region in the middle is cut, so the error messages becomes misleading:

```
disabled-test.hs:8:1
```

Note line 8.

This is, in fact, was the main motivation for doing things the way it is done right now. Fortunately, I think we can work this around if we insert the right number of blank lines between `disableMarker` and `enableMarker`.

kukimik: Thanks! I've tried the idea with blank lines and it seems to work. I'm going for holidays for now, will get back to this mid-July.

kukimik: Looks like the blank lines work. Significant changes in the last commit:

* Replacing disabled regions with blank lines.
* Normalize all `{- ORMOLU_ENABLE -}` magic comments (in the previous version only the comments at the beginning and the end of a disabled region were normalized, which differs from `master`). I decided not to normalize the `{- ORMOLU_DISABLE -}` comments inside disabled regions. See `data/examples/other/disabling/normalize-magic-comments.hs`.
* `Ormolu.Utils.getIndentationLevel` was made total (with the provided default value used for empty lists, which is not very clean).
* Rewording of `cutDisabledRegions` and `pasteDisabledRegions`. I find the new code somewhat more readable, thought I'm not really satisfied with it.

However, this is still work in progress. I'm now afraid that moving the processing of disabled regions out of `processLine` was a wrong decision, as it may cause problems in interplay between disabled regions and CPP. I'll have to check this, but it's late today and, once again, I won't have time for it for a few days.

Perhaps I'll also try to include your `disabled-test.hs` as a test case.

kukimik: It turns out that mixing magic comments with CPP or LINE pragmas results in a mess, especially when one disables formatting in a fragment of a definition (which is not much of a surprise). In particular running the `master` version of  `ormolu` on any of the following files result in the "AST of input and AST of formatted code differ" error, and the `--unsafe` output doesn't look very good.

```haskell
f =
  g
    {- ORMOLU_DISABLE -}
#ifdef ABC
    h
#else
    i
#endif
    j
    {- ORMOLU_ENABLE -}
    k
```

```haskell
f =
  g
    {- ORMOLU_DISABLE -}
    h
    i
{-# LINE 42 "Foo.vhs" #-}
    j
    {- ORMOLU_ENABLE -}
    k
```

Unfortunately keeping backward compatibility with the present behaviour and improving it are both somewhat tricky. Perhaps the latter may  be easier.

kukimik:
This took me longer than expected.

I've fixed another bug along the way: LINE pragma in a disabled region no longer leads to an AST change error.

Some thoughts and remarks about this MR:
* `Ormolu.Processing.Common` was moved to `Ormolu.Processing.Disabling`, as it only contains definitions related to disabling. This is analogous to `Ormolu.Processing.Cpp`.
* The disabled regions could perhaps be pretty-printed in `Ormolu.Parser.Result.prettyPrintParseResult` (currently I just `show` them). I didn't find this necessary.
* The preprocessing part is ugly. I've found it difficult to make it look better using the current structure of the code. I thought about restructuring it as a composition, more or less like: `preprocess = removeLinePragmas . removeDisabledRegions . maskCPP . removeShebang`, or using state to make it more readable. I decided that it's a big change, and I'd rather hear your opinion first.
* The test cases in `tests/Ormolu/Parser/ParseFailureSpec.hs` could perhaps be specified using srcSpans instead of strings. The current implementation relies on the text representation of the srcSpan, which may change (but I believe it is unlikely to change), It is, at least in part, a result of me being lazy, but it is also short and readable. If you think this should be rewritten, let me know.

amesgen:
Thanks for your effort on this PR! We chose to go with a slightly different approach (formatting the non-disabled snippets individually, see #773). Your work here is very appreciated, and your new tests are actually reused (of course, you are attributed as a coauthor).

kukimik:
@amesgen Ok, thanks! I haven't read through #773 yet, but the solution here wasn't very clean, so I'm pretty sure you've done better. I strongly agree with the decision to remove support for #601 . It was a major pain point while I was working on this PR. I've found this behaviour inconsistent and considered it a misfeature. Perhaps I should have complained to Mark about it instead of making workarounds.

PR #719, opened 2021-07-03
Implement HsCmdApp (fixes #716)

### LINKED ISSUE #716: Certain arrow expressions cause "not implemented yet: HsCmdApp"
**Describe the bug**
Running ormolu on valid Haskell source files with Arrows extension cause
ormolu to exit with "ormolu: not implemented yet: HsCmdApp". It seems
the core fragment is `(| ... |) thing`, i.e. the extra word after the brackets.

**To Reproduce**

```
$ cat Arrows.hs
{-# LANGUAGE Arrows #-}

f = proc x -> do
        y <- f -< x+1
        (| runReader (returnA -< ()) |) y
$ ormolu Arrows.hs
ormolu: not implemented yet: HsCmdApp
CallStack (from HasCallStack):
  error, called at src/Ormolu/Utils.hs:59:22 in rml-0.1.4.1-f355d56b:Ormolu.Utils
```

**Expected behavior**
I expect ormolu to run successfully.

**Environment**
 - macos 10.14.6
 - Version of the code HEAD as of today

**Additional context**

GHC documentation; this example was pieced together from that: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/arrows.html#primitive-constructs
Source code comment says to ping @yumiova.

robx:
I took a stab at implementing a fix to #716.

- this seems to work, but comes mostly from mental pattern matching, not understanding
- the examples are based on GHC test suite test cases; should they be simplified and/or should the origin be documented?
- not sure about naming of the example file (or whether the example should go in one of the existing files)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * Made record dot pre-processor rendering idempotent in certain specific
   cases. [Issue 701](https://github.com/tweag/ormolu/issues/701).
 
+* Added support for arrow command application. [Issue
+  716](https://github.com/tweag/ormolu/issues/716).
+
 ## Ormolu 0.1.4.1
 
 * Added command line option `--color` to control how diffs are printed.
--- a/data/examples/declaration/value/function/arrow/proc-applications2-out.hs
+++ b/data/examples/declaration/value/function/arrow/proc-applications2-out.hs
@@ -1,0 +1,11 @@
+{-# LANGUAGE Arrows #-}
+
+t = proc ys ->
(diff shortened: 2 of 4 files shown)


mrkkrp:
It all looks good to me as it is. The only thing I have doubts about whether the separator should be `space` or `breakpoint`. I'd say we usually use breakpoints in that case, but it looks too spacious (which is still fine). Ideally, I guess, the same treatment that we use for function application should be applied because supposedly these arrow applications can be as complex, but that's quite a bit of work. The use case seems rather rare, so let's go with the space for now.

PR #720, opened 2021-07-14
WIP: Migrate to ghc-lib-parser-9.0.1

### LINKED ISSUE #688: Switch to ghc-lib-parser-9.0.1.xxx
This should resolve some long-standing issues such as #474 and #480. Not sure how much work this is.

berberman:
Close #688.

This PR is very unpolished, though currently it builds, and all tests can pass with `cfgUnsafe = True`. Since I'm not very familiar with the changes in ASTs since GHC 9.0, basically I write these code following the hints from the powerful type system and intuition. It would be great if someone could help :)

### TODO
- [ ] Figure out why we need `ImpredicativeTypes` to typecheck `Ormolu.Diff.ParseResult` in GHC 9 (perhaps something related to simplified subsumption?)
- [ ] Make tests pass with `cfgUnsafe = False`
- [ ] Add tests for new syntax in GHC 9
- [ ] Cleanup code and squash commits

I also have a question that is there any use case of [`BufSpan`](https://hackage.haskell.org/package/ghc-lib-parser-9.0.1.20210324/docs/GHC-Types-SrcLoc.html#t:BufSpan)s in ormolu? They are now ignored.

### DIFF
--- a/data/examples/declaration/value/function/pattern/as-pattern.hs
+++ b/data/examples/declaration/value/function/pattern/as-pattern.hs
@@ -1,3 +1,3 @@
 main = case [1] of
-  xs @ (x:_) -> print (x, xs)
+  xs@(x:_) -> print (x, xs)
   xs@[] -> print xs
--- a/data/examples/declaration/value/function/pragmas-out.hs
+++ b/data/examples/declaration/value/function/pragmas-out.hs
@@ -3,9 +3,3 @@ sccfoo = {-# SCC foo #-} 1
 sccbar =
   {-# SCC "barbaz" #-}
   "hello"
-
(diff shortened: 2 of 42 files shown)


mrkkrp:
Thanks for starting this!

> I also have a question that is there any use case of `BufSpans` in ormolu? They are now ignored.

I don't see any mention of `BufSpan` in the source

amesgen:
Thank you for this! I started to work on this independently (probably should've created a draft PR earlier), and I think #722 mostly finished by now. Your remark that `ImpredicativeTypes` makes `Ormolu.Diff.ParseResult` compile was very useful (and it is indeed due to simplified subsumption).

PR #725, opened 2021-07-26
preserve CTYPE pragmas in data declarations

### LINKED ISSUE #689: AST differs when using CAPI CTYPE pragma
**Describe the bug**

ormolu fails to run on files with CTYPE pragmas, e.g. (from the [GHC docs](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/ffi.html#extension-CApiFFI)):

```Haskell
data    {-# CTYPE "unistd.h" "useconds_t" #-} T
```
This produces the error:
```
Test2.hs
    at Test2.hs:1:1-47
```

**To Reproduce**
Make a file that contains the above example and run ormolu on it.

**Expected behavior**
Ormolu should correctly format such files without errors.

**Environment**
 - OS name + version: Debian 10 buster
 - Version of the code: 
     ormolu 0.1.4.1 UNKNOWN UNKNOWN
     using ghc-lib-parser 8.10.4.20210206

amesgen:
Closes #689

This targets the ghc-9.0 branch right now, which will have to change as soon as #722 has been merged.

---

The `dd_cType` field of [`HsDataDefn`](https://hackage.haskell.org/package/ghc-lib-parser-9.0.1.20210324/docs/GHC-Hs-Decls.html#v:dd_cType) had been ignored previously, which means that `CTYPE` pragmas were simply dropped on formatting.

As an aside: The `dd_ctxt` is also ignored, but the corresponding extension ([`DatatypeContexts`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/datatype_contexts.html#extension-DatatypeContexts)) is deprecated, so I don't know if it is even worth to respect it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@
    * Due to a change in Haddock parsing, empty Haddock comments on function
      arguments now get deleted.
 
+* CTYPE pragmas are now preserved. [Issue 689](
+  https://github.com/tweag/ormolu/issues/689).
+
 ## Ormolu 0.1.4.1
 
 * Added command line option `--color` to control how diffs are printed.
--- a/data/examples/declaration/data/ctype-out.hs
+++ b/data/examples/declaration/data/ctype-out.hs
@@ -1,0 +1,1 @@
+data {-# CTYPE "unistd.h" "useconds_t" #-} T
--- a/data/examples/declaration/data/ctype.hs
+++ b/data/examples/declaration/data/ctype.hs
@@ -1,0 +1,1 @@
+data    {-# CTYPE "unistd.h" "useconds_t" #-} T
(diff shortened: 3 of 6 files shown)


amesgen:
Maybe we want to use [safe-wild-cards](https://hackage.haskell.org/package/safe-wild-cards-1.0.0/docs/SafeWildCards.html) in selected places where we want to ensure to actually use all available fields.

mrkkrp:
`safe-wild-cards` looks interesting, but for the sake of simplicity let's try to rely on existing testing methods. There might be fields that are ignored on purpose / are irrelevant yet they come bundled together with important fields like the one from this PR. I'm afraid it's more trouble than it's worth.

mrkkrp (approved):

PR #734, opened 2021-08-02
improve handling of negative literals

### LINKED ISSUE #694: Extra space with negation operator
**Describe the bug**

**To Reproduce**
```hs
f x = 1 / (1 + exp (-x))
```
gets reformatted as
```hs
f x = 1 / (1 + exp (- x))
```

Apparently GHC still treats this as equivalent, but it's very easy to mistakenly read it as a partially applied `-`

**Expected behavior**
`(-x)` should kept as `(-x)`

**Environment**
 - OS name + version:
 - Version of the code:

**Additional context**

### LINKED ISSUE #733: Minus sign in literal patterns is deleted sometimes
While working on #694, I noted that the minus sign in (negative) literal patterns is deleted sometimes. With Ormolu 0.1.4.1 (`--unsafe` necessary):

```haskell
foo = \case - 1 -> 1
```
to
```haskell
foo = \case 1 -> 1
```

and

```haskell
{-# LANGUAGE NoNegativeLiterals #-}

foo = \case -1 -> 1
```
to
```haskell
{-# LANGUAGE NoNegativeLiterals #-}

foo = \case 1 -> 1
```

This is due to us ignoring the third field of `NPat`.

amesgen:
Close #694 
Close #733

Note that CI would fail without the second commit, as the addition of `NegativeLiterals` to `manualExts` triggers #733 in a lot of places.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@
   Previously, if an extension was disabled via the CLI, it could not be
   re-enabled per file.
 
+* `LexicalNegation` is no longer enabled by default. Also, spaces after
+  negation via `-` are removed where possible. [Issue
+  694](https://github.com/tweag/ormolu/issues/694).
+
 ## Ormolu 0.1.4.1
 
 * Added command line option `--color` to control how diffs are printed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@
   negation via `-` are removed where possible. [Issue
   694](https://github.com/tweag/ormolu/issues/694).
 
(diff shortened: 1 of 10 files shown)


brandonchinn178:
Why is this being merged into the 9.0 branch? Does this fix require GHC 9 fixes? If so, would this still fix the issue for pre-GHC 9?

amesgen:
I don't think that this is GHC 9 related, but I am basing all my work off the ghc-9.0 branch to avoid resolving boring conflicts as soon as #722 has been merged into master. But ormolu will continue to work for all older compiler versions, and in particular, the issue will be fixed for them, too.

mrkkrp (approved):

PR #736, opened 2021-08-04
improve detection of blank lines between spans

### LINKED ISSUE #678: Extraneous newline in if-then-else
Formatting

```haskell
x =
  if undefined
    then ()
    -- comment
    else
      ()
```

with 4e01690e0586186735459b7b81eb0d5efbbbbdbe results in the following output:

```haskell
x =
  if undefined
    then ()
    else -- comment

      ()
```

with an extraneous newline after the `else` line.

This only happens with the comment on the line above and the body in hanging position:

```haskell
y =
  if undefined
    then ()
    else -- comment
      ()

z =
  if undefined
    then ()
    -- comment
    else ()
```

is correctly formatted as

```haskell
y =
  if undefined
    then ()
    else -- comment
      ()

z =
  if undefined
    then ()
    else -- comment
      ()
```

This behavior can also be observed in the example that was added by a5c4d6c439b01648cc44e307e0fe10e1a55c9bd5, but it is already present in earlier commits.

amesgen:
Closes #678 

Previously, `needsNewlineBefore` only checked that the two spans are separated by at least one line, but did not account for the possibility of them being non-blank.

---

This fix creates an idempotency issue in stack. Minimized example:

```haskell
test = do
  let test =
        test . -- comment
          test
  -- comment
  test
```
to
```haskell
test = do
  let test =
        test
          . test -- comment
  -- comment
  test
```
to
```haskell
test = do
  let test =
        test
          . test -- comment
          -- comment
  test

```
Previously, the first snippet got formatted into
```haskell
test = do
  let test =
        test
          . test -- comment

  -- comment
  test
```

This seems to be rather orthogonal to this PR, so maybe it is a good idea to create a separate issue for this idempotency failure.

### DIFF
--- a/expected-failures/stack.txt
+++ b/expected-failures/stack.txt
@@ -1,0 +1,18 @@
+src/Stack/Path.hs
+@@ -65,9 +65,9 @@
+              (set (globalOptsL . globalOptsBuildOptsMonoidL . buildOptsMonoidHaddockL) (Just x))
+              . withConfig YesReexec
+              . withDefaultEnvConfig -- FIXME this matches previous behavior, but doesn't make a lot of sense
+-     -- MSS 2019-03-17 Not a huge fan of rerunning withConfig and
+-     -- withDefaultEnvConfig each time, need to figure out what
+-     -- purpose is served and whether we can achieve it without two
+-     -- completely separate Config setups
++             -- MSS 2019-03-17 Not a huge fan of rerunning withConfig and
++             -- withDefaultEnvConfig each time, need to figure out what
++             -- purpose is served and whether we can achieve it without two
++             -- completely separate Config setups
+      runHaddock True $ printKeys with singlePath
+      runHaddock False $ printKeys without singlePath
+
+  Formatting is not idempotent.
+  Please, consider reporting the bug.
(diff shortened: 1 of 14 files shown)


src/Ormolu/Printer/Internal.hs
```diff
@@ -593,3 +603,9 @@ indentStep = 2
 
 isExtensionEnabled :: Extension -> R Bool
 isExtensionEnabled ext = R . asks $ EnumSet.member ext . rcExtensions
+
+----------------------------------------------------------------------------
+-- Extensions
```
mrkkrp, line 608: This should probably say "Blank lines" instead?

expected-failures/stack.txt
```diff
@@ -0,0 +1,18 @@
+src/Stack/Path.hs
```
mrkkrp, line 1: Have you looked at this? Why does it happen?

amesgen: See the description of this PR. I am not sure whether to fix this in this PR or create a separate issue.

mrkkrp:
As discussed off-line this solution adds complexity and it's not clear that it actually a strict improvement due to a new type of idempotence issue it introduces.

PR #738, opened 2021-08-04
Support left-to-right arrow application (fixes #737)

### LINKED ISSUE #737: Left-to-right arrow application is mirrored (AST differs)
**Describe the bug**

Ormolu formats `proc a -> x >- a` to `proc a -> a -< x`, which has a different AST.

**To Reproduce**

```
$ cat test.hs
{-# LANGUAGE Arrows #-}

foo x = proc a -> x >- a
$ ormolu test.hs
test.hs
    at test.hs:3:19-24
$ ormolu --unsafe test.hs
{-# LANGUAGE Arrows #-}

foo x = proc a -> a -< x
```

**Expected behavior**
ormolu should not fail.

Probably, it shouldn't reformat the example at all, though I suppose an argument could be made for normalizing the direction. In that case, the ASTs should be considered equivalent.

**Environment**
current master on macos

robx:
Close #737

Compare https://hackage.haskell.org/package/ghc-8.10.2/docs/src/GHC.Hs.Expr.html#HsCmdArrApp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,9 @@
 * Minus signs in literal patterns are now preserved in all cases. [Issue
   733](https://github.com/tweag/ormolu/issues/733).
 
+* Added support for left-to-right arrow application. [Issue
+  737](https://github.com/tweag/ormolu/issues/737).
+
 ## Ormolu 0.1.4.1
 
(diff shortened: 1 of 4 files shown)


mrkkrp (approved):
To be merged after #722.

PR #755, opened 2021-08-10
format `do` commands like `do` blocks in expression contexts

### LINKED ISSUE #753: Non-idempotence with bracketed `do` commands
**Describe the bug**

This snippet is not idempotent (minimized from [here](https://github.com/hasura/graphql-engine/blob/a0118c99dbdabf24e17008afd5c361d1a896e32a/server/src-lib/Control/Arrow/Extended.hs#L79-L81)):

```haskell
{-# LANGUAGE Arrows #-}

foo = proc x -> (do b -< x)
```
to
```haskell
{-# LANGUAGE Arrows #-}

foo = proc x -> (do
                  b -< x)
```
to
```haskell
{-# LANGUAGE Arrows #-}

foo = proc x ->
  ( do
      b -< x
  )
```

**Expected behavior**
The first snippet should be a fixed point.

**Environment**
Ormolu 0.2.0.0

**Additional context**
This should be fixed when we preserve single vs multiline formatting of `do` commands.

amesgen:
Closes #753 

As we no longer always insert `newline`s after the "do" in `do` commands, we also have to ensure that the left hand side has braces if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 * Infix arrow command formations are formatted like usual operators.
   This fixes [Issue 748](https://github.com/tweag/ormolu/issues/748).
 
+* `do` arrow commands are formatted more flexibly. Fixes [Issue
+  753](https://github.com/tweag/ormolu/issues/753).
+
 ## Ormolu 0.2.0.0
 
 * Now standalone kind signatures are grouped with type synonyms. [Issue
--- a/data/examples/declaration/value/function/arrow/proc-do-complex-out.hs
+++ b/data/examples/declaration/value/function/arrow/proc-do-complex-out.hs
@@ -60,3 +60,6 @@ foo
                   + x
                   * y -- Just do the calculation
               )
+
+traverseA_ f = proc (e, (xs, s)) ->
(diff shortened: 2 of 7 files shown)


mrkkrp:
It looks like do is always hanged, I'm curious why it is not the case in `data/examples/declaration/value/function/arrow/proc-forms2-out.hs`:

```haskell
expr' = proc x -> do
                returnA -< x
        <+> do
                symbol Plus -< ()
                y <- term -< ()
                expr' -< x + y
        <+> do
                symbol Minus -< ()
                y <- term -< ()
                expr' -< x - y
```

which gets formatted into

```haskell
expr' = proc x ->
  do
    returnA -< x
    <+> do
      symbol Plus -< ()
      y <- term -< ()
      expr' -< x + y
    <+> do
      symbol Minus -< ()
      y <- term -< ()
      expr' -< x - y
```

mrkkrp (approved):

amesgen:
The placement code does not see the `do` cmd, but the infix formation, which is not a hanging construct. The same thing happens with `do` blocks without `Arrows`:

```haskell
foo = \x -> do
                x
        <+> do
                x
```
is formatted into
```haskell
foo = \x ->
  do
    x
    <+> do
      x
```

PR #757, opened 2021-08-10
Format deriving clauses with Haddocks idempotently

### LINKED ISSUE #752: Non-idempotence with Haddock comment before deriving clause
This snippet (minimized from [here](https://github.com/hasura/graphql-engine/blob/a0118c99dbdabf24e17008afd5c361d1a896e32a/server/src-lib/Hasura/Backends/BigQuery/Types.hs#L381-L402)) is non-idempotent:

```haskell
data A = A
  -- | B
  deriving (Eq)
```
to
```haskell
data A = A
  deriving (-- | B
            Eq)
```
to
```haskell
data A = A
  deriving
    ( -- | B
      Eq
    )
```

**Expected behavior**
Formatting should be idempotent, but this case is tricky as the comment is an "accidental" Haddock comment (a constructor variant is commented out).

**Environment**
Ormolu 0.2.0.0

amesgen:
Closes #752 

There might be other constructs that are written in a single line by the user, but contain Haddock comments, such that they will have to printed using the multiline layout.

### DIFF
--- a/data/examples/declaration/deriving/haddocks-out.hs
+++ b/data/examples/declaration/deriving/haddocks-out.hs
@@ -1,0 +1,11 @@
+data A = A
+  deriving
+    ( -- | B
+      Eq
+    )
+
+data B = B
+  deriving
+    ( -- | test
+      Eq
+    )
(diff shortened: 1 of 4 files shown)


mrkkrp:
As discussed off-line, this seems like a hacky solution. We can't realistically hope to handle all cases like this "manually". This and other comment-related similar issues will have to wait for a more principled approach to handling comments.

PR #760, opened 2021-08-11
Read and write text files using UTF8 and ignoring native line ending conventions

### LINKED ISSUE #717: Inconsistent line endings between Windows and Unix
**Describe the bug**
Running ormolu on Windows results in CRLF line endings, but running on Unix systems results in LF line endings.

I looked into why this happens and it's because [Data.Text.IO](https://hackage.haskell.org/package/text-1.2.1.1/docs/Data-Text-IO.html) automatically converts `\n` to the system's line ending conventions (this is not a bug in the text package).

**To Reproduce**
Format the same Haskell source code using Windows and a Unix system with `--mode inplace`. Diffing the resulting files will show the line endings are different.

**Expected behavior**
The line ending used by ormolu should be consistent across all platforms. I'd suggest that LF be used. To do this, [`Data.Text.IO.encodeUtf8`](https://hackage.haskell.org/package/text-1.2.1.1/docs/Data-Text-Encoding.html#v:encodeUtf8) can be used to get a `ByteString` with the correct line endings.

**Environment**
 - OS name + version: Windows 10 / Ubuntu 18.04 LTS in WSL2
 - Version of the code: v0.1.4.1

**Additional context**
N/A

amesgen:
Closes #717 

This should affect behavior on Linux/macOS with a UTF8 locale in no way.


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 * `do` arrow commands are formatted more flexibly. Fixes [Issue
   753](https://github.com/tweag/ormolu/issues/753).
 
+* Always read and write formatted source code using UTF8 and ignoring
```
mrkkrp, line 12: Can you link to the relevant issue?

src/Ormolu/Utils/IO.hs
```diff
@@ -0,0 +1,19 @@
...
+  )
+where
+
+import Control.Exception (throwIO)
+import Control.Monad.IO.Class
+import qualified Data.ByteString as B
+import Data.Text (Text)
+import qualified Data.Text.Encoding as TE
+
+writeFileUtf8 :: MonadIO m => FilePath -> Text -> m ()
```
mrkkrp, line 15: Let's add doc strings to these helpers.

src/Ormolu/Utils/IO.hs
```diff
@@ -0,0 +1,19 @@
...
+writeFileUtf8 p = liftIO . B.writeFile p . TE.encodeUtf8
+
+readFileUtf8 :: MonadIO m => FilePath -> m Text
+readFileUtf8 p = liftIO $ either throwIO pure . TE.decodeUtf8' =<< B.readFile p
```
mrkkrp, line 19: Since you're throwing anyway, why not use `TE.decodeUtf8`?

amesgen: I did this out of habit to avoid "hidden" exceptions in lazy values (i.e. with `TE.decodeUtf8`, `void $ readFileUtf8 /path/to/existing/nonutf8file` wil succeed), but I can also change it if this is not considered to be important enough.

PR #762, opened 2021-08-16
Switch to haskell.nix

### LINKED ISSUE #723: Move to haskell.nix?
[haskell.nix](https://github.com/input-output-hk/haskell.nix) is an alternative infastructure for Haskell development via nix. Its [motivation](https://input-output-hk.github.io/haskell.nix/motivation/) gives a good overview, but I think that these two properties are particularly appealing:

 - The cabal package versions are selected by cabal, so we don't have to manually override versions if the default version in the current nixpkgs snapshot does not happen to be what we want.
 - It is very easy to statically link via musl or cross compile to Windows, so we could nixify the release binaries:
   ```
   nix-build -A projectCross.musl64.hsPkgs.ormolu.components.exes.ormolu
   nix-build -A projectCross.mingwW64.hsPkgs.ormolu.components.exes.ormolu
   ```
   Cross compilation to GHCJS is also supported, which might be interesting w.r.t. #724.

A possible downside is that we have to introduce a new binary cache, `hydra.iohk.io`.

mrkkrp:
Close #723.


mrkkrp:
@amesgen Is the generation of release artifacts finished?

amesgen:
> @amesgen Is the generation of release artifacts finished?

Yes, the test run finished successfully in my fork:

 - https://github.com/amesgen/ormolu/actions/runs/1136365274
 - https://github.com/amesgen/ormolu/releases/tag/test-haskell.nix-latest

PR #773, opened 2021-09-13
Improve handling of disabled regions

### LINKED ISSUE #673: DISABLED ormolu remove empty lines

### LINKED ISSUE #708: Error on "-}" in regions with disabled formatting

mrkkrp:
Close #673, close #708, close #714.


src/Ormolu/Processing/Preprocess.hs
```diff
@@ -4,126 +4,71 @@
...
-  String ->
-  -- | Region deltas
-  RegionDeltas ->
-  -- | Literal prefix, pre-processed input, literal suffix, extra comments
-  (String, String, String, [RealLocated String])
-preprocess path input RegionDeltas {..} =
-  go 1 OrmoluEnabled Cpp.Outside id id regionLines
+-- | All lines we are not supposed to format.
+linesNotToFormat :: RegionDeltas -> String -> IntSet
+linesNotToFormat region@RegionDeltas {..} input = unconsideredLines <> disabledLines
```
mrkkrp, line 23: Can you try to insert some newlines so that no line is longer than 80 characters?

src/Ormolu/Parser.hs
```diff
@@ -72,6 +76,58 @@ parseModule Config {..} path rawInput = liftIO $ do
...
+        maybeFlip interleave (Left <$> rawSnippets) (Right <$> regionsToFormat)
+          >>= patchSeparatingBlankLines
+          & dropWhile isBlankRawSnippet
+          & L.dropWhileEnd isBlankRawSnippet
+      patchSeparatingBlankLines = \case
+        Right r@RegionDeltas {..} ->
+          if all isSpace (linesInRegion r rawInput)
+            then [blankRawSnippet]
+            else
+              [blankRawSnippet | isBlankLine (regionPrefixLength)] <> [Right r]
```
mrkkrp, line 98: Redundant parentheses around `regionPrefixLength`?

src/Ormolu/Parser.hs
```diff
@@ -72,6 +76,58 @@ parseModule Config {..} path rawInput = liftIO $ do
...
+      maybeFlip = case regionsNotToFormat of
+        r : _ | regionPrefixLength r == 0 -> id
+        _ -> flip
+      rawSnippets = RawSnippet . T.pack . flip linesInRegion rawInput <$> regionsNotToFormat
+      rawSnippetsAndRegionsToFormat =
+        maybeFlip interleave (Left <$> rawSnippets) (Right <$> regionsToFormat)
+          >>= patchSeparatingBlankLines
+          & dropWhile isBlankRawSnippet
+          & L.dropWhileEnd isBlankRawSnippet
+      patchSeparatingBlankLines = \case
```
mrkkrp, line 93: Can you explain what this function does? And also two lines above:

```haskell
          & dropWhile isBlankRawSnippet
          & L.dropWhileEnd isBlankRawSnippet
```

src/Ormolu/Processing/Common.hs
```diff
@@ -1,27 +1,73 @@
...
+-- | Add indentation to a 'Text'.
+reindent :: Int -> Text -> Text
+reindent i = T.unlines . fmap (T.replicate i " " <>) . T.lines
 
--- | Marker for the end of the region where Ormolu should be disabled.
-endDisabling :: IsString s => s
-endDisabling = "ORMOLU_DISABLE_END -}"
+-- | All lines in the region specified by 'RegionDeltas'.
+linesInRegion :: RegionDeltas -> String -> String
+-- TODO trailing newlines (unlines -> splitOn "\n")
```
mrkkrp, line 39: A forgotten TODO?

src/Ormolu/Processing/Common.hs
```diff
@@ -1,27 +1,73 @@
...
+  [RegionDeltas]
+intSetToRegions total (IntSet.toAscList -> indices) =
+  regionIndicesToDeltas total <$> go Nothing indices
+  where
+    go Nothing [] = []
+    go (Just (a, b)) [] = [RegionIndices (Just a) (Just b)]
+    go Nothing (i : is) = go (Just (i, i)) is
+    go (Just (a, b)) (i : is)
+      | b + 1 == i = go (Just (a, i)) is
+      | otherwise = (RegionIndices (Just a) (Just b)) : go (Just (i, i)) is
```
mrkkrp, line 73: Redundant parentheses?

mrkkrp (commented):
Great work! I have left some comments. We also will need to include the corresponding changelog entries.

src/Ormolu/Parser.hs
```diff
@@ -72,6 +76,58 @@ parseModule Config {..} path rawInput = liftIO $ do
...
          in throwIO (OrmoluParsingFailed loc err)
+  let linesNotToFormat' = linesNotToFormat cfgRegion rawInput
+      regionsToFormat =
+        intSetToRegions rawLineLength $
+          IntSet.fromAscList [1 .. rawLineLength] IntSet.\\ linesNotToFormat'
+      regionsNotToFormat = intSetToRegions rawLineLength linesNotToFormat'
+      maybeFlip = case regionsNotToFormat of
+        r : _ | regionPrefixLength r == 0 -> id
+        _ -> flip
+      rawSnippets = RawSnippet . T.pack . flip linesInRegion rawInput <$> regionsNotToFormat
```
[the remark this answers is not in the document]
amesgen: I inlined `maybeFlip` and added an explanation, is this clearer?

PR #779, opened 2021-09-20
GHC 9.2

### LINKED ISSUE #659: Support nested update syntax of record-dot-preprocessor
Currently, nested update syntax of record-dot-preprocessor is not supported.
So, something like this `company{owner.name = "John"}` results in a parse error.

I think it would be nice to support this as the nested update syntax is part of the record dot syntax proposal.
https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0282-record-dot-syntax.rst

### LINKED ISSUE #705: RDP record updates are converted to regular record updates

### LINKED ISSUE #709: GHC 9.2 records support
[Record Dot Syntax](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0282-record-dot-syntax.rst) proposal landed in GHC 9.2. I expect some work will be needed to correctly (and beautifully) format the code using `OverloadedRecordDot` and `OverloadedRecordUpdate` extensions.

### LINKED ISSUE #768: AST differs on promoted infix data constructor
**Describe the bug**
`ormolu` chokes on promoted infix data constructors.

**To Reproduce**

```bash
$ echo "type X = () '`PromotedInfix` ()" | ormolu
<stdin>
    at <stdin>:1:10-31
```

**Expected behavior**
`ormolu` handles promoted infix data constructors.

**Environment**
 - OS name + version: NixOS 20.03
 - Version of the code: `ormolu` 2.0.0.0

### LINKED ISSUE #794: Switch to ghc-lib-parser-9.2
GHC 9.2 is soon to be released, with a release candidate already available. `ghc-lib-parser-9.2` is likely to resolve some long-standing issues.

amesgen:
Closes #794 
Closes #768 (promoted infix data constructors)
Closes #659 Closes #709 Closes #705 (`RecordDot*` stuff)

---

The most significant in `ghc-lib-parser-9.2` is EPA (exact print annotations). [This blog post](https://blog.shaynefletcher.org/2021/05/annotations-in-ghc.html) gives a good overview. A few points:

 - Most constructs are no longer wrapped in a `Located`, but rather in a `GenLocated (SrcSpanAnn' a)`, where `SrcSpanAnn' ~ (a, SrcSpan)`. [`GHC.Parser.Annotation`](https://downloads.haskell.org/~ghc/9.2.1-rc1/docs/html/libraries/ghc-9.2.0.20210821/GHC-Parser-Annotation.html) gives a good overview of the new combinators for these types.
 - API annotations are contained in the AST, so we can remove our `Anns` stuff.
 - TTG (trees that grow) fields are now way more often actually used for EPAs, rather than being `NoExtField`.

---

EDIT: This PR introduces a benign style issue: #810


amesgen:
Upgrading from 9.2.1-rc1 to 9.2.1 only required to minor changes: https://gist.github.com/amesgen/c77ab54959903ebc5e3734ad630e8c93

The only thing missing now is https://github.com/mrkkrp/ghc-syntax-highlighter/pull/24

mrkkrp:
Is there something that still prevents us from testing with 9.2.1 on CI?

amesgen:
- GHC 9.2.1 is not yet available via haskell.nix.
 - Many dependencies do not yet support GHC 9.2.1.

cabal.project
```diff
@@ -2,13 +2,8 @@ packages: . ormolu-live
...
--- (will be fixed by next ghc-lib-parser release)
--- This removes a redundant dependency on hpc, which
--- is incompatible with haskell.nix+GHCJS.
 source-repository-package
   type: git
-  location: https://github.com/amesgen/stuff
-  tag: 5d173a747118880ded6568001c8d0d56b57ac69f
-  subdir: ghc-lib-parser-9.0.1.20210324
-  --sha256: 0qh07100qvxf0bh2rz64n21cz92wc1pj4niy3xs027l8xv52syz2
+  location: https://github.com/amesgen/ghc-syntax-highlighter
```
mrkkrp, line 7: I released a new version (0.0.8.0) of `ghc-syntax-highlighter` yesterday, do you know how long it takes for `haskell.nix` to pick that up?

src/Ormolu/Diff/ParseResult.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -Wno-incomplete-record-updates #-}
```
mrkkrp, line 10: Why is this necessary?

src/Ormolu/Parser/CommentStream.hs
```diff
@@ -46,21 +48,19 @@ import Ormolu.Utils (onTheSameLine, showOutputable, unSrcSpan)
 newtype CommentStream = CommentStream [RealLocated Comment]
   deriving (Eq, Data, Semigroup, Monoid)
 
--- | Create 'CommentStream' from 'GHC.PState'. The pragmas are
+-- | Create 'CommentStream' from 'PState'. The pragmas are
```
mrkkrp, line 51: It looks like there is a discrepancy between the haddock and the type signature of the function. It doesn't take `PState` anymore, only `HsModule`. This means the per-argument comment also should be updated?

tests/Ormolu/Diff/TextSpec.hs
```diff
@@ -45,7 +45,7 @@ stdTest name pathA pathB = it name $ do
   expectedDiffText <-
     parseRelFile expectedDiffPath
       >>= readFileUtf8 . toFilePath . (diffOutputsDir </>)
-  let Just actualDiff = diffText inputA inputB "TEST"
+  Just actualDiff <- pure $ diffText inputA inputB "TEST"
```
mrkkrp, line 48: Why like this?

src/Ormolu/Utils/Extensions.hs
```diff
@@ -21,6 +22,9 @@ import qualified Distribution.ModuleName as ModuleName
 import Distribution.PackageDescription
 import Distribution.PackageDescription.Parsec
 import qualified Distribution.Types.CondTree as CT
+#if MIN_VERSION_Cabal(3,6,0)
```
mrkkrp, line 25: Rather than introducing CPP, let's use `Cabal >=3.6 && <3.7`.

mrkkrp (commented):
Here is another round of review, since I saw some new changes. I think we are very close to finally merging this into master.

cabal.project
amesgen: I was just about to push a commit to remove this; haskell.nix updates its Hackage snapshot daily.

src/Ormolu/Utils/Extensions.hs
amesgen: Then Ormolu will then no longer be in Stackage, which was the reason why we did not use Cabal 3.6 earlier: https://github.com/tweag/ormolu/pull/759#discussion_r687553174 Personally, I don't care about Ormolu being in Stackage, especially as we have prebuild binaries.

src/Ormolu/Diff/ParseResult.hs
amesgen: Due to this code below:
```haskell
classDeclCtxEq :: TyClDecl GhcPs -> GenericQ ParseResultDiff
classDeclCtxEq tc@ClassDecl {tcdCtxt = Just (L _ [])} tc' = genericQuery tc {tcdCtxt = Nothing} tc'
classDeclCtxEq tc tc' = genericQuery tc tc'
```
But it just occurred to me that this could be rewritten using `RecordWildCards` without an incomplete record update.

src/Ormolu/Utils/Extensions.hs
mrkkrp: I see. Sorry, I forgot about that aspect. In that case let's keep CPP, but let's also move it down so it is sort of standalone and not in the middle of normal imports, so that normal imports can be sorted properly.

tests/Ormolu/Diff/TextSpec.hs
amesgen: Otherwise, I get (but only on GHC 9.2)
```haskell
tests/Ormolu/Diff/TextSpec.hs:48:7: error: [-Wincomplete-uni-patterns, -Werror=incomplete-uni-patterns]
    Pattern match(es) are non-exhaustive
    In a pattern binding:
        Patterns of type ‘Maybe TextDiff’ not matched: Nothing
   |
48 |   let Just actualDiff = diffText inputA inputB "TEST"
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
which I found a bit confusing, as `-Wincomplete-uni-patterns` does not seem to be enabled for the test suite, but it was a boring enough change so I just went with it.

src/Ormolu/Utils/Extensions.hs
mrkkrp: Actually it seems that Stackage ties the version of `ghc-lib-parser` to the version of GHC. This means that they are going to update to `ghc-lib-paser-9.2.x.x` when they switch to GHC 9.2. This will force them also to switch to Cabal-3.6 if I understand correctly. This means that if we are switching to `ghc-lib-parser-9.2.x.x` we can as well simply switch to `Cabal >=3.6 && <3.7` too.

amesgen: I see, will change :+1:

mrkkrp (approved):

PR #783, opened 2021-09-22
Improve `--cabal-default-extensions`

### LINKED ISSUE #781: Crash if unable to find a .cabal file with --cabal-default-extensions
**Describe the bug**

ormolu crashes when it's unable to find a `*.cabal` file, which it searches for when given `--cabal-default-extensions`, if its search comes by `~/.cabal`, which is a directory.

**To Reproduce**

Run `ormolu --cabal-default-extensions SomeFile.hs` in a directory without a cabal file. `ormolu` will search until it finds `~/.cabal` and crash, because it's a directory and not a regular file.

**Expected behavior**

I would expect ormolu to be able to continue even if it couldn't find a cabal file for some of the files given on the command line. It might be nice to issue a warning message on stderr however.

**Environment**
 - OS name + version:
 - Version of the code: 0.3.0.0

amesgen:
Closes #781

The new warning output also revealed another bug with empty `hs-source-dirs`.

### DIFF
--- a/expected-failures/pandoc.txt
+++ b/expected-failures/pandoc.txt
@@ -1,3 +1,4 @@
+Found .cabal file pandoc.cabal, but it did not mention Setup.hs
 src/Text/Pandoc/Readers/Vimwiki.hs
 @@ -618,7 +618,8 @@
       <$ ( skipMany1 spaceChar
@@ -12,3 +13,4 @@ src/Text/Pandoc/Readers/Vimwiki.hs
 
   Formatting is not idempotent.
   Please, consider reporting the bug.
+Found .cabal file pandoc.cabal, but it did not mention test/command/3510-src.hs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## Unreleased
+
+* Improvements to `.cabal` file handling:
+   * When looking for a `.cabal` file, directories were previously
+     erroneously also considered. [Issue 781](
+     https://github.com/tweag/ormolu/issues/781).
+   * We now print a note if Ormolu was told to consider
+     `.cabal` files, but no suitable one could be found.
+   * Handle an empty `hs-source-dirs` correctly.
+   * Also consider modules which are only conditionally listed
+     in the `.cabal` file.
+
 ## Ormolu 0.3.0.0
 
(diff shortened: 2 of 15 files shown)


expected-failures/pandoc.txt
```diff
@@ -1,3 +1,6 @@
+Found .cabal file /build/pandoc-2.14.0.3/pandoc.cabal, but it did not mention Setup.hs
+Found .cabal file /build/pandoc-2.14.0.3/pandoc.cabal, but it did not mention prelude/Prelude.hs
+Found .cabal file /build/pandoc-2.14.0.3/pandoc.cabal, but it did not mention src/Text/Pandoc/Data.hs
```
amesgen, line 3: This file is not detected as it is behind a feature flag: [see here](https://github.com/jgm/pandoc/blob/227f675cc46a680d4b8f2fd71e0c4cbb92a3fd0b/pandoc.cabal#L512)

We could use [`ignoreConditions`](https://hackage.haskell.org/package/Cabal-3.6.1.0/docs/Distribution-Types-CondTree.html#v:ignoreConditions) to also detect modules like this.

amesgen: I could think of a disadvantage when following this approach, so I went with it in 20532c816d2e90384a08060ba6c85fa081bdcc8c.

amesgen: Yes, but I imagine this to be very rare, especially for parsing-related extensions which are relevant for Ormolu. But I could write a variant of `ignoreConditions` which only merges `exposed-modules` etc.

---

Somewhat related: Even right now, a module could be part of multiple components with differing `default-extensions` (in #759, we decided to pick an arbitrary component in this case).

mrkkrp: What I'm thinking about is that if default extensions can be conditional and also modules can be conditional then it makes sense to check if there are groups of extensions/modules that share the same condition. This can get pretty hairy quickly though, as I imagine "or" and "and" can be used in conditions...

mrkkrp: I think what you have right now is good enough. Let's approach this iteratively and address when/if real-world cases are discovered and reported.

mrkkrp (approved):

plcplc:
🎉

PR #796, opened 2021-09-28
Allow ormolu -m check for STDIN content

### LINKED ISSUE #634: Allow `--mode check` when reading from stdin
Checking of stdin would make sense and would ease integration with git hooks (consider a pre-commit hook checking "every touched file should be formatted correctly", and using `git add -p` to commit only some changes to a file. One wants to check the contents of the index, rather than the working tree, i.e. I would like to check the output of `git show :./Mod.hs`).

One work-around is dump the input to a file, then call ormolu on that, but I feel it would be nicer if ormolu would handle this case itself.

A non-workaround is `cat M.hs | ormolu -m check /dev/stdin`, which always fails, since `ormolu` reads the input file twice (once at the start for contents to format, once at the end to check against).

The simple solution is read stdin and cache a copy to compare with the result of formatting it (this is what `brittany` does (somewhere inside `coreIO` in https://github.com/lspitzner/brittany/blob/master/src/Language/Haskell/Brittany/Main.hs#L252)).

I'm happy to submit a patch!

**Other context**
(There was some discussion of this in https://github.com/tweag/ormolu/pull/150#issuecomment-508833403, where @mrkkrp says:
> Now looking at this I'm somewhat reluctant because:
...
> - Some modes of operation do not work with stdin. For example in place editing, as well as checking."

 but it is not clear to me whether this means "checking/in-place-editing stdin makes no sense in general, so reluctant to support stdin at all", or  "the then-proposed/now-current implementation does not support it, but it would be nice if it did"
(My position: in-place-editing of stdin makes no sense in general, but checking does, and would be nice to have))

tbagrel1:
Closes #634 

+ Removes the `ormoluStdin` function
+ Generalizes the `ormoluFile` function, to work on either a real file path,
  or the special string `"<stdin>"`
+ Refactors the `formatOne` function (to avoid repetitions)

**Identified drawbacks**

+ Real file / STDIN switch is made with a comparison to a special string
  (`"<stdin>"`)
+ When using the `in-place` mode with STDIN input, the error message is
  shown only after that the STDIN content has been read

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -11,6 +11,7 @@ import Control.Monad
 import Data.Bool (bool)
 import Data.List (intercalate, sort)
 import Data.Maybe (mapMaybe)
+import qualified Data.Text as T
 import qualified Data.Text.IO as TIO
 import Data.Version (showVersion)
 import Development.GitRev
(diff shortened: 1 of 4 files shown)


app/Main.hs
```diff
@@ -64,45 +64,46 @@ formatOne ::
...
+                Just cabalStdinInputFile ->
+                  getCabalExtensionDynOptions cabalStdinInputFile
+                Nothing -> throwIO OrmoluMissingStdinInputFile
+              else pure []
+          )
+        Just realInputFile ->
+          ( realInputFile,
+            if optUseCabalDefaultExtensions
+              then getCabalExtensionDynOptions realInputFile
+              else pure []
```
amesgen, line 81: The logic `if optUseCabalDefaultExtensions then getCabalExtensionDynOptions ... else pure []` could be shared.

amesgen (commented):
Cool, I recently wanted to use this exact feature [here](https://github.com/tweag/ormolu/blob/c7ede903a0a2a4b5726d0108667f3c14c4b26d0b/.github/workflows/binaries.yml#L62-L63)!

A changelog entry (under `## Unreleased`) would be nice.

> When using the `in-place` mode with STDIN input, the error message is shown only after that the STDIN content has been read

Just for clarity: This is not introduced by this PR, but rather the current behavior. Given that reading the input of Ormolu should never take a relevant amount of time, I think that it is not very important to "fix" this (but it should also not be hard to do).

app/Main.hs
```diff
@@ -60,60 +60,58 @@ formatOne ::
   Mode ->
   -- | Configuration
   Config RegionIndices ->
-  -- | File to format or stdin as 'Nothing'
+  -- | File path supplied on the command line
```
mrkkrp, line 63: I think the old comment was better because it explained the meaning of the argument when it is `Nothing`.

src/Ormolu.hs
```diff
@@ -4,8 +4,8 @@
 -- | A formatter for Haskell source code.
 module Ormolu
   ( ormolu,
-    ormoluFile,
```
mrkkrp, line 7: I think we should go by the path of the minimal API change. For this kind of PR, I would expect the API to stay the same.  In general, it is a good idea to make your change least disruptive unless there is a good reason to refactor. IMO, having both `ormoluStdin` and `ormoluFile` is a reasonable choice and is marginally better than having an extra wrapper type just for this one function. As you pointed out elsewhere, the code in the docs is obsolete though and it is probably best to remove it altogether so that there is nothing in the docs that can go out of date easily. Also, this module is expected to be imported unqualified, this is why the main functions have `ormolu` in their names. `readAndFormat` is too general for a name. I understand this is a Haskell-specific naming habit and in languages like OCaml you would be indeed encouraged not to repeat module name in the names of functions.

app/Main.hs
```diff
@@ -60,60 +60,58 @@ formatOne ::
...
   Maybe FilePath ->
   IO ExitCode
 formatOne CabalDefaultExtensionsOpts {..} mode config mpath =
-  withPrettyOrmoluExceptions (cfgColorMode config) $
```
mrkkrp, line 67: I think the best way to implement the feature would be abstracting what happens in the `Check` case for reading from file (define a helper function for that) and then putting that in the `Check` case for stdin. The stdin vs file cases are sufficiently different to have different sub-`case`s matching modes.

app/Main.hs
tbagrel1: Yes, I agree on this. I changed it because of a misunderstanding so, I will revert the change :)

src/Ormolu.hs
tbagrel1: I agree with the minimal API change, but atm, when using ormolu on a file, this file needs to be read twice: the first time to store the original input (for diff later), and the second time in `ormoluFile` for formatting. In addition, I will be forced to change the API of `ormoluStdin`, because I need to store the original input (for diff in check mode), and AFAIK I can't read STDIN twice.
So in the first version I wrote (which I didn't commit), I just changed `ormoluStdin` to also returns the original input read from STDIN, but then there was almost a1:1 copy between code for file processing and STDIN processing, so I tried to improve it.
In fact, `ormoluStdin` and `ormoluFile` don't do much except hiding the original content read, which is useful in check mode, so I'm not sure what to do.

app/Main.hs
tbagrel1: So, we would have,

- `ormoluFile` for in-place and STDOUT modes on file
- `ormoluStdin` for STDOUT mode on STDIN
- (new) `ormoluCheck` for check mode on either file or STDIN?

I don't really like the idea of splitting helper functions on two axes at the same time (mode and type of source), but I will do it if you think it's better

src/Ormolu.hs
mrkkrp: Thanks, this is helpful. In that case let's keep the old `ormoluStdin` in the `Ormolu` module for people who potentially might want that function from calling from Haskell, but inline it in `Main.hs` this way we'll have the original input for comparison.

app/Main.hs
mrkkrp: Let's leave handling of files in all modes remains as it is, and make a helper function out of this fragment:

```haskell
            case diffText originalInput formattedInput inputFile of
              Nothing -> return ExitSuccess
              Just diff -> do
                runTerm (printTextDiff diff) (cfgColorMode config) stderr
                -- 100 is different to all the other exit code that are emitted
                -- either from an 'OrmoluException' or from 'error' and
                -- 'notImplemented'.
                return (ExitFailure 100)
```

Then use that function to handle diffing for both files and stdin. Formatting of `stdin` can be done with an inlined version of `ormoluInline`.

mrkkrp (approved):

PR #798, opened 2021-10-01
Fix multiline pattern bind indent (#712)

### LINKED ISSUE #712: Layout is sometimes indented less than the line of the layout trigger
Ormolu currently formats this way:

```
foo | x <-
        5 = case x of
  5 -> True
  _ -> False
```

The combination of the extra indent from wrapping the guard, and the case layout to the left of that indent, is very difficult to read.  Lots of options would be better, including:

1. Always Indenting a layout context at least as far as the indent on the line where the layout trigger occurs.
2. Insisting on a newline after the `=`, so that this doesn't arise in the first place.

I don't have strong preference between the two, but the current behavior was a real WTF moment.

tbagrel1:
Close #712.

I don't think this solution is acceptable, but it's still a starting
point to see that it can be made if we really want it.

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -248,17 +248,22 @@ p_match' placer render style isInfix strictness m_pats GRHSs {..} = do
           endOfPats
       placement =
         case endOfPats of
-          Nothing -> blockPlacement placer grhssGRHSs
-          Just spn ->
-            if onTheSameLine spn grhssSpan
-              then blockPlacement placer grhssGRHSs
-              else Normal
+          Just spn
+            | any guardNeedsLineBreak grhssGRHSs
+                || not (onTheSameLine spn grhssSpan) ->
+              Normal
+          _ -> blockPlacement placer grhssGRHSs
+      guardNeedsLineBreak :: Located (GRHS GhcPs body) -> Bool
+      guardNeedsLineBreak (L _ (GRHS _ guardLStmts _)) = case guardLStmts of
+        [] -> False
+        [L l _] -> not (isOneLineSpan l)
+        _ -> True
       p_body = do
         let groupStyle =
               if isCase style && hasGuards
                 then RightArrow
                 else EqualSign
-        sep newline (located' (p_grhs' placer render groupStyle)) grhssGRHSs
+        sep breakpoint (located' (p_grhs' placer render groupStyle)) grhssGRHSs
       p_where = do
         let whereIsEmpty = eqEmptyLocalBinds (unLoc grhssLocalBinds)
         unless (eqEmptyLocalBinds (unLoc grhssLocalBinds)) $ do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 * Allow check mode (`-m check`) when working with STDIN input ([Issue 634](
   https://github.com/tweag/ormolu/issues/634)).
+* Now guards are printed on a new line if at least one guard is multiline or
+  if all guards together occupy more than one line. The body of each guard
+  is also indented one level deeper in that case. [Issue
+  712](https://github.com/tweag/ormolu/issues/712).
 
 ## Ormolu 0.3.0.1
 
--- a/data/examples/declaration/value/function/multiple-guards.hs
+++ b/data/examples/declaration/value/function/multiple-guards.hs
@@ -1,7 +1,6 @@
 foo :: Int -> Int
-foo x
-  | x == 5 = 10
-  | otherwise = 12
+foo x | x == 5 = 10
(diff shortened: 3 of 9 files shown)


src/Ormolu/Printer/Internal.hs
```diff
@@ -26,6 +26,7 @@ module Ormolu.Printer.Internal
     enterLayout,
     vlayout,
     getLayout,
+    setAnchor,
```
mrkkrp, line 29: Try to solve the issue differently without introducing new primitives. We should try to keep our vocabulary of combinators from exploding and try to work within the DSL that we have. Look at my comments in #712 they should suggest how this can be solved and what layout is desirable.

mrkkrp (changes requested):
Also, for every change like this a changelog entry and a new test(s) are necessary.

src/Ormolu/Printer/Internal.hs
mrkkrp: I'm sorry for unclear communication from my side, this alignment feature is not what I had in mind. As a rule, we do not align anything in Ormolu, so `y` would never have to be aligned with `x` because if `x` moved `y` would also need to move—this is exactly what we try to avoid in order to help with minimization of diffs.

tbagrel1: No problem, we didn't really discover the case with multiple pattern binds at first. At least, it was a fun exercise to add this functionality in ormolu, and helped me to get a better understanding of how internals work

tbagrel1:
Ok, I think I found a way to achieve the requested behavior, and it doesn't seem to introduce changes except in the desired place when applied to the ormolu codebase (via `format.sh`)
However, some tests don't pass (which is something I expected), but I don't know where to start to see whether I need to update the tests, or improve my fix.

tbagrel1:
I think it works as expected now.
I also added tests and changelog entry.

src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -249,10 +249,17 @@ p_match' placer render style isInfix strictness m_pats GRHSs {..} = do
...
+          Just spn -> case style of
+            Function _ | hasGuards && any guardNeedsLineBreak grhssGRHSs -> Normal
+            _ ->
+              if onTheSameLine spn grhssSpan
+                then blockPlacement placer grhssGRHSs
+                else Normal
+      guardNeedsLineBreak :: Located (GRHS GhcPs body) -> Bool
+      guardNeedsLineBreak (L _ (GRHS _ guardLStmts _)) = case guardLStmts of
+        [] -> False
+        [(L l _)] -> not $ isOneLineSpan l
```
amesgen, line 261: Nitpick: The braces are not necessary here.
```suggestion
        [L l _] -> not $ isOneLineSpan l
```

src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -249,10 +249,17 @@ p_match' placer render style isInfix strictness m_pats GRHSs {..} = do
       placement =
         case endOfPats of
           Nothing -> blockPlacement placer grhssGRHSs
-          Just spn ->
-            if onTheSameLine spn grhssSpan
-              then blockPlacement placer grhssGRHSs
-              else Normal
+          Just spn -> case style of
+            Function _ | hasGuards && any guardNeedsLineBreak grhssGRHSs -> Normal
```
amesgen, line 253: I am not sure if this was discussed, but is it intended that this change is only applied for the `Function` style? Example: Even with this PR, the input
```haskell
\case x | x <-
          foo -> case x of
         5 -> True
         _ -> False
```
is formatted to
```haskell
\case
  x | x <-
        foo -> case x of
    5 -> True
    _ -> False
```
(note that there is no linebreak before the `|`). Simply ignoring `style` passes all tests locally, and the output would be
```haskell
\case
  x
    | x <-
        foo -> case x of
      5 -> True
      _ -> False
```
which now is consistent with the output for regular function match groups.

tbagrel1: That's a good point indeed, and we didn't discussed it before. However, I don't have the authority to decide whether we should extend it to all styles or not :)
But IMO, it's better to extend it!

amesgen (approved):
mplementation looks good!

mrkkrp (approved):
Great work! This was not an easy one to figure out. I took the liberty of adjusting this a little bit:

* The new examples that you have added should be without the module headers and the main function, because those parts are completely optional. In fact, they only add visual noise. I removed them.
* There is one interesting corner case that you have skipped: all guards in a single line. I added this example and it showed that we had a bug in our guard printing code! We need to separate guards with breakpoints, not with newlines.
* Please format your code so that it fits in 80 line limit. Try to avoid very long lines.

PR #799, opened 2021-10-06
Backpack support

### LINKED ISSUE #600: Backpack .hsig file support
see https://wiki.haskell.org/Module_signature

tbagrel1:
Closes #600

Co-authored-by: amesgen <amesgen@users.noreply.github.com>

~~The blockers for this work are explained in https://github.com/tweag/ormolu/issues/600#issuecomment-939895929~~ (no longer valid)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   806](https://github.com/tweag/ormolu/issues/806).
 * Now a space is forced after `--` in line comments. [Issue
   808](https://github.com/tweag/ormolu/issues/808).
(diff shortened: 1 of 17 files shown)


amesgen:
Concerning the CPP support: This is a side effect of the recent rework of CPP handling, as everything above and below the CPP section is formatted separately, and (in contrast to `parseModule`), `parseSignature` seems to be more strict, by requiring a signature header on top. This could be worked around by manually inserting (and later removing) a bogus signature header.

tbagrel1:
Ok, I will try to do this this afternoon

PR #805, opened 2021-10-15
Make it easier to regenerate test output

brandonchinn178:
Running this command now regenerates output files, making it easier to regenerate all the files after making a change (to be inspected with `git diff` afterwards, of course)
```bash
REGENERATE_PRINTER_OUTPUT=true stack build
```


mrkkrp:
Thanks!

PR #809, opened 2021-10-19
Force a space after -- in line comments

### LINKED ISSUE #808: Add a rule that formats missing spaces after `--` characters

As far as I can tell, there's no good tool for catching missing spaces after comments in Haskell. For example, when code reviewing, I sometimes see things like this:
```
--this is a comment
```
when I think most folks would prefer
```
-- this is a comment
```

This type of comment isn't super fun to make in code reviews, since it's more of a nit than anything, and it seems like something that should be enforced by a linting/formatting tool. Other formatters, such as eslint and SwiftLint have such rules, and it'd be awesome if fourmolu could support something similar.

Let me know if there's anything helpful I can add to this issue!

As I said in the description, I'd like it if Ormolu would support adding spaces in front of the comment, itself, but after the `--` characters.

I looked at `HLint`, since other languages use linters for this type of thing, but the [folks at HLint think it's more appropriate to do with a formatter](https://github.com/ndmitchell/hlint/issues/522).

**Additional context**

So this issue comes with a PR (LINK TODO), and I already [implemented this change in fourmolu](https://github.com/fourmolu/fourmolu/pull/120), but the folks there figured that such a rule may be applicable to Ormolu as well, so I figured I'd run it past y'all and see if this seems like something that would be worth adding to Ormolu and upstreaming to Fourmolu.  If not, I'll just add it there.

dmarticus:
Close #808.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+* Now a space is forced after `--` in line comments. [Issue
+  808](https://github.com/tweag/ormolu/issues/808).
+
 ## Ormolu 0.3.1.0
 
 * Allow check mode when working with stdin input. [Issue 634](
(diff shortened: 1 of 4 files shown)


mrkkrp:
Thanks!

PR #820, opened 2021-11-01
Add commit hash alias for Ormolu Live

amesgen:
Trivial workaround as `--prod` and `--alias` can not be combined.

### DIFF
--- a/ormolu-live/deploy.sh
+++ b/ormolu-live/deploy.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
 set -eo pipefail
-[[ $BUILDKITE_BRANCH == "master" ]] && FLAGS="--prod" || true
-netlify deploy $FLAGS -d $(nix-build -A ormoluLive.website)
+ORMOLU_LIVE=$(nix-build -A ormoluLive.website)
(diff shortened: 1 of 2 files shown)


mrkkrp:
The CI failure seems odd. Do you know why it happens?

amesgen:
They seem to happen from time to time with GHCJS. I have also (quite rarely) seen them locally.

ormolu-live/deploy.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 set -eo pipefail
 [[ $BUILDKITE_BRANCH == "master" ]] && FLAGS="--prod" || true
-netlify deploy $FLAGS -d $(nix-build -A ormoluLive.website)
+ORMOLU_LIVE=$(nix-build -A ormoluLive.website)
+netlify deploy --alias=$(git log -1 --format="%H") -d $ORMOLU_LIVE
+netlify deploy $FLAGS -d $ORMOLU_LIVE
```
mrkkrp, line 6: Maybe we could replace `$FLAGS` by `--prod` directly since `deploy.sh` is only called when we are on `master` anyway?

PR #824, opened 2021-11-02
Preserve rewrite rules that are never active

### LINKED ISSUE #823: `[~]` phase control of a rewrite rule is removed.
**Describe the bug**
A "never active" phase control annotation (`[~]`, see [here](https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/rewrite_rules.html#syntax) why this exists) for rewrite rules is removed by Ormolu, leading to differing AST:

```haskell
{-# RULES "map-loop" [ ~  ]  forall f . map' f = map' (id . f) #-}
```
to
```haskell
{-# RULES "map-loop" forall f. map' f = map' (id . f) #-}
```

**Additional context**
Found via #821

amesgen:
Closes #823

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Signature.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Signature.hs
@@ -123,7 +123,7 @@ p_inlineSig name InlinePragma {..} = pragmaBraces $ do
     ConLike -> txt "CONLIKE"
     FunLike -> return ()
   space
-  p_activation inl_act
+  when (inl_act /= NeverActive) $ p_activation inl_act
   space
   p_rdrName name
 
@@ -157,7 +157,7 @@ p_inlineSpec = \case
 
 p_activation :: Activation -> R ()
 p_activation = \case
-  NeverActive -> return ()
+  NeverActive -> txt "[~]"
   AlwaysActive -> return ()
   ActiveBefore _ n -> do
     txt "[~"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
   differing ASTs in some special cases. [Issue
   726](https://github.com/tweag/ormolu/issues/726).
 
+* Rewrite rules that are never active are now formatted correctly.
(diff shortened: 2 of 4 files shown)


src/Ormolu/Printer/Meat/Declaration/Signature.hs
```diff
@@ -123,7 +123,7 @@ p_inlineSig name InlinePragma {..} = pragmaBraces $ do
     ConLike -> txt "CONLIKE"
     FunLike -> return ()
   space
-  p_activation inl_act
+  when (inl_act /= NeverActive) $ p_activation inl_act
```
amesgen, line 126: This is necessary as for some reason, this code
```haskell
{-# NOINLINE foo #-}
```
is parsed as having activation `NeverActive`, but 
```haskell
{-# NOINLINE [~] foo #-}
```
is invalid. This does not happen with e.g. `INLINE`, and is also not fixed by GHC 9.2. Maybe this is worth reporting/asking if it is intentional?

mrkkrp: Yeah, it is worth asking. I believe the issue tracker lives here now: https://gitlab.haskell.org/ghc/ghc/-/issues

mrkkrp:
We will need a changelog entry for this, too.

src/Ormolu/Printer/Meat/Declaration/Signature.hs
amesgen: https://gitlab.haskell.org/ghc/ghc/-/issues/20603

amesgen: I got an explanation in the GHC issue, this is intentional behavior: the activation is not related to the `InlineSpec` (which is `NoInline` in this case), so `NeverActive` for this pragma does *not* mean that (as I thought and found confusing) that `NOINLINE` is never active. Rather, it means that inlining is never active.

mrkkrp (approved):

Issue #528, opened 2020-03-27
do notation in right operand looks ambiguous


roberth:
Ormolu formats to this code, which isn't very clear:

```
y = do
  _ <- Nothing
  undefined
  <|> do
    Just 2
```

If you're not intimately familiar with `do` syntax, you may wonder whether `y` is `Nothing` or `Just 2`.

This layout is easier to understand:

```
y = do
    _ <- Nothing
    undefined
  <|> do
    Just 2
```

mrkkrp:
I think we should try to shift the infix operator one indentation step to the left in this particular case. As the last resort we could also shift half-step, but we need a good reason to do that.

tbagrel1:
@roberth, yes, I suppose the diff is smaller when going from single to multi-line `do` if we stay consistent for both single-line and multi-line `do` blocks.

I currenlty have a hard time figuring out what would be the desired output of a chain of {`do` block} `op` {`do` block} `op` {`do` block}:

Option 1:
```haskell
x = do
  something
 <> do
  something
 <> do
  something
```

Option 2:
```haskell
x = do
  something
 <> do
    something
 <> do
    something
```

Option 3:
```haskell
x = do
  something
 <> do
    something
   <> do
      something
```

roberth:
Shouldn't the first `<>` in your current implementation be non-indented because it is technically in a hanging context?

```haskell
x =
  do
    something
   <> do    -- unnecessary indent
      something
    <> do   -- correct given context but still indented too much because of first <>'s rhs
      something
    <> do   -- correct given context but still indented too much because of first <>'s rhs
      something
```

mrkkrp:
Good question. It's hard to tell :-D Seriously though, It looks like `(:)` will be applied to the whole do block. And yes, that's what I had in mind. Simply because it is easier to spot `(:)` than deduce that all operands are do blocks (what to do with a mixed sequence?).

tbagrel1:
At the beginning, I tried something like this, which would work even with mixed sequences
```haskell
x =
  1 :
  do
    2
   :
  do
    3
   :
  do
    [4]
```
But I got parsing errors when the do block is in hanging position, and it would be hard to implement cleanly

mrkkrp:
Can we have this rule: if at least one operand in an operator chain is a do block, then do this:

```haskell
x =
  1
 : do
  blah
  blah
 : do
  foo
  bar
```

Frankly, I think this is not a common kind of expression, so we can do whatever is easier to implement in a consistent way.

Issue #342, opened 2019-08-28
Ormolu built with nix does not show branch and commit it was compiled from


mrkkrp:
Somewhat minor, but still when `ormolu` is compiled using nix, we get this:

```
mark@fou ~/p/t/ormolu> ormolu --version
ormolu 0.0.1.0 UNKNOWN UNKNOWN
using ghc 8.6.5
```

It would be nice to get actual branch and commit info here.

mrkkrp:
This becomes more of a nuisance now that we're building release binaries with Nix. It would be nice for the binaries to show correct branch and commit info, e.g.:

```
$ ormolu --version
ormolu 0.2.0.0 amesgen/issue-517 4889228fa09f1651c9bbdfd7d7fefc09fa3125a1
using ghc-lib-parser 9.0.1.20210324
```

PR #830, opened 2021-11-03
Improve formatting of operator chains

### LINKED ISSUE #690: Inconsistency in the number of operators per line
**Describe the bug**

Input:

```haskell
foo gen =
  Foo <$> gen <*> gen <*> gen <*> gen <*> gen <*> gen
      <*> gen <*> gen <*> gen <*> gen <*> gen<*> gen <*> gen
```

Output:

```haskell
foo gen =
  Foo <$> gen <*> gen <*> gen <*> gen <*> gen <*> gen
    <*> gen
    <*> gen
    <*> gen
    <*> gen
    <*> gen
    <*> gen
    <*> gen
```

**Expected behavior**
I would have expected that Ormolu treated the first line and the second line of the input the same way.

**Environment**
34bdf62429768f24b70d0f8ba7730fc4d8ae73ba

### LINKED ISSUE #785: Idempotency issue involving operators and case
```
$ cat repro.hs
foo = a &&
    b
    <> case d of
         e -> e
    && f

$ ormolu -c repro.hs
repro.hs
@@ -3,4 +3,4 @@
      && b
      <> case d of
        e -> e
-       && f
+     && f

  Formatting is not idempotent.

$ ormolu --version
ormolu 0.3.0.0 HEAD 06b767c8b70e60321a4debc4599127446f958e5b
using ghc-lib-parser 9.0.1.20210324
```

This is the smallest repro i could find. When changing `&&` to `<>`, the issue does not reproduce.

### LINKED ISSUE #825: Unable to format ":" chains containing do blocks
Ormolu is currently unable to format this (`Parsing of formatted code failed`):

```haskell
module A (a) where

b :: [Int]
b =
  1 :
    do
      2
     : do
        3
       : do
          [4]
```

With `--unsafe`, we get the following formatted output:

```haskell
module A (a) where

b :: [Int]
b =
  1 :
  do
    2 :
  do
    3 :
    do
      [4]
```

### LINKED ISSUE #826: Improvement of operator chain handling
At the point where Ormolu works, operator fixities and precedence levels are not known, so the operator call trees (`OpApp`) in the AST might not match the actual order of operations. Currently, we do some "educated guesswork" to reform the ordered binary call tree, and special-case the formatting of some specific operators (`:`, `$`).

As an attempt to improve the operator chain formatting, I propose new areas of work:

+ Extracting operator fixity/precedence information from Hoogle or Hackage, to improve heuristics of the formatter, and to be able to "display" the actual call tree using indents in the formatted output
+ Transforming operator call trees from binary trees to n-ary trees, where every operator with the same precedence level is located at the same level in the tree

A PoC is in progress :)

tbagrel1:
Close #826, close #785, close #690, close #825.


tbagrel1:
When we have something like this:
```haskell
a =
  1 + 2 + do
    3
```
Should it stay as it is, or because the layout of the whole tree is multiline, should it be formatted to
```haskell
a =
  1
    + 2
    + do
      3
```
?

mrkkrp:
I would say it should stay as is, it is also the status quo.

data/examples/declaration/type-synonyms/multi-line-out.hs
```diff
@@ -13,6 +13,10 @@ type Foo =
...
-  "route1" :> ApiRoute1
-    :<|> "route2" :> ApiRoute2 -- comment here
-    :<|> OmitDocs :> "i" :> ASomething API
+  "route1"
+    :> ApiRoute1
+    :<|> "route2"
+    :> ApiRoute2 -- comment here
+    :<|> OmitDocs
+    :> "i"
+    :> ASomething API
```
tbagrel1, line 22: With the default fixity map extracted from Hoogle, `:>` and `:<|>` can't be ordered, so the expected behaviour is to place each operation on a new line as the whole operator chain is multiline

data/examples/declaration/value/function/infix/lenses-out.hs
```diff
@@ -1,9 +1,12 @@
...
     M.fromList $
-      "type" .= ("user.connection" :: Text)
-        # "connection" .= uc
-        # "user" .= case name of
+      "type"
+        .= ("user.connection" :: Text)
+        # "connection"
+        .= uc
+        # "user"
+        .= case name of
```
tbagrel1, line 9: With the default fixity map extracted from Hoogle, `.=` and `#` can't be ordered, so the expected behaviour is to place each operation on a new line as the whole operator chain is multiline

data/examples/declaration/value/function/list-notation-1-out.hs
```diff
@@ -1,8 +1,10 @@
...
-      case updateGlobalSecondaryIndexUpdates a of
-        [] -> []
-        l -> ["GlobalSecondaryIndexUpdates" .= l]
+      "TableName"
+        .= updateTableName a
+        : "ProvisionedThroughput"
+        .= updateProvisionedThroughput a
+        : case updateGlobalSecondaryIndexUpdates a of
+          [] -> []
+          l -> ["GlobalSecondaryIndexUpdates" .= l]
```
tbagrel1, line 10: same thing here again

format.sh
```diff
@@ -10,6 +10,6 @@ export LANG="C.UTF-8"
 cabal format
 
 ormolu -e -m inplace $(find app -type f -name "*.hs")
-ormolu -e -m inplace $(find src -type f \( -name "*.hs" -o -name "*.hs-boot" \))
+ormolu -e -m inplace $(find src -type f \( -name "*.hs-boot" -o -name "*.hs" ! -path 'src/Ormolu/Printer/OperatorFixityMap.hs' \))
```
tbagrel1, line 13: Ormolu is currently unable to format `OperatorFixityMap.hs` because of its size

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -269,22 +267,6 @@ p_conDeclField ConDeclField {..} = do
   breakpoint
   sitcc . inci $ p_hsType (unLoc cd_fld_type)
 
-tyOpTree :: LHsType GhcPs -> OpTree (LHsType GhcPs) (Located RdrName)
```
tbagrel1, line 272: I moved `tyOpTree` and `p_tyOpTree` to the `Value.hs`, because they have a lot of symmetry with `cmdOpTree`, `exprOpTree` and `p_cmdOpTree`, `p_exprOpTree` and also need to reuse some helper functions defined but not exposed by `Value.hs`. If this is not acceptable, maybe we could create an `OpTree.hs` file?

tbagrel1 (commented):
About examples where operator fixities cannot be ordered properly:

At the moment, with the non-configurable context-unaware "default" fixity map, we loose a lot of information about operators because of conflicting definitions. However, in a few days, I will improve the fixity map construction, by allowing

+ user custom fixity declarations
+ package detection from `.cabal` files
+ manual package selection
+ and more...

which should give way better results requiring little to no explicit configuration.
The fixity map generation will also be rewritten in Haskell.

tbagrel1:
The new formatting logic also solves open idempotency issues:
+ https://github.com/tweag/ormolu/issues/785

data/examples/declaration/type-synonyms/multi-line-out.hs
mrkkrp: Can you explain why they "can't be ordered"? IIRC they should come from Servant with `:<|>` being infix right 3 and `:>` infix right 4, so `:>` should bind more tightly. Or are there other packages defining the same operators with conflicting fixities/precedence?

data/examples/declaration/value/function/operator-chains-out.hs
```diff
@@ -0,0 +1,245 @@
+module A (a) where
```
mrkkrp, line 1: This file is too long. Whenever something fails in it it'll be very hard to locate the problem. Split it into several files, once per case.

data/examples/declaration/value/function/operator-chains-out.hs
```diff
@@ -0,0 +1,245 @@
...
+a =
+  1
+    ++ 2
+    ++ 3
+    ++ 4
+
+-- Right chain, : case, 1 operator type
+b :: Int
+b =
+  1
```
mrkkrp, line 26: Is it really necessary to repeat the same operator so many times? I think just 2 operators are enough in cases like this.

data/examples/declaration/value/function/operator-chains-out.hs
```diff
@@ -0,0 +1,245 @@
...
+      3 $
+        4 $
+          5 $
+            6 $
+              7 $
+                8
+
+-- Right chain, normal case, 2 operators with p(a) == p(b)
+e :: Int
+e =
```
mrkkrp, line 49: What about:

```
e =
  1
    ^^ 4
    ^ 5
    ^^ 8
```

data/examples/declaration/value/function/operator-chains-out.hs
```diff
@@ -0,0 +1,245 @@
...
+    ^^ 4
+    : 5
+    ^^ 6
+    ^^ 7
+    ^^ 8
+
+-- Right chain, : case, 2 operators with p(a) < p(:)
+m :: Int
+m =
+  1 `seq`
```
mrkkrp, line 146: Why is `seq` put at the end of the line? Only `$` gets this treatment.

mrkkrp (commented):
Thanks for making such wonderful progress on this! I did the first pass and left some comments. I did not look at the implementation yet. At this point we want to make sure that the behavior is right. Once we are happy with that we can focus more on how this is implemented.

mrkkrp:
Also please note that we have merged https://github.com/tweag/ormolu/pull/779 and this PR needs to be rebased on master.

data/examples/declaration/value/function/operator-chains-out.hs
tbagrel1: (Answering to your comment about why ":>" and ":<|>" can't be ordered, but for some reason in the Github interface I can only answer to this (just next) comment)

This is what I get from my hoogle/hackage extraction:
```haskell
      ( ":>",
        -- infixl 5 (7876/24508 = 32.1%) in lens (2096/24508 = 8.6%), containers (1489/24508 = 6.1%), fingertree (790/24508 = 3.2%), diagrams-lib (753/24508 = 3.1%), optics-core (691/24508 = 2.8%), rio (626/24508 = 2.6%), optics-extra (598/24508 = 2.4%), preamble (259/24508 = 1.1%), hgeometry-combinatorial (148/24508 = 0.6%), microformats2-parser (65/24508 = 0.3%), sized (58/24508 = 0.2%), hw-fingertree-strict (46/24508 = 0.2%), hw-fingertree (46/24508 = 0.2%), sequence (45/24508 = 0.2%), sitepipe (35/24508 = 0.1%), knit-haskell (34/24508 = 0.1%), lens-utils (29/24508 = 0.1%), strict-containers (22/24508 = 0.1%), raw-feldspar (19/24508 = 0.1%), planet-mitchell (12/24508 = 0.0%), polysemy-optics (10/24508 = 0.0%), kmonad (5/24508 = 0.0%)
        FixityInfo {fixDir = Just InfixL, fixMinPrec = 5, fixMaxPrec = 5}
        -- infixl 9 (6002/24508 = 24.5%) in futhark (2353/24508 = 9.6%), opaleye (1589/24508 = 6.5%), parsley-core (535/24508 = 2.2%), quickcheck-state-machine (244/24508 = 1.0%), parameterized-utils (231/24508 = 0.9%), jukebox (194/24508 = 0.8%), streams (188/24508 = 0.8%), wumpus-basic (104/24508 = 0.4%), parsley (102/24508 = 0.4%), haskus-utils-data (82/24508 = 0.3%), feldspar-language (69/24508 = 0.3%), presburger (60/24508 = 0.2%), liboleg (57/24508 = 0.2%), type-aligned (52/24508 = 0.2%), hobbits (32/24508 = 0.1%), yices (26/24508 = 0.1%), kansas-lava (22/24508 = 0.1%), linear-maps (15/24508 = 0.1%), serv (14/24508 = 0.1%), data-sketches-core (13/24508 = 0.1%), exposed-containers (7/24508 = 0.0%), spine (5/24508 = 0.0%), VecN (5/24508 = 0.0%), fingertree-tf (3/24508 = 0.0%), language-sh (?/24508 = ?%), joinlist (?/24508 = ?%), darcs-beta (?/24508 = ?%), clean-unions (?/24508 = ?%), ForSyDe (?/24508 = ?%)
          <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 9, fixMaxPrec = 9}
        -- infixr 5 (4971/24508 = 20.3%) in clash-prelude (2290/24508 = 9.3%), massiv (1047/24508 = 4.3%), streaming (680/24508 = 2.8%), generic-lens-core (622/24508 = 2.5%), biohazard (114/24508 = 0.5%), sdp (85/24508 = 0.3%), ribbit (63/24508 = 0.3%), forsyde-shallow (41/24508 = 0.2%), linear-base (29/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 5, fixMaxPrec = 5}
        -- infixr 4 (2238/24508 = 9.1%) in servant (1330/24508 = 5.4%), prolude (543/24508 = 2.2%), gogol-core (183/24508 = 0.7%), servant-js (147/24508 = 0.6%), ocaml-export (35/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 4, fixMaxPrec = 4}
        -- infixr 3 (1418/24508 = 5.8%) in HaXml (621/24508 = 2.5%), pandora (599/24508 = 2.4%), PrimitiveArray (161/24508 = 0.7%), joint (37/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 3, fixMaxPrec = 3}
        -- infix 0 (948/24508 = 3.9%) in extensible (879/24508 = 3.6%), membership (69/24508 = 0.3%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 0, fixMaxPrec = 0}
        -- infixr 1 (342/24508 = 1.4%) in darcs (342/24508 = 1.4%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 1, fixMaxPrec = 1}
        -- infixr 0 (163/24508 = 0.7%) in lorentz (80/24508 = 0.3%), llvm-pretty (56/24508 = 0.2%), indigo (27/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 0, fixMaxPrec = 0}
        -- infixr 6 (161/24508 = 0.7%) in type-of-html (161/24508 = 0.7%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 6, fixMaxPrec = 6}
        -- infixr 9 (158/24508 = 0.6%) in tfp (68/24508 = 0.3%), sequence (45/24508 = 0.2%), sandwich (45/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 9, fixMaxPrec = 9}
        -- infix 6 (56/24508 = 0.2%) in type-combinators (56/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 6, fixMaxPrec = 6}
        -- infix 4 (51/24508 = 0.2%) in ghc-typelits-presburger (51/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 4, fixMaxPrec = 4}
        -- infixl 8 (49/24508 = 0.2%) in zippers (49/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 8, fixMaxPrec = 8}
        -- infix 5 (39/24508 = 0.2%) in limp (39/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 5, fixMaxPrec = 5}
        -- infixr 8 (25/24508 = 0.1%) in graphql-api (25/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 8, fixMaxPrec = 8}
        -- infixr 7 (11/24508 = 0.0%) in hreq-core (11/24508 = 0.0%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 7, fixMaxPrec = 7}
      ),
      ( ":<|>",
        -- infixr 3 (2238/2317 = 96.6%) in servant (1330/2317 = 57.4%), prolude (543/2317 = 23.4%), gogol-core (183/2317 = 7.9%), servant-js (147/2317 = 6.3%), ocaml-export (35/2317 = 1.5%)
        FixityInfo {fixDir = Just InfixR, fixMinPrec = 3, fixMaxPrec = 3}
        -- infixl 9 (57/2317 = 2.5%) in magicbane (45/2317 = 1.9%), solga (12/2317 = 0.5%)
        --  (threshold reached) <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 9, fixMaxPrec = 9}
        -- infixl 4 (22/2317 = 0.9%) in eved (22/2317 = 0.9%)
        --  (threshold reached) <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 4, fixMaxPrec = 4}
      ),
```

So basically, for `":>"`, `servant` didn't reach the arbitrary threshold (> 90% of the total "popularity") to be considered *The Only Declaration That Matters ™️*, with its little 5.4% of total popularity. As a result, we get a final `FixityInfo {fixDir = Nothing, fixMinPrec = 0, fixMaxPrec = 9}` for this operator, which in simpler terms means "we don't know anything about this operator".
If we had automatic cabal dependency detection, or manual user whitelisting of some packages (like in https://github.com/tweag/ormolu/pull/832), this issue would become almost nonexistant

data/examples/declaration/value/function/operator-chains-out.hs
tbagrel1: When I was working on #528 , I had inconsistencies which could only be detected in chains with 4 operators, that's why I chose 4 identical operators, or 3-1-3

data/examples/declaration/type-synonyms/multi-line-out.hs
tbagrel1: (Answering to your comment about why ":>" and ":<|>" can't be ordered, but for some reason in the Github interface I can only answer to this (just next) comment)

This is what I get from my hoogle/hackage extraction:
```haskell
      ( ":>",
        -- infixl 5 (7876/24508 = 32.1%) in lens (2096/24508 = 8.6%), containers (1489/24508 = 6.1%), fingertree (790/24508 = 3.2%), diagrams-lib (753/24508 = 3.1%), optics-core (691/24508 = 2.8%), rio (626/24508 = 2.6%), optics-extra (598/24508 = 2.4%), preamble (259/24508 = 1.1%), hgeometry-combinatorial (148/24508 = 0.6%), microformats2-parser (65/24508 = 0.3%), sized (58/24508 = 0.2%), hw-fingertree-strict (46/24508 = 0.2%), hw-fingertree (46/24508 = 0.2%), sequence (45/24508 = 0.2%), sitepipe (35/24508 = 0.1%), knit-haskell (34/24508 = 0.1%), lens-utils (29/24508 = 0.1%), strict-containers (22/24508 = 0.1%), raw-feldspar (19/24508 = 0.1%), planet-mitchell (12/24508 = 0.0%), polysemy-optics (10/24508 = 0.0%), kmonad (5/24508 = 0.0%)
        FixityInfo {fixDir = Just InfixL, fixMinPrec = 5, fixMaxPrec = 5}
        -- infixl 9 (6002/24508 = 24.5%) in futhark (2353/24508 = 9.6%), opaleye (1589/24508 = 6.5%), parsley-core (535/24508 = 2.2%), quickcheck-state-machine (244/24508 = 1.0%), parameterized-utils (231/24508 = 0.9%), jukebox (194/24508 = 0.8%), streams (188/24508 = 0.8%), wumpus-basic (104/24508 = 0.4%), parsley (102/24508 = 0.4%), haskus-utils-data (82/24508 = 0.3%), feldspar-language (69/24508 = 0.3%), presburger (60/24508 = 0.2%), liboleg (57/24508 = 0.2%), type-aligned (52/24508 = 0.2%), hobbits (32/24508 = 0.1%), yices (26/24508 = 0.1%), kansas-lava (22/24508 = 0.1%), linear-maps (15/24508 = 0.1%), serv (14/24508 = 0.1%), data-sketches-core (13/24508 = 0.1%), exposed-containers (7/24508 = 0.0%), spine (5/24508 = 0.0%), VecN (5/24508 = 0.0%), fingertree-tf (3/24508 = 0.0%), language-sh (?/24508 = ?%), joinlist (?/24508 = ?%), darcs-beta (?/24508 = ?%), clean-unions (?/24508 = ?%), ForSyDe (?/24508 = ?%)
          <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 9, fixMaxPrec = 9}
        -- infixr 5 (4971/24508 = 20.3%) in clash-prelude (2290/24508 = 9.3%), massiv (1047/24508 = 4.3%), streaming (680/24508 = 2.8%), generic-lens-core (622/24508 = 2.5%), biohazard (114/24508 = 0.5%), sdp (85/24508 = 0.3%), ribbit (63/24508 = 0.3%), forsyde-shallow (41/24508 = 0.2%), linear-base (29/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 5, fixMaxPrec = 5}
        -- infixr 4 (2238/24508 = 9.1%) in servant (1330/24508 = 5.4%), prolude (543/24508 = 2.2%), gogol-core (183/24508 = 0.7%), servant-js (147/24508 = 0.6%), ocaml-export (35/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 4, fixMaxPrec = 4}
        -- infixr 3 (1418/24508 = 5.8%) in HaXml (621/24508 = 2.5%), pandora (599/24508 = 2.4%), PrimitiveArray (161/24508 = 0.7%), joint (37/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 3, fixMaxPrec = 3}
        -- infix 0 (948/24508 = 3.9%) in extensible (879/24508 = 3.6%), membership (69/24508 = 0.3%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 0, fixMaxPrec = 0}
        -- infixr 1 (342/24508 = 1.4%) in darcs (342/24508 = 1.4%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 1, fixMaxPrec = 1}
        -- infixr 0 (163/24508 = 0.7%) in lorentz (80/24508 = 0.3%), llvm-pretty (56/24508 = 0.2%), indigo (27/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 0, fixMaxPrec = 0}
        -- infixr 6 (161/24508 = 0.7%) in type-of-html (161/24508 = 0.7%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 6, fixMaxPrec = 6}
        -- infixr 9 (158/24508 = 0.6%) in tfp (68/24508 = 0.3%), sequence (45/24508 = 0.2%), sandwich (45/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 9, fixMaxPrec = 9}
        -- infix 6 (56/24508 = 0.2%) in type-combinators (56/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 6, fixMaxPrec = 6}
        -- infix 4 (51/24508 = 0.2%) in ghc-typelits-presburger (51/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 4, fixMaxPrec = 4}
        -- infixl 8 (49/24508 = 0.2%) in zippers (49/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 8, fixMaxPrec = 8}
        -- infix 5 (39/24508 = 0.2%) in limp (39/24508 = 0.2%)
          <> FixityInfo {fixDir = Just InfixN, fixMinPrec = 5, fixMaxPrec = 5}
        -- infixr 8 (25/24508 = 0.1%) in graphql-api (25/24508 = 0.1%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 8, fixMaxPrec = 8}
        -- infixr 7 (11/24508 = 0.0%) in hreq-core (11/24508 = 0.0%)
          <> FixityInfo {fixDir = Just InfixR, fixMinPrec = 7, fixMaxPrec = 7}
      ),
      ( ":<|>",
        -- infixr 3 (2238/2317 = 96.6%) in servant (1330/2317 = 57.4%), prolude (543/2317 = 23.4%), gogol-core (183/2317 = 7.9%), servant-js (147/2317 = 6.3%), ocaml-export (35/2317 = 1.5%)
        FixityInfo {fixDir = Just InfixR, fixMinPrec = 3, fixMaxPrec = 3}
        -- infixl 9 (57/2317 = 2.5%) in magicbane (45/2317 = 1.9%), solga (12/2317 = 0.5%)
        --  (threshold reached) <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 9, fixMaxPrec = 9}
        -- infixl 4 (22/2317 = 0.9%) in eved (22/2317 = 0.9%)
        --  (threshold reached) <> FixityInfo {fixDir = Just InfixL, fixMinPrec = 4, fixMaxPrec = 4}
      ),
```

So basically, for `":>"`, `servant` didn't reach the arbitrary threshold (> 90% of the total "popularity") to be considered *The Only Declaration That Matters ™️*, with its little 5.4% of total popularity. As a result, we get a final `FixityInfo {fixDir = Nothing, fixMinPrec = 0, fixMaxPrec = 9}` for this operator, which in simpler terms means "we don't know anything about this operator".
If we had automatic cabal dependency detection, or manual user whitelisting of some packages (like in https://github.com/tweag/ormolu/pull/832), this issue would become almost nonexistant

data/examples/declaration/value/function/operator-chains-out.hs
tbagrel1: I'm working on this :)

data/examples/declaration/value/function/operator-chains-out.hs
tbagrel1: I don't really understand your suggestion here. Do you mean that we should make the example shorter?

data/examples/declaration/value/function/operator-chains-out.hs
tbagrel1: I gave this treatment to any `InfixR 0` operator at the moment, so that `$`-like operators (`$!`, `$!!` ...) would be formatted in a consistent way. We could easily tweak this behaviour by changing the body of this function (and updating some comments in `p_exprOpTree`):
```haskell
-- (Operators.hs, l. 375)

-- | Indicates if an operator has InfixR 0 fixity. We special-case this class of operators because they often have, like '$', a specific 'separator' use-case, and we sometimes format them differently than other operators.
isHardSplitterOp :: FixityInfo -> Bool
```

data/examples/declaration/value/function/operator-chains-out.hs
mrkkrp: Yes.

mrkkrp:
I think we should incorporate your ideas from the #832 in this PR. In particular:

* Perform generation of the fixity/precedence map in Haskell.
* Store info for different packages but do not make a choice at compile time regarding what operator is more popular.
* Rename `--cabal-default-extensions` to `--cabal` which should, in addition to what it is doing now, be used to extract the list of direct dependencies. Do not try to resolve transitive dependencies. Also add `-p / --package` option that can be repeated many times to provide "manually" a list of packages. These packages (either extracted from Cabal or supplied with `-p`) should be used to select the correct operator info. Otherwise, use a heuristic. An explicit config file can be added in a subsequent PR.

tbagrel1:
I'm currently working on
+ rebasing on the new master, and fixing conflicts
+ splitting the huge operator chain example file into smaller bits
+ incorporating ideas from #832 in this PR

src/Ormolu/Printer/Meat/Type.hs
mrkkrp: Since `OpTree` is used for both types and values, I think it makes sense to keep it in a separate file, and not in `Type.hs` or `Value.hs`.

extract-hoogle-hackage-info/src/Main.hs
```diff
@@ -0,0 +1,305 @@
...
+  liftIO $ putFmtLn "Processing file {}..." (Only filePath)
+  state@State{sProcessedFiles} <- get
+  fileContent <- liftIO . readFileUtf8Latin1 $ filePath
+  packageName <- liftIO . fromRightWithIOError $ getPackageName hoogleDatabasePath filePath
+  let state' = foldl' (onSymbolDecl packageName) state (subgroups1 <$> symbolDecls fileContent)
+      state'' = foldl' (onFixityDecl packageName) state' (subgroups3 <$> fixityDecls fileContent)
+      subgroups1 (_ :| g1 : []) = g1
+      subgroups1 _ = error "invalid number of capturing groups"
+      subgroups3 (_ :| g1 : g2 : g3 :[]) = (g1, g2, g3)
+      subgroups3 _ = error "invalid number of capturing groups"
```
amesgen, line 204: You can elide the `error` calls and make this more typesafe by using [type-indexed capture groups](https://hackage.haskell.org/package/pcre2-2.0.3/docs/Text-Regex-Pcre2.html#g:7). See here for the minimal diff: https://gist.github.com/amesgen/90fe533fea1c14cda4f86f70d657259d

default.nix
```diff
@@ -39,6 +39,7 @@ let
         })
         (gitTH "ormolu" "")
         (gitTH "ormolu-live" "../")
+        (gitTH "extract-hoogle-hackage-info" "../")
```
tbagrel1, line 42: I'm not really sure of what I did in `default.nix`, I mostly copied/adapted what was already there (but with little understanding of what it means). So a review is really needed on this file

amesgen: `gitTH` is only necessary if the project uses sth like [`gitrev`](https://hackage.haskell.org/package/gitrev), where the `.git` subfolder has to be available at compile time for Template Haskell, cf. https://github.com/tweag/ormolu/issues/342. For `extract-hoogle-hackage-info`, that is not the case, so this particular line can be removed.

The rest of the changes to `default.nix` seem fine :+1:. One thing that could be changed is that `extract-hoogle-hackage-info` currently does not compile on GHC 9.2.1 as some dependencies are not yet up-to-date. But that is not a problem, you can move it to the 8.10-only part of the file: https://gist.github.com/amesgen/78b2e76d103a3caebda1f820f8d0e642

tbagrel1:
When trying a full nix-build, I encounter an error for the weeder  target:
```
[nix-shell:~/tweag/ormolu]$ nix-build -A weeder
trace: No index state specified for haskell-project, using the latest index state that we know about (2021-11-16T00:00:00Z)!
trace: No index state specified for weeder, using the latest index state that we know about (2021-11-16T00:00:00Z)!
these derivations will be built:
  /nix/store/zh9nsakyl5sn18f7bky9ymjb10c4l8rv-ormolu-weeder.drv
building '/nix/store/zh9nsakyl5sn18f7bky9ymjb10c4l8rv-ormolu-weeder.drv'...
src/Ormolu/Fixity.hs:129: defaultFixityMap
src/Ormolu/Fixity.hs:65: hoogleHackageInfoFile
builder for '/nix/store/zh9nsakyl5sn18f7bky9ymjb10c4l8rv-ormolu-weeder.drv' failed with exit code 1
```

+ defaultFixityMap is used by ormolu-live , but not by ormolu directly (so this error is expected, but I don't know how to silence it)
+ hoogleHackageInfoFile  contains the embedded JSON fixity file as a ByteString, and is actually used by the rest of the ormolu code, so I don't really know why weeder spots it :thinking_face:

amesgen:
- We currently build Ormolu Live with GHCJS only (which does not support HIE files used by weeder), so we could either add `defaultFixityMap` as an exception (by adding `"^Ormolu.Fixity.defaultFixityMap\$"` to `roots` in `weeder.dhall`), or inline the definition of `defaultFixityMap` in Ormolu Live (and then remove it).
 -  This appears to be a bug in weeder (I have a minimal reproducible example), I will report/try to fix this upstream. Until then, it can also be added to `roots` in `weeder.dhall`.

tbagrel1:
TODO list:

+ [ ] Check and complete comments/documentation
+ [ ] Rename `operator-chain-xxx.hs` files with a more descriptive name
+ [ ] Add tests to check that fixity config is correctly parsed from command line
+ [ ] Add tests to check that fixity config gives the correct fixity map
+ [ ] Improve packages popularity rating by taking reverse dependencies into account

amesgen:
>  Decide what to do with (duplicated sometimes) `Could not find a .cabal file for ...`

Do you mean the ones in `expected-failures`? It is expected that these warnings appear from time to time, as many projects have `.hs` files which do not appear in a `.cabal` file (e.g. `Setup.hs`, a file mostly present for legacy cabal reasons, or other example/test files).

tbagrel1:
> > Decide what to do with (duplicated sometimes) `Could not find a .cabal file for ...`
> 
> Do you mean the ones in `expected-failures`? It is expected that these warnings appear from time to time, as many projects have `.hs` files which do not appear in a `.cabal` file (e.g. `Setup.hs`, a file mostly present for legacy cabal reasons, or other example/test files).

I enabled by default the autodetection of cabal dependencies to build the fixity map. If a cabal file is not found, a warning is emitted (in the same way `detectCabalExtensions` do, but this option is disabled by default). As a result, a warning can be displayed in contexts where there wasn't any before. Moreover, when both detectCabalExtensions and detectCabalDependencies options are enabled, the same warning can be emitted twice (because atm, each function is responsible for finding the cabal file associated to a given source file).

How do you think we can improve this?

amesgen:
One possible thing would be to try (if necessary) to find the .cabal files for all input files in the beginning, so warnings would only appear once (and all in one place), and the .cabal files would also only be parsed once (and again and again not for every file formatted, even though I don't think that this is very relevant for performance). While I don't think it is hard to do, it might require quite some refactoring, so I would suggest to leave it as is for now?

tbagrel1:
Also, the CI still fails, but I don't know why -_-

amesgen:
This is due to weird cabal behavior with in multi package projects (it tries to resolve dependencies for packages it does not build). You can fix this by adding these two lines to the extract-hoogle .cabal file:

https://github.com/tweag/ormolu/blob/3781d3748874126d688e4896de0f8d71bfdb3330/ormolu-live/ormolu-live.cabal#L34-L35

tbagrel1:
What would I do without you :sweat_smile: 
Thank you, I'll try that!

app/Main.hs
```diff
@@ -9,24 +12,35 @@ module Main (main) where
 import Control.Exception (throwIO)
 import Control.Monad
 import Data.Bool (bool)
+import qualified Data.HashMap.Strict as HashMap
 import Data.List (intercalate, sort)
 import Data.Maybe (fromMaybe, mapMaybe)
+import Data.Text (Text, pack, replace, unpack)
```
mrkkrp, line 18: Only import `Text`, for others use `T.pack`, etc. This is the style that is used in the rest of the codebase.

app/Main.hs
```diff
@@ -172,11 +202,12 @@ data Mode
     Check
   deriving (Eq, Show)
 
--- | Configuration for how to account for default-extension
--- from .cabal files
-data CabalDefaultExtensionsOpts = CabalDefaultExtensionsOpts
+-- | Configuration for how to use .cabal files
```
mrkkrp, line 205: "to use" is a bit too vague. Let's say "what kind of information to extract from .cabal files".

amesgen:
Ah now there are other dependency resolution failures, in particular as you excluded `aeson-2.*` which adds GHC 9.2 support. You can test it locally using `nix-build --keep-going --no-out-link --argstr ormoluCompiler ghc921` (see the `.buildkite/pipeline.yml` file).

app/Main.hs
```diff
@@ -204,7 +235,7 @@ optsParserInfo =
               $gitBranch,
               $gitHash
             ],
-          "using ghc-lib-parser " ++ VERSION_ghc_lib_parser
+          "using ghc-lib-parser " ++ "9.2.1.20211101"
```
mrkkrp, line 238: This should be reverted.

app/Main.hs
```diff
@@ -230,21 +261,31 @@ optsParser =
...
     <*> (many . strArgument . mconcat)
       [ metavar "FILE",
         help "Haskell source files to format or stdin (the default)"
       ]
 
-cabalDefaultExtensionsParser :: Parser CabalDefaultExtensionsOpts
-cabalDefaultExtensionsParser =
-  CabalDefaultExtensionsOpts
+cabalOptsParser :: Parser CabalOpts
+cabalOptsParser =
```
mrkkrp, line 272: IMO there is no good reason to take default extensions and language, but not the list of packages. I can imagine two distinct cases:

* There is no Cabal file or it cannot be found. Then the user is going to provide the extensions and other info on the command line.
* There is a Cabal file that we can find. Then take all the useful info from that file.

Let's have a single flag `--cabal` that does all at once like a originally suggested.

tbagrel1:
> Ah now there are other dependency resolution failures, in particular as you excluded `aeson-2.*` which adds GHC 9.2 support. You can test it locally using `nix-build --keep-going --no-out-link --argstr ormoluCompiler ghc921` (see the `.buildkite/pipeline.yml` file).

I had no idea which particular package versions I needed to pin, and I did something not really consistent... sometimes I pinned both major & minor version numbers, sometimes only the major version number...

Is there a convention / rule of hand for how should be pinned dependency versions?

app/Main.hs
```diff
@@ -296,6 +337,35 @@ configParser =
                 help "End line of the region to format (inclusive)"
               ]
         )
+    <*> ( FixityConfig
```
mrkkrp, line 340: Let's do this PR without these options and then consider what to do with them in a separate PR. I feel like already there is going to be a lot of changes and discussions.

format.sh
```diff
@@ -10,6 +10,7 @@ export LANG="C.UTF-8"
 cabal format
 
 ormolu -e -m inplace $(find app -type f -name "*.hs")
-ormolu -e -m inplace $(find src -type f \( -name "*.hs" -o -name "*.hs-boot" \))
+ormolu -e -m inplace $(find src -type f -name "*.hs-boot" -o -name "*.hs")
 ormolu -e -m inplace $(find tests -type f -name "*.hs")
 ormolu -e -m inplace $(find ormolu-live/src -type f -name "*.hs")
+ormolu -e -m inplace $(find extract-hoogle-hackage-info/src -type f -name "*.hs")
```
mrkkrp, line 16: Please configure your editor not to eat up the trailing line.

amesgen:
There are a lot of different opinions on this, this is one sensible way: https://github.com/wasp-lang/haskell-handbook/blob/master/dependencies-version-bounds.md (recently discussed on https://redd.it/qncnvd). For now, I would just go with `>= 1 && < 3` (which is very broad, but aeson does not change much).

app/Main.hs
tbagrel1: I don't know what this line is, and I don't remember making this change... Maybe I missclicked in my IDE? :thinking:

app/Main.hs
tbagrel1: I added a separate flag, because dependency detection should be enabled by default IMO, but extension detection isn't enabled by default at the moment (If I understand correctly the Main.hs file)

app/Main.hs
tbagrel1: Should we still detect cabal dependencies in this PR then?

app/Main.hs
mrkkrp: The current `--cabal-default-extensions` flag tells Ormolu to try to look for a Cabal file to take default extensions from there. We want to keep giving the user this choice (whether or not they want Ormolu to look at Cabal files), but if a Cabal file is found we want to also take the list of direct dependencies from there. So, yes, it would take packages from the Cabal file by default but only if user wants Ormolu to look for Cabal files at all.

app/Main.hs
mrkkrp: Yes, like I described in the comment above. When `--cabal` is passed, get the list of direct dependencies from there, otherwise use a heuristic. These options are useful and we will have them in some form later, but I want to us to first settle on and merge a simpler version.

app/Main.hs
tbagrel1: I'm just afraid that people who do not use the `--cabal-default-extensions` flag at the moment (which will be renamed `--cabal`) will miss out the accuracy improvement for operator chains formatting

mrkkrp: I understand, but we should not try to look for Cabal files unconditionally at this point. What we can do is discuss whether we want `--cabal` to be enabled by default. We could also add `--no-cabal` to opt out from this. But let's tackle questions one by one and not try to do too much in a single PR.

app/Main.hs
tbagrel1: I saved the current state of this branch to https://github.com/tweag/ormolu/tree/tbagrel1/fixity-with-config, so we could easily integrate the more granular fixity config later

src/Ormolu/Printer/Meat/Type.hs
tbagrel1: I moved them to `OpTree.hs`. I suppose we can resolve this conversation thread?

data/examples/declaration/value/function/operator-chains-out.hs
tbagrel1: We agreed to extend this special case to InfixR 0 operators on monday, so I'll mark this as resolved

extract-hoogle-hackage-info/src/Main.hs
tbagrel1: Done!

app/Main.hs
tbagrel1: Fixed!

app/Main.hs
tbagrel1: Fixed!

app/Main.hs
tbagrel1: Fixed!

src/Ormolu.hs
```diff
@@ -83,12 +85,14 @@ ormolu cfgWithIndices path str = do
...
       parseModule'
         cfg
+        fixityMap
         OrmoluOutputParsingFailed
         path
         (T.unpack txt)
     unless (cfgUnsafe cfg) $ do
       when (length result0 /= length result1) $
-        liftIO $ throwIO (OrmoluASTDiffers path [])
+        liftIO $
```
mrkkrp, line 94: Do you know why this is happening? `($)` is right-associative, so `liftIO $ throwIO (OrmoluASTDiffers path [])` should be grouped together and they are originally on the same line:

```haskell
      when (length result0 /= length result1) $
        liftIO $ throwIO (OrmoluASTDiffers path [])
```

however a new line is introduced after the second `$`.

extract-hoogle-hackage-info/extract-hoogle-hackage-info.cabal
```diff
@@ -0,0 +1,29 @@
...
+        aeson >= 1.0 && <3.0,
+        pcre2 >= 2.0 && < 3.0,
+        directory >= 1.0 && < 2.0,
+        unix >= 2.0 && < 3.0,
+        tagsoup >= 0.14 && < 0.15,
+        text-format >= 0.3 && < 0.4,
+        ghc-lib-parser >=9.2 && <9.3,
+        bytestring >= 0.10 && < 0.11
+    ghc-options: -O2 -Wall -rtsopts
+    if !impl(ghc == 8.10.*)
```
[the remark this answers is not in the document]
tbagrel1: The `extract-hoogle-hackage-info` script doesn't build yet with 9.x, because of some dependencies not yet available. So I added `extract-hoogle-hackage-info` to the 8.10.x-only part of the `default.nix` build instructions, but it seems that this line is necessary because I added the `extract-hoogle-hackage-info` subproject to the `cabal.project` file. @amesgen [wrote a comment about it](https://github.com/tweag/ormolu/pull/830#issuecomment-979296670)

data/examples/declaration/value/function/infix/operator-chain-a-out.hs
```diff
@@ -0,0 +1,7 @@
+-- Right chain, normal case, 1 operator type
```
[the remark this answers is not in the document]
tbagrel1: Still in my TODO list :)

src/Ormolu.hs
tbagrel1: The whole `$` chain is multiline, so each element of the chain will be on its own line. That's why a line break is introduced after the 2nd `$`

app/Main.hs
```diff
@@ -263,8 +264,12 @@ configParser =
...
         help "GHC options to enable (e.g. language extensions)"
       ]
-    -- We cannot access cabal dependencies here
-    <*> pure []
+    <*> (many . strOption . mconcat)
+      [ long "package",
+        short 'p',
+        metavar "PACKAGE",
+        help "Explicitly specified dependency (for operator fixity/precedence only)"
+      ]
```
tbagrel1, line 272: I think that if we allow manual (CLI) configuration for the fixity map, it makes sense to allow users to specify the fixity of a single operator (not only packages), for example the fixity of an operator created by the user themself in their project. If we allow manual configuration, we also need a lot of the stuff I discarded from https://github.com/tweag/ormolu/tree/tbagrel1/fixity-with-config.

If we only add a `--package` option like you did in this commit, this option won't allow people using also `--cabal` to resolve a conflicting definition for an operator coming from cabal dependencies, because the manually specified packages will be at the same priority level as cabal dependencies (so that can only create more conflicts).

I know this commit is labelled as "WIP", but I just wanted to share some of the motivations/reasons about the amount of configuration I allow in https://github.com/tweag/ormolu/tree/tbagrel1/fixity-with-config

data/examples/declaration/value/function/infix/operator-chain-a-out.hs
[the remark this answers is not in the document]
tbagrel1: I just changed the names with my last commit, let me know if it looks good for you.

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+    `--no-cabal` switch.
+  * Dependencies can be selected manually by using the `-p / --package`
+    option (can be repeated many times).
+  * The default heuristic algorithm will still try to guess the right
+    fixities and precedence.
+  This resolves the following issues: [Issue
+    826](https://github.com/tweag/ormolu/issues/826), [Issue
+    785](https://github.com/tweag/ormolu/issues/785), [Issue
+    690](https://github.com/tweag/ormolu/issues/690).
+
```
tbagrel1, line 21: I just saw that it will also close #825

src/Ormolu.hs
mrkkrp: Let's keep it as it is right now.

mrkkrp:
@tbagrel1 Can you please ensure for all new modules (fixity, op tree, operator) that:

* Line length does not exceed 80, or better 76
* Every top-level definition in the modules has a Haddock explaining what that is.

tbagrel1:
> @tbagrel1 Can you please ensure for all new modules (fixity, op tree, operator) that:
> 
> * Line length does not exceed 80, or better 76
> * Every top-level definition in the modules has a Haddock explaining what that is.

Ok!
Is there any tool/recommandations to wrap the comments on column 80/76?

mrkkrp:
> Is there any tool/recommandations to wrap the comments on column 80/76?

I do it manually using "common sense" (tm).

tbagrel1:
> @tbagrel1 Can you please ensure for all new modules (fixity, op tree, operator) that:
> 
> * Line length does not exceed 80, or better 76
> * Every top-level definition in the modules has a Haddock explaining what that is.

@mrkkrp Done with https://github.com/tweag/ormolu/pull/830/commits/646f801fe5c9e38510e7b0b72ab770ee7c77354f !

src/Ormolu/Fixity.hs
```diff
@@ -0,0 +1,273 @@
...
+  } =
+    fromJust . decodeStrict $ hoogleHackageInfoFile
+
+-- | Gives fixity information (direction and precedence level) about an
+-- infix operator, but takes the uncertainty that can arise from conflicting
+-- definitions into account.
+data FixityInfo = FixityInfo
+  { -- | Fixity direction (InfixL, InfixR, or InfixN (not associative)),
+    -- if it is known
+    fixDir :: Maybe FixityDirection,
```
mrkkrp, line 93: In what case `fixDir` is going to be `Nothing`?

src/Ormolu/Fixity.hs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+-- | Gives the ability to merge two (maybe conflicting) definitions for an
+-- operator, keeping the higher level of compatible information from both.
+instance Semigroup FixityInfo where
+  FixityInfo {fixDir = dir1, fixMinPrec = min1, fixMaxPrec = max1}
+    <> FixityInfo {fixDir = dir2, fixMinPrec = min2, fixMaxPrec = max2} =
+      FixityInfo
+        { fixDir = dir',
+          fixMinPrec = min min1 min2,
```
mrkkrp, line 125: What was your logic for this?

mrkkrp:
What kind of tests could we have for the fixity module? I think we should have some. We could check that the `"extract-hoogle-hackage-info/hoogle-hackage-info.json"` contains reasonable data for

* operators from the `base` package that I should have `FixityInfo` from `base`
* operators from some other packages possibly conflicting ones to check the merging algorithm?

tbagrel1:
I think we could create `buildFixityMap' :: PackageToOps -> PackageToPop -> [String] -> Float -> FixityMap`, and then set `buildFixityMap = buildFixityMap' packageToOps packageToPop`. We would then be able to test separately the merging/override algorithm (+ base priorization), and the correctness of the data inside `extract-hoogle-hackage-info.json`

tbagrel1:
I just saved the `HackageFixityMap.hs` file generated by my old python script at https://gist.github.com/tbagrel1/fa785b38433393dcefc2465b7338c8be
This file is quite useful to find operators with conflicting fixities, in order to write unit tests for `Fixity.hs`. 
I will also include this link in the `HackageInfoSpec.hs` test module  that I'm currently writing.

tbagrel1:
At this point, I'm starting to wonder, shouldn't we just implement our own `FixityDirection` enum instead of reusing GHC's one?
With this change, we could get rid of orphan instances, and derive easily `FromJSON`, `ToJSON`, and `Show`.

tbagrel1:
Now that `FixityDirection` is no longer the GHC defined one, why do we implement `FromJSON` and `ToJSON` manually? Is there a drawback to just deriving them thanks to `Generic`?

tbagrel1:
WIP list of librairies shipped with GHC: https://docs.google.com/spreadsheets/d/1VO_FYsCVOgdfA5b3cJnPKq4IiCBKcW1D0ZoNyt-1_uA/edit?usp=sharing

tbagrel1:
I just finished most of my assigned tasks (see [d6fb306a41cb75ec18995eec443b233baaaff72c](https://github.com/tweag/ormolu/pull/830/commits/d6fb306a41cb75ec18995eec443b233baaaff72c) ); but I have a few questions:
+ do you think we need more comments/documentation in `extract-hackage-info/src/Mains.hs`?
+ do you think we need tests for the n-ary tree reassociation?
+ should we get rid of the manual implementation of `FromJSON` and `ToJson` for `FixityDirection` and `FixityInfo` (and derive it instead)?
+ should we get rid of `Data.HashMap.Strict` and just use `Data.Map` instead?

Also, I think that I will have time to implement the "only use `NaryOpTree`, get rid of `OpTree`" feature, if it is ok for you.

amesgen:
> should we get rid of `Data.HashMap.Strict` and just use `Data.Map` instead?

I did this in [`issue-826-fixities-th`](https://github.com/tweag/ormolu/tree/amesgen/issue-826-fixities-th), but it is not isolated there.

src/Ormolu/Fixity.hs
```diff
@@ -160,6 +161,35 @@ packageToPopularity :: HashMap String Int
 HackageInfo packageToOps packageToPopularity =
   fromJust . A.decodeStrict $ hackageInfoFile
 
+-- | List of packages shipped with GHC, for which the download count from
+-- Hackage does not reflect their high popularity.
+-- See https://github.com/tweag/ormolu/pull/830#issuecomment-986609572.
+-- "base" is not is this list, because it is already whitelisted
+-- by buildFixityMap'.
+bootPackages :: [String]
```
mrkkrp, line 169: Can this be an internal detail of the extraction script?

mrkkrp:
@tbagrel1 Thanks, I'll review your work later this week.

> do you think we need more comments/documentation in extract-hackage-info/src/Mains.hs?

Not at the moment. I'll need to find time to look in more detail at that file anyway, which make cause it to change :)

> do you think we need tests for the n-ary tree reassociation?

Let's wait till we simplify the situation with trees.

> should we get rid of the manual implementation of FromJSON and ToJson for FixityDirection and FixityInfo (and derive it instead)?

I like to control JSON instances manually because I'm not always happy with auto-generated instances. Let's keep the manual definitions for now.

> should we get rid of Data.HashMap.Strict and just use Data.Map instead?

Let's not do this in this branch, instead let's take it from @amesgen 's branch when it is ready.

> I will have time to implement the "only use NaryOpTree, get rid of OpTree" feature

Feel free to attempt this if you have time. Like we discussed last week, the idea would be to get rid of `OpTree` by i) directly converting AST to `NaryOpTree` and ii) printing `NaryOpTree` directly after it has been reassociated properly. We also want to get rid of `SubTreeInfo` if possible and instead do span calculations, etc. directly.

src/Ormolu/Fixity.hs
tbagrel1: I don't think that it is a good idea to put the boot packages override in the extraction script:

- we moved the "base" whitelisting + colon definition from the extraction script to `Fixity.hs`, so to stay consistent, I think we should keep any other kind of override in `Fixity.hs`
- In the extraction script, we don't know the threshold that's going to be used to build the fixity map, and thus, we can't know how many "virtual" downloads we should add to the download count of boot packages so that they become the reference for the operators they define

amesgen:
Here are the benchmarks for fixity parsing at compile time: https://gist.github.com/amesgen/64e86711d21fb7e2c4fa5b597438a349

tbagrel1:
Note to myself: maybe we can get rid of the `OpInfo` wrapper at printing time

tbagrel1:
Another round of questions:
+ Should we rename `OpBranches` to `OpBranch` now that `OpTree` is actually a N-ary tree? (the idea behind that would be to reduce diff size, maybe)?
+ In `p_cmdOpTree` and `p_tyOpTree`, there is no "hard splitter operator" (`$`-like) special case, so I don't need every operator to be wrapped in `OpInfo op`. Should I keep `OpInfo` to stay consistent with `p_exprOpTree`, or should I get rid of it? Or maybe there are operators in cmd and types contexts that would benefit from being placed in a trailing position?
+ With a really minor change, we could display operator priorities in multiline contexts (like I wanted initially when I started this PR, but with no known blockers this time). Here is a link to a before/after preview of what it would give: https://gist.github.com/tbagrel1/d86e067fbdfea317713c561743a6dbe1. I think that it might be interesting to reconsider this idea now that the implementation is way more robust.

mrkkrp:
> Should we rename OpBranches to OpBranch now that OpTree is actually a N-ary tree?

I think we can rename `NaryOpTree` to `OpTree`. Not sure about `OpBranches` because the way it is, it reflects the possibility of having multiple branches there, which is nice.

> Should I keep OpInfo to stay consistent with p_exprOpTree, or should I get rid of it?

If this is going to require an extra function that strips `OpInfo`, I'd say let's keep it even if it is not used.

> display operator priorities in multiline contexts

I'd say let's not do this because this is not how Haskellers write code. For example here

```haskell
type PermuteRef =
         "a"
           :> ( "b" :> "c" :> End
                  :<|> "c" :> "b" :> End
              )
           :<|> "b"
             :> ( "a" :> "c" :> End
                    :<|> "c" :> "a" :> End
                )
```

No one would actually indent `:>` one level more just to show that it has higher priority. I think it's going to look a bit strange and perhaps even provoke bug reports.

data/examples/declaration/value/function/operators-0-out.hs
```diff
@@ -1,5 +1,6 @@
 a =
-  b & c .~ d
+  b
+    & c .~ d
     & e
-      %~ f
```
tbagrel1, line 4: Here, it seems that the previous output accounted for operator precedence

data/examples/declaration/value/function/operators-6-out.hs
```diff
@@ -4,6 +4,6 @@ type PermuteRef =
            :<|> "c" :> "b" :> End
        )
     :<|> "b"
-      :> ( "a" :> "c" :> End
```
tbagrel1, line 7: Same thing here, which contrasts with your slack message @mrkkrp

src/Ormolu/Printer/Internal.hs
```diff
@@ -266,7 +273,7 @@ spit stype text = do
   R $ do
     i <- asks rcIndent
     c <- gets scColumn
-    closestEnclosing <- listToMaybe <$> asks rcEnclosingSpans
+    closestEnclosing <- asks (listToMaybe . rcEnclosingSpans)
```
tbagrel1, line 276: Do you know who made this change, and how is it related to operator chains? It might be a hint suggested by HLS/VSCode that I would have applied by mistake

src/Ormolu/Printer/Meat/Declaration/Warning.hs
```diff
@@ -34,10 +34,11 @@ p_topLevelWarning :: [LocatedN RdrName] -> WarningTxt -> R ()
 p_topLevelWarning fnames wtxt = do
   let (pragmaText, lits) = warningText wtxt
   switchLayout (fmap getLocA fnames ++ fmap getLoc lits) $
-    pragma pragmaText . inci $ do
-      sep commaDel p_rdrName fnames
-      breakpoint
-      p_lits lits
+    pragma pragmaText . inci $
+      do
```
tbagrel1, line 38: We need to check whether this is the expected behaviour or not (`$` chains with 2+ `$` and a hanging last operand)
We should also add test cases for this.

src/Ormolu/Printer/Meat/Pragma.hs
```diff
@@ -16,7 +15,7 @@ import qualified Data.Text as T
 import GHC.Types.SrcLoc
 import Ormolu.Parser.CommentStream
 import Ormolu.Parser.Pragma (Pragma (..))
-import Ormolu.Printer.Combinators
+import Ormolu.Printer.Combinators hiding (Placement (..))
```
tbagrel1, line 18: Why should we hide placement here?

tbagrel1 (commented):
I did another pass on every changed file, and I added a few questions along the way

src/Ormolu/Printer/Meat/Declaration/Warning.hs
mrkkrp: `do` should stay in the hanging position.

data/examples/declaration/value/function/operators-0-out.hs
mrkkrp: Yes, we should probably go back to having more indentation indicative of the actual precedence.

src/Ormolu/Printer/Meat/Pragma.hs
amesgen: Otherwise we get ambiguity errors:
```
src/Ormolu/Printer/Meat/Pragma.hs:85:22: error:
    Ambiguous occurrence ‘Normal’
    It could refer to
       either ‘Ormolu.Printer.Combinators.Normal’,
              imported from ‘Ormolu.Printer.Combinators’ at src/Ormolu/Printer/Meat/Pragma.hs:18:1-33
              (and originally defined
                 at src/Ormolu/Printer/Combinators.hs:321:5-10)
           or ‘Ormolu.Printer.Meat.Pragma.Normal’,
              defined at src/Ormolu/Printer/Meat/Pragma.hs:41:5
   |
85 |           Nothing -> Normal
   |                      ^^^^^^

src/Ormolu/Printer/Meat/Pragma.hs:89:20: error:
    Ambiguous occurrence ‘Normal’
    It could refer to
       either ‘Ormolu.Printer.Combinators.Normal’,
              imported from ‘Ormolu.Printer.Combinators’ at src/Ormolu/Printer/Meat/Pragma.hs:18:1-33
              (and originally defined
                 at src/Ormolu/Printer/Combinators.hs:321:5-10)
           or ‘Ormolu.Printer.Meat.Pragma.Normal’,
              defined at src/Ormolu/Printer/Meat/Pragma.hs:41:5
   |
89 |               else Normal
   |                    ^^^^^^

src/Ormolu/Printer/Meat/Pragma.hs:90:12: error:
    Ambiguous occurrence ‘Normal’
    It could refer to
       either ‘Ormolu.Printer.Combinators.Normal’,
              imported from ‘Ormolu.Printer.Combinators’ at src/Ormolu/Printer/Meat/Pragma.hs:18:1-33
              (and originally defined
                 at src/Ormolu/Printer/Combinators.hs:321:5-10)
           or ‘Ormolu.Printer.Meat.Pragma.Normal’,
              defined at src/Ormolu/Printer/Meat/Pragma.hs:41:5
   |
90 |       _ -> Normal
   |            ^^^^^^
```

data/examples/declaration/value/function/operators-0-out.hs
tbagrel1: Just in case, if at some point we decide to not display precedences anymore, see https://gist.github.com/tbagrel1/a5b9db62db23295113048ac410f30045

src/Ormolu/Printer/Meat/Declaration/Warning.hs
tbagrel1: Fixed with my last commit

data/examples/declaration/value/function/operators-0-out.hs
tbagrel1: Fixed with the last commit

extract-hackage-info/src/Main.hs
```diff
@@ -0,0 +1,406 @@
...
+    conflictLines ((packageName, opName), fixities) =
+      format
+        "(in {}) {}"
+        (packageName, opName)
+        : indentLines (showT <$> fixities)
+
+limitMap :: Ord k => Int -> Map k v -> Map k v
+limitMap n = Map.fromList . take n . Map.toList
+
+getCounts :: Map String FixityMap -> (Int, Int, Int)
```
mrkkrp, line 305: This function should have a Haddock. What counts?

default.nix
```diff
@@ -21,21 +21,28 @@ let
...
             pkgs.lib.mkOption { apply = pkgs.lib.remove "Cabal"; };
+        }
+        ({ pkgs, ... }: mkIf pkgs.stdenv.hostPlatform.isGhcjs {
+          packages.ormolu = {
+            flags.fixity-th = false;
+            writeHieFiles = pkgs.lib.mkForce false;
+          };
+          packages.ormolu-live.ghcOptions =
+            # Remove stack size limit for Ormolu Live
+            [ "+RTS" "-K0" "-RTS" ] ++
```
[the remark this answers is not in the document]
amesgen: Ah yes, that is probably a better place. Is this a small enough change that I can just push it to this PR?

weeder.dhall
```diff
@@ -1,6 +1,8 @@
 { roots =
   [ "^Main.main\$"
   , "^Paths_"
+  , "^Ormolu.Fixity.defaultFixityMap\$"
```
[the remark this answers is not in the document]
amesgen: - The weeder bug https://github.com/ocharles/weeder/issues/92 is actually no longer being triggered since #841 as I inlined `hackageInfoFile` (so `"^Ormolu.Fixity.hackageInfoFile\$"` from `weeder.dhall`).
 -  `"^Ormolu.Fixity.defaultFixityMap\$"` is used in Ormolu Live, but we don't build it with GHC (only with GHCJS) on CI, so weeder can't know that it is used there. This could be easily fixed by inlining `defaultFixityMap`, but Ormolu Live's behavior w.r.t. the new fixity stuff will probably be changed either way (e.g. by adding configurability in the interface), so we can do this then.   
 - `type-class-roots = True` is necessary due to a [known limitation](https://github.com/ocharles/weeder#calling-weeder) of weeder which is triggered by `showFixityDirection` in `ToJSON FixityInfo`.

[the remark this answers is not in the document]
tbagrel1: I think that we should remove `showFixityDirection` now that `FixityDirection` implements `Show`?

default.nix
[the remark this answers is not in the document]
mrkkrp: Yes, please go ahead.

extract-hackage-info/src/Main.hs
tbagrel1: My latest commit should fix everything you listed in your latest review. Let me know if I still need to ajust something.

src/Ormolu/Fixity.hs
```diff
@@ -0,0 +1,247 @@
...
+          `difference` (dependencies ++ higherPriorityPackages)
+      buildPackageFixityMap packageName =
+        ( packageName,
+          fromMaybe Map.empty $
+            Map.lookup packageName operatorMap
+        )
+      difference l1 l2 =
+        Set.toList $
+          Set.fromList l1 `Set.difference` Set.fromList l2
+      mergeAll = mergeFixityMaps Map.empty 10.0
```
mrkkrp, line 163: Why 10.0 ?

src/Ormolu/Fixity.hs
```diff
@@ -0,0 +1,247 @@
...
+
+-- | Merge a list of individual fixity maps, coming from different packages.
+-- Package popularities and the given threshold are used to choose between
+-- the "keep best only" (>= threshold) and "merge all" (< threshold)
+-- strategies when conflicting definitions are encountered for an operator.
+mergeFixityMaps ::
+  -- | Map from package to popularity
```
mrkkrp, line 170: Here and in the definition of `HackageInfo` it should be explained what `Int` means.

tests/Ormolu/HackageInfoSpec.hs
```diff
@@ -0,0 +1,472 @@
...
+import qualified Data.Map.Strict as Map
+import Ormolu.Fixity
+import Test.Hspec
+
+(|->) :: a -> b -> (a, b)
+a |-> b = (a, b)
+
+infixr 0 |->
+
+checkFixityMap' ::
```
mrkkrp, line 12: I think we need to add docs for `checkFixityMap` and `checkFixityMap'`.

tests/Ormolu/HackageInfoSpec.hs
```diff
@@ -0,0 +1,472 @@
+module Ormolu.HackageInfoSpec where
+
+import qualified Data.Map.Strict as Map
+import Ormolu.Fixity
+import Test.Hspec
+
+(|->) :: a -> b -> (a, b)
```
mrkkrp, line 7: Can we get rid of this operator? Using tuples for this kind of purpose is idiomatic in Haskell, so a custom operator rather adds fiction than helps readability.

src/Ormolu/Printer/Operators.hs
```diff
@@ -1,182 +1,365 @@
...
 import GHC.Types.Name.Reader
-import GHC.Types.SourceText
 import GHC.Types.SrcLoc
+import Ormolu.Fixity
 import Ormolu.Utils
 
--- | Intermediate representation of operator trees. It has two type
--- parameters: @ty@ is the type of sub-expressions, while @op@ is the type
--- of operators.
+-- | Intermediate representation of operator trees, where a branching is not
```
mrkkrp, line 24: Can you please go through the module and try to make sure that the haddocks are up to date with the current code?

mrkkrp:
I played with profiling today a little bit and it looks like if we delay construction of `remainingFixityMap` until we really don't find an operator in the "more common" maps we can spend in `buildFixityMap` less than 1 % of time in most cases (e.g. when all operators come from dependencies or are built-in). We could define the resulting fixity map as a type that wraps two maps perhaps and then make a two-stage lookup function. I think this might solve the performance issue.

tbagrel1:
> I played with profiling today a little bit and it looks like if we delay construction of `remainingFixityMap` until we really don't find an operator in the "more common" maps we can spend in `buildFixityMap` less than 1 % of time in most cases (e.g. when all operators come from dependencies or are built-in). We could define the resulting fixity map as a type that wraps two maps perhaps and then make a two-stage lookup function. I think this might solve the performance issue.

Who is/will be working on this? Should I take care of it?

src/Ormolu/Fixity.hs
tbagrel1: We need an arbitrary value strictly greater than 1.0, so that no package can "win" the popularity game.

mrkkrp:
> Who is/will be working on this? Should I take care of it?

I think we should stick to our plan that we outlined on Monday:

* @tbagrel1 works on the tests for reassociation logic and addresses my feedback as I do the final pass of my review.
* @amesgen Tries to improve the performance.

mrkkrp:
I have finished reviewing the PR. There are two things that we still need to address before we merge it:

- [ ] Tests for operator chain reassociation.
- [ ] Improve performance.

Other than that, it looks really good. Great job @tbagrel1 !

tests/Ormolu/OpTreeSpec.hs
```diff
@@ -0,0 +1,169 @@
...
+spec = do
+  it "flattens a tree correctly" $ do
+    let inputTree =
+          tree
+            ( tree
+                ( tree
+                    (n "a")
+                    "+"
+                    (n "b") ::
+                    T
```
tbagrel1, line 62: Does anyone have an idea about how I could get rid of the type annotations here?

mrkkrp: Try enabling `FunctionalDependencies` and changing your type class definition to:

```haskell
class TreeBuildList ty op t | t -> ty op where
  buildTree :: [OpTree ty op] -> [op] -> t
```

mrkkrp: Also you'll probably need to change the signature of `tree`

```haskell
tree :: (TreeBuildList ty op t) => OpTree ty op -> op -> OpTree ty op -> t
```

tbagrel1:
I think that we also need to fix this before merging: https://github.com/tweag/ormolu/issues/845#issuecomment-994705438
I'll do it right now

mrkkrp:
@tbagrel1 Let's do it in a separate PR. I'm going to merge this once it is green.

PR #837, opened 2021-11-25
Escape triggers of invalid haddocks

### LINKED ISSUE #816: Leading space added when a multiline invalid haddock comment is located at top-level
When a haddock comment is:

- located at top level AND
- invalid in this context AND
- multiline

Then the  `mkComment` function (in `CommentStream.hs`) get the wrong `n = minimum (startIndent : fmap getIndent xs)` number of leading spaces to drop in front of subsequent comment lines (it always results in `n == 0` because `startIndent == 0` in this case).

As a result, the space following `--` in `-- this is the second line of my comment` gets included in the actual comment string, to which `commentPrefix == "-- "` is then prepended. Thus, we get `--  this is the second line of my comment`.

This can't be reproduced in regular modules (only in signature files), because in regular modules, a top-level haddock comment can't be invalid AFAIK. We also can't reproduce it with non top-level comments, because in that case `startIndent > 0` so `n` gets the correct value.

Example:

```haskell
signature Implementation ( Prim
                         , name
                         , primName
                         , randomBlocks
                         ) where

import Raaz.Www
import Raaz.Core

data Prim
instance Primitive Prim

name :: String
-- | The name of the primitive used as the seed stretcher
-- second line of comment
primName :: String
randomBlocks :: AlignedBlockPtr BufferAlignment Prim
             -> BlockCount Prim
             -> Internals
             -> IO ()
```

tbagrel1:
Closes #816


src/Ormolu/Parser/CommentStream.hs
```diff
@@ -163,7 +166,22 @@ mkComment ls (L l s) = (ls', comment)
             "{-" -> (False, ls'')
             _ -> (True, ls'')
     dropTrailing = L.dropWhileEnd isSpace
-    startIndent = srcSpanStartCol l - 1
+    startIndent
+      -- multiline comments have an "absolute" start at column 0
```
mrkkrp, line 170: I don't think this comment is correct. What happens if that if a multiline comment sits at column 6 we actually drop all 6 spaces (unless there are other lines with fewer number of spaces in the indent) and this becomes the output of `mkComment`. I think we subtract 1 to account for the fact that `SrcSpan` counts columns from 1. Spaces are indeed restored but later by the printing engine, based on the actual indent the comment has when printing.

data/examples/other/invalid-haddock.hs
```diff
@@ -4,14 +4,18 @@ test = undefined
       -- ** foo
       Int ->
       -- | misplaced
+      --  blablabla
```
mrkkrp, line 7: This example is rather long now. Let's split it in two.

src/Ormolu/Parser/CommentStream.hs
tbagrel1: Yeah you're right, I misunderstood what was happening there. I will correct it :)
I suppose that the next comment (relative to the single-line comment case) is correct though?

mrkkrp: Yes, the second one is correct.

mrkkrp:
To solve the issue with double pipes like this:

```haskell
foo =
  -- || b
  -- || c
  b
```

Move the function `escapeHaddockTriggers` outside of `mkComment` like so:

```haskell
-- | Escape characters that can turn a line into a Haddock.
escapeHaddockTriggers :: String -> String
escapeHaddockTriggers string
  | h : _ <- string, h `elem` "|^" = '\\' : string
  | otherwise = string
```

And then use it in `unAnnotationComment` by applying it before `dashPrefix`.

Don't forget to add a new test case for this one. Create a new file, don't add it to existing files.

tbagrel1:
I implemented what you suggested in [this comment](https://github.com/tweag/ormolu/pull/837#issuecomment-979299877), but I'm not really satisfied by the inconsistency issue between the first and the other lines of an invalid haddock comment: the first line get its first haddock trigger removed, which I would avoid. The examples show what I would like to achieve.

My questions are now:

- should we also include `$` and `*` to the escape list?
- Is there a way to "get back" the haddock trigger which is removed in case of an invalid haddock comment?

mrkkrp:
> should we also include $ and * to the escape list?

Yes.

> Is there a way to "get back" the haddock trigger which is removed in case of an invalid haddock comment?

What are you trying to achieve?

> I'm not really satisfied by the inconsistency issue between the first and the other lines of an invalid haddock comment

Frankly, at this point I'm okay with just escaping everything :)

tbagrel1:
> > Is there a way to "get back" the haddock trigger which is removed in case of an invalid haddock comment?

> What are you trying to achieve?

>> I'm not really satisfied by the inconsistency issue between the first and the other lines of an invalid haddock comment

> Frankly, at this point I'm okay with just escaping everything :)

I would just escape everything in an invalid haddock comment. It seems that the haddock trigger of the first line is removed by GHC-lib-parser itself, so we can't "escape it" without bringing it back first.

Basically, I would like this:
```haskell
foo =
  -- | invalid
  b
```
to be turned into this:
```haskell
foo =
  -- \| invalid
  b
```

The only info we have about this missing trigger is the variant of `GHC.EpaCommentTok`, which is one of

| Variant | desc. |
|------|------|
| EpaDocCommentNext String | something beginning '-- \|' |
| EpaDocCommentPrev String | something beginning '-- ^' |
| EpaDocCommentNamed String | something beginning '--$' | 
| EpaDocSection Int String | a section heading |
| EpaDocOptions String | doc options (prune, ignore-exports, etc) |
| EpaLineComment String | comment starting by "--" |
| EpaBlockComment String | comment in {- -} |
| EpaEofComment | empty comment, capturing location of EOF |

according to https://downloads.haskell.org/~ghc/9.2.1/docs/html/libraries/ghc-9.2.1/GHC-Parser-Annotation.html#t:EpaCommentTok

So theorically, the first three variants (and maybe a fourth one,  `EpaDocSection`) should allow us to guess the trigger that had been removed.

But this documentation seems incorrect, because in our test cases, I got two examples, one with `GHC.EpaDocCommentNext`, and one with `GHC.EpaDocSection`, where the original file misses the resp. tokens `|` and `@`...
The commit that I will push in a minute illustrates these weird examples.

I'm desesperatly looking for the grammar definition of haddock comments, but I don't know where to start looking...

tbagrel1:
EDIT:
The case with `-- ** foo` being turned to `-- @` comes from a typing error on my side (I wrote `haddock '@' s` instead of `haddock '*' s`), that I will correct in the next commit

We still have a weird case in `other/invalid-haddock-weird.hs`, where `foo = foo -- # ${` is interpreted as `GHC.EpaDocCommentNext` and thus is turned into `foo = foo -- \|# ${`

tbagrel1:
It looks good to me now. It is in a reviewable state :)

mrkkrp (approved):

PR #841, opened 2021-12-08
Parse fixity info at compile time by default

amesgen:
This PR adds a flag, `fixity-th` (enabled by default), which will enable parsing of fixity info at compile time via TH. This increases build times (see [here](https://gist.github.com/amesgen/75c1da6cc51d204972e1b39e73797041]), but decreases run times (see [here](https://gist.github.com/amesgen/64e86711d21fb7e2c4fa5b597438a349)).

For Ormolu Live (GHCJS does not seem to like huge modules) we disable `fixity-th`.

Also, it makes sense to disable `fixity-th` locally when working on fixity stuff by adding
```cabal
constraints: ormolu -fixity-th
```
to `cabal.project.local` to get e.g. faster feedback times in GHCJS and a haskell-language-server which does not crash.


ormolu.cabal
```diff
@@ -98,10 +103,17 @@ library
...
         directory ^>=1.3,
         Cabal >=3.6 && <3.7,
-        unordered-containers >=0.2 && <0.3,
-        hashable >=1.0 && <2.0,
         aeson >=1.0 && <3.0,
-        file-embed >=0.0.15 && <0.0.16
+        template-haskell,
+        th-lift-instances >=0.1 && <0.2
+
+    mixins:           ghc-lib-parser hiding (Language.Haskell.TH.Syntax)
```
mrkkrp, line 110: Could we do an import with explicit package name instead?

mrkkrp (approved):

ormolu.cabal
amesgen: Sure; I would have a very slight preference for using mixins as the disambiguation only happens in one place, this is how it would look like:

```diff
diff --git a/ormolu.cabal b/ormolu.cabal
index fec2f3f..bf21686 100644
--- a/ormolu.cabal
+++ b/ormolu.cabal
@@ -107,8 +107,6 @@ library
         template-haskell,
         th-lift-instances >=0.1 && <0.2
 
-    mixins:           ghc-lib-parser hiding (Language.Haskell.TH.Syntax)
-
     if flag(fixity-th)
         cpp-options: -DFIXITY_TH
 
diff --git a/src/Ormolu/Fixity.hs b/src/Ormolu/Fixity.hs
index 948583e..9636b34 100644
--- a/src/Ormolu/Fixity.hs
+++ b/src/Ormolu/Fixity.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE PackageImports #-}
 {-# LANGUAGE TemplateHaskell #-}
 
 -- | Definitons for fixity analysis.
@@ -32,7 +33,7 @@ import qualified Data.Set as Set
 import Ormolu.Fixity.Types
 #if FIXITY_TH
 import Control.Monad.IO.Class (MonadIO (liftIO))
-import qualified Language.Haskell.TH.Syntax as TH
+import qualified "template-haskell" Language.Haskell.TH.Syntax as TH
 #else
 import Data.FileEmbed (embedFile)
 import Data.Maybe (fromJust)
diff --git a/src/Ormolu/Fixity/Types.hs b/src/Ormolu/Fixity/Types.hs
index a8e5bfb..16ba712 100644
--- a/src/Ormolu/Fixity/Types.hs
+++ b/src/Ormolu/Fixity/Types.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DeriveLift #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PackageImports #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE StandaloneDeriving #-}
 
@@ -20,7 +21,7 @@ import Data.Map.Strict (Map)
 import Data.Text (Text)
 import qualified Data.Text as T
 import Instances.TH.Lift ()
-import qualified Language.Haskell.TH.Syntax as TH
+import qualified "template-haskell" Language.Haskell.TH.Syntax as TH
 
 -- | Fixity direction.
 data FixityDirection
```

mrkkrp: Okay, let's go with mixins :+1:

PR #848, opened 2021-12-16
Cache operator fixity maps

amesgen:
We cache the fixity map (depending on the package set). This speeds up scenarios when ormolu formats many files in the same package context, especially when they use non-standard operators (where #847 doesn't help).

## Timings

Test suite: from 2.0998s to 1.0540s

`nix-build -A hackageTests`: from 2m14,382s to 1m41,690s

### Formatting the same file multiple times

Test.hs:
```haskell
1++++++1
```

`ormolu` is 0.4.0.0, `./ormolu-pre` is before this PR and `./ormolu` is with this PR.

Formatting Test.hs once: No change as expected
| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `ormolu -e Test.hs ` | 5.8 ± 0.6 | 5.1 | 8.5 | 1.00 |
| `./ormolu-pre Test.hs ` | 72.5 ± 1.6 | 70.3 | 81.5 | 12.56 ± 1.31 |
| `./ormolu Test.hs ` | 72.6 ± 1.0 | 71.1 | 75.2 | 12.57 ± 1.29 |

Formatting Test.hs 10x and 100x: big speedups

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `ormolu -e Test.hs ... ` | 21.1 ± 1.2 | 19.6 | 24.0 | 1.00 |
| `./ormolu-pre Test.hs ... ` | 411.7 ± 8.1 | 403.9 | 430.5 | 19.52 ± 1.15 |
| `./ormolu Test.hs ... ` | 80.9 ± 0.8 | 79.3 | 83.5 | 3.84 ± 0.22 |

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `ormolu -e Test.hs ... ` | 159.3 ± 3.5 | 155.6 | 169.3 | 1.00 |
| `./ormolu-pre Test.hs ... ` | 3995.6 ± 96.0 | 3822.2 | 4158.4 | 25.08 ± 0.81 |
| `./ormolu Test.hs ... ` | 210.6 ± 2.6 | 206.3 | 214.9 | 1.32 ± 0.03 |

PR #849, opened 2021-12-16
Ormolu Live: Propagate fixity map cache at startup

amesgen:
Previously, as soon as there was any operator in the input, it would become horribly slow. With #848, it is only slow for the first time an operator is entered, and fast afterwards. With this PR, we now do this at startup/loading time (which is already so long that an extra 300ms is probably worth the snapiness afterwards).


amesgen:
Maybe this can be resolved more nicely by evaluating `defaultFixityMap` at startup, I will try this.

amesgen:
> Maybe this can be resolved more nicely by evaluating `defaultFixityMap` at startup, I will try this.

I am sure that this would work, but it would require evaluating `defaultFixityMap` deeply, but deriving the necessary `NFData` instances seems like more work than what we are doing in this PR.

PR #850, opened 2021-12-16
Move defaultFixityMap to Ormolu live

mrkkrp:


amesgen:
I will fix the boring import issues and remove `defaultFixityMap` from `weeder.dhall`.

PR #851, opened 2021-12-16
Format bidirectional pattern synonyms nicer

### LINKED ISSUE #843: Unnecesary newline with bidirectional pattern synonyms
With 0.4.0.0, the snippet
```haskell
{-# LANGUAGE PatternSynonyms #-}

pattern Seq l l' <- Seq' l l'
  where
    Seq l l' = case l' of
      Empty -> l
      _ -> Seq' l l'
```
is formatted to 
```haskell
{-# LANGUAGE PatternSynonyms #-}

pattern Seq l l' <-
  Seq' l l'
  where
    Seq l l' = case l' of
      Empty -> l
      _ -> Seq' l l'
```
where the newline after the `<-` seems a bit unnecessary (but it probably should be present if the pattern were mult-line).

mrkkrp:
Close #843.

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -858,21 +858,25 @@ p_hsExpr' s = \case
 
 p_patSynBind :: PatSynBind GhcPs GhcPs -> R ()
 p_patSynBind PSB {..} = do
-  let rhs = do
+  let rhs conSpans = do
         space
+        let pattern_def_spans = [getLocA psb_id, getLocA psb_def] ++ conSpans
         case psb_dir of
-          Unidirectional -> do
-            txt "<-"
-            breakpoint
-            located psb_def p_pat
-          ImplicitBidirectional -> do
-            equals
-            breakpoint
-            located psb_def p_pat
+          Unidirectional ->
+            switchLayout pattern_def_spans $ do
+              txt "<-"
+              breakpoint
+              located psb_def p_pat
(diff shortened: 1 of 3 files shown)


src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -858,21 +858,25 @@ p_hsExpr' s = \case
 
 p_patSynBind :: PatSynBind GhcPs GhcPs -> R ()
 p_patSynBind PSB {..} = do
-  let rhs = do
+  let rhs conSpans = do
         space
+        let pattern_def_spans = getLocA psb_def : conSpans
```
amesgen, line 863:
```suggestion
        let pattern_def_spans = [getLocA psb_id, getLocA psb_def] <> conSpans
```

The snippet
```haskell
{-# LANGUAGE PatternSynonyms #-}

pattern NumbersNumbersEverywhereeeeeeeeeeeeeeeee =
  B 1234 1243 1234 1234 1234 1234 1234 1234 1234 1234
```
a fixed point, but now it is formatted to
```haskell
{-# LANGUAGE PatternSynonyms #-}

pattern NumbersNumbersEverywhereeeeeeeeeeeeeeeee = B 1234 1243 1234 1234 1234 1234 1234 1234 1234 1234
```
If the LHS has arguments, the issue disappears. I added a quick fix which I think fixes the issue (it also requires one update in the test suite).

amesgen (approved):
LGTM apart from the one minor issue mentioned below.

PR #852, opened 2021-12-17
Handle multi-operator fixity declarations in the extraction script

tbagrel1:


mrkkrp:
Have you made sure that it indeed correctly parses multi-operator fixity declarations?

tbagrel1:
>  Have you made sure that it indeed correctly parses multi-operator fixity declarations?

I tried it with `infixr 6 ====>>,====<<, ===>>> ,<<<===`, and I got them in the output map with the expected fixities.

mrkkrp (approved):

tbagrel1:
FYI, there is no fixity declaration with multiple operators on the same line in the current hoogle database. So this change is only intended to future-proof the extraction script.

mrkkrp:
What about e.g. https://hackage.haskell.org/package/base-4.16.0.0/docs/src/GHC.List.html ?

It has:

```haskell
infix  4 `elem`, `notElem`
```

Shouldn't it be accounted for?

tbagrel1:
[base.txt](https://github.com/tweag/ormolu/files/7735907/base.txt)

I think that Hoogle might pre-process hackage source files, because I can't find
```haskell
infix 4 `elem`, `notElem`
```
in this file.

PR #853, opened 2021-12-17
Rename Ormolu.Fixity.Types to Ormolu.Fixity.Internal

mrkkrp:
I think `Internal` makes it more obvious that the module is not part of the
public API (e.g. the data constructor of `LazyFixityMap` should normally
stay hidden).

### DIFF
--- a/ormolu.cabal
+++ b/ormolu.cabal
@@ -74,7 +74,7 @@ library
         Ormolu.Printer.Meat.Type
         Ormolu.Printer.Operators
         Ormolu.Fixity
-        Ormolu.Fixity.Types
+        Ormolu.Fixity.Internal
         Ormolu.Printer.SpanStream
         Ormolu.Processing.Common
         Ormolu.Processing.Cpp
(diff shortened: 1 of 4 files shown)


mrkkrp:
They are re-exported from `Ormolu.Fixity`, which is part of the public API.

tbagrel1:
In that case, my argument no longer holds. Let's do it so!

amesgen (approved):
Good idea, the entire reason why this module exists in the first place is due to Template Haskell, so it is really an implementation detail that it even exists.

We *could* even go one step further and move `Ormolu.Fixity.Internal` to `other-modules`, but this would require at least exporting the constructor of `LazyFixityMap` for `OpTreeSpec.hs`, which should not break the abstraction.

mrkkrp:
In general, I think `other-modules` is almost always a bad idea. I used to use it myself to hide things that are not supposed to be used by users, but the experience proves that there might be some cases when people need those hidden internal things when they want to do something the author of the code hadn't thought about. The impossibility to opt in for using them really makes people angry for no good reason. I currently advocate for keeping everything importable/public, but calling modules that are not supposed to be used "internal".

Issue #710, opened 2021-06-09
Placement of comma for lists, records, tuples, function type signatures...


lehins:
**TLDR** It is unreasonable to start putting commas at the end of lines. 

There was a ticket already on this topic that was about  making this configurable in #497 

In this ticket, however, I am not proposing making ormolu configurable. This issue is about changing the default, because current formatting choice is vastly worse than what the defacto Haskell style is.

Of course, my personal preference plays a role in me opening this ticket, but I will put it aside and will show facts instead.

I have looked at arguments in the [blogpost](https://www.tweag.io/blog/2019-10-11-ormolu-first-release/) in favor of comma placement at the end rather that at the beginning:

## Multiline pattern matching

> While this popular formatting choice works in expressions, it’s a parse error if used in a pattern, because everything in a multiline pattern should be more indented than the opening parenthesis.

Multiline pattern matching in case statements is rather uncommon, especially when compared with popular formatting choice. I even had to google it in order to see what it actually means. 

For anyone reading this here is an example. These two will parse:
```haskell
bar :: Num a => [a] -> a
bar [ x
    , y
    , z
    ] = x + y + z
baz :: Num a => [a] -> a
baz xs =
  case xs of
    [ x,
      y,
      z ] -> x + y + z
```
But this will not:

```haskell
baz :: Num a => [a] -> a
baz xs =
  case xs of
    [ x
    , y
    , z ] -> x + y + z
```
For the curious here is a [ghc issue](https://gitlab.haskell.org/ghc/ghc/-/issues/17028) that tries to improve the error message about it. IMHO it would be better to fix ghc to support such syntax, rather than use it as an argument to change formatting style.

## Trailing comma is now supported by ghc

```haskell
module Foo 
  ( foo,
    bar,
  ) where
```
vs
```haskell
module Foo 
  ( foo
  , bar
  ) where
```
> Did you notice that we also add trailing commas where possible, for example, in export lists? Our ability to do this comes from a relatively new feature in Haskell—it helps with Ormolu’s goal of achieving minimal diffs too. If we try to remember where leading commas come from, Johan Tibell’s style guide comes to mind. The author said later:

> "[…] I designed [Haskell style guide] to work with the lack of support for a trailing comma. If we supported a trailing comma my style guide would probably be different."

> There you have it: GHC supports a trailing comma now, so it’s not unreasonable to start putting commas at the end of lines. 

That argument works for exports **only** and we do lack trailing comma support, so Johan Tibell's design is still valid.

[...]

PR #857, opened 2022-01-08
Add reference to the trailing comma problem and fourmolu

itsfarseen:
Ref: https://github.com/tweag/ormolu/issues/710#issuecomment-864436964

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
     * [Exit codes](#exit-codes)
 * [Limitations](#limitations)
 * [Running on Hackage](#running-on-hackage)
+* [Forks and modifications](#forks-and-modifications)
 * [Contributing](#contributing)
 * [License](#license)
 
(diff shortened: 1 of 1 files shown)


mrkkrp:
Thanks for your interest! I think there is no problem with referencing fourmolu in the readme because it indeed may provide functionality that suits certain users better than what's available in Ormolu. However, I don't think it is fair to include a deliberate stylistic choice we made in Ormolu as a "limitation". If you look at the existing items in the list of limitations, you'll notice that they are all hard and genuine problems. The way commas are placed can be changed very easily and so it's not exactly the same kind of thing.

itsfarseen:
> However, I don't think it is fair to include a deliberate stylistic choice we made in Ormolu as a "limitation". If you look at the existing items in the list of limitations, you'll notice that they are all hard and genuine problems.

Yeah, totally agree. I put it there because I couldn't find a better place to put.  
I have changed it to be a separate heading. But I feel that it does look kind a weird though.

itsfarseen:
I have moved it under the Notes section. It felt weird to put `Comma placement` as a top level heading.

mrkkrp:
I took the liberty to adjust the PR. I think the main point is that there are actively maintained forks of Ormolu, namley Fourmolu that might be of interest to the reader of the readme. I made a section called "forks" to talk about it.

PR #860, opened 2022-02-07
Maintenance: update versions of some dependencies

mrkkrp:
Also update haskell.nix.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,9 +3,9 @@ steps:
     command: |
       nix-build --keep-going --no-out-link --argstr ormoluCompiler ghc8107
     timeout: 100
-  - label: Build and test with GHC 9.0.1
+  - label: Build and test with GHC 9.0.2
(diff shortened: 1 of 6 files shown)


amesgen (commented):
- Updating GHC 9.0.1 to 9.0.2?
 - Allowing text-2.0? (compiles fine locally)
 - The CI failure on GHC 9.0 and 9.2 is due to https://github.com/input-output-hk/haskell.nix/issues/1177 (but I am not really sure why this only pops up now). I adapted the fix suggested in the linked PR here, and it works fine: https://github.com/tweag/ormolu/commit/cfc7848af17c99c1fbe9648435189e09980c53f7
  - Ormolu Live build failure: As miso does not specify bounds on its dependencies, we somehow end up building against an ancient servant version. I fixed this here: https://github.com/tweag/ormolu/commit/0d95fbd07416bbcbb9cf15743940e615887892d8

amesgen (approved):

PR #867, opened 2022-04-03
Fix stack.yaml

brandonchinn178:
The current `stack.yaml` doesn't build because of nightly's prior issues with happy: https://github.com/commercialhaskell/stackage/issues/6294


mrkkrp:
Thanks!

PR #868, opened 2022-04-04
Add back sitcc to ExplicitTuple.p_arg

brandonchinn178:
`sitcc` was removed as part of the `ghc-lib-parser-9.2` upgrade (#779), but it looks like that was erroneous; the only change that needed to happen in `p_larg` was the removal of `located`. This causes problems in `fourmolu` with leading commas (see https://github.com/fourmolu/fourmolu/issues/148#issuecomment-1086988846), and this fixes it. I'm not quite sure how to come up with a failing test case for this in ormolu, but I'm fairly certain that the removal of `sitcc` is incorrect, and may cause other unexpected behaviors.

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -705,9 +705,10 @@ p_hsExpr' s = \case
         isMissing = \case
           Missing _ -> True
           _ -> False
-        p_arg = \case
-          Present _ x -> located x p_hsExpr
-          Missing _ -> pure ()
+        p_arg =
+          sitcc . \case
+            Present _ x -> located x p_hsExpr
+            Missing _ -> pure ()
         parens' =
           case boxity of
             Boxed -> parens


brandonchinn178:
cc @amesgen, just because you did the ghc-lib-parser-9.2 upgrade

mrkkrp:
Can you come up with an example that would demonstrate the difference between the current behavior and the behavior with `sitcc` added?

brandonchinn178:
so I found some differences with do-blocks in tuple sections:
```hs
-- without sitcc
x =
  [ (do
      x <- asdf
      return x,)
  ]
x =
  [ (do
      x <- asdf
      return x,do
      x <- asdf2
      return x,)
  ]

-- with sitcc
x =
  [ (do
       x <- asdf
       return x,)
  ]
x =
  [ (do
       x <- asdf
       return x,do
                  x <- asdf2
                  return x,)
  ]
```
without `sitcc`, do blocks in a tuple section don't know that they're in a tuple; they're indented only once (in the list context) as opposed to twice (in the list + tuple context). In the second case, I'm actually surprised that the first isn't a parse error, since I thought `x <- asdf2` needs to be to the right of the `do` keyword, but GHC accepts it. Regardless, I think the case with sitcc is better than the case without (although both should be fixed to add the appropriate space/newline before the `do`).

This only comes up with tuple sections because `commaDel` adds a breakpoint before, which seems to handle the missing sitcc logic. So it might not be that big of a deal (who does tuple sections with multiline blocks?), but I'd argue that:
1. adding `sitcc` does the least-unexpected thing in the uncommon case
2. adding `sitcc` does nothing in the common case
3. adding `sitcc` fixes an issue in the downstream fourmolu project

I know (3) probably doesn't matter to you, but in combination with (1), i don't see any reason _not_ to do it, and it has a nice side benefit of helping downstream. plus, i suspect that if ormolu changes something around styling tuples here, this `sitcc` will be needed, so this is also future-proofing

amesgen (approved):
I can confirm that removing `sitcc` in #779 was just an oversight, so I agree that there is no reason not to add it back.

mrkkrp:
Thanks!

PR #873, opened 2022-04-10
Remove discrepancies between unboxed types and terms

### LINKED ISSUE #872: Discrepency between unboxed types and terms
**Describe the bug**
The spacing for unboxed types is different than the spacing for unboxed terms.

**To Reproduce**
The code is formatted like this:
```haskell
{-# LANGUAGE MagicHash #-}
{-# LANGUAGE UnboxedSums #-}
{-# LANGUAGE UnboxedTuples #-}

type Maybe# x = (# (# #)| (# x #) #)

type Tup = (# #)

tup :: ()
tup = ()
  where
    x = (##)

type Thingy# = (# Int| Bool #)

test = ()
  where
    x :: Maybe# Int
    x = (# (##) | #)

thingy :: ()
thingy = ()
  where
    x :: Thingy#
    x = (# 1 | #)

    y :: Thingy#
    y = (# | True #)
```

**Expected behavior**
The spacing for the term formatting should be the same as type formatting.

amesgen:
Closes #872


oberblastmeister:
Shouldn't unboxed unit tuples have a space too?

amesgen:
Ah, that didn't catch my eye. Any preferences whether we should use `(##)` or `(# #)` (the former seems to be used in the [GHC user guide](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/primitives.html#unboxed-tuples))?

oberblastmeister:
I would prefer `(# #)`, but as long as they match the types it should be okay.

mrkkrp (approved):

PR #878, opened 2022-05-03
Make errors about AST differences print diffs

### LINKED ISSUE #877: Output some diff on "AST differs" errors
The current error shown when Ormolu formats things in a way that changes the AST is very opaque. It's perhaps intentionally so, as the error always indicates a bug on Ormolu's side, but it does cause a block in the immediate work waiting on a response after opening an issue. There is also no mention of using --unsafe to bypass it temporarily.

The error message already shows the lines affected. It would be great to:
1. Keep the "Please report this as a bug" message
2. Show a diff between the original version and formatting with --unsafe on the particular lines / show some representation of the AST that differs
    * Alternatively, mention --unsafe as a temporary solution

**Additional context**

mrkkrp:
Close #877.

Print the diff between the original input and the erroneous result that
Ormolu produces. Also mention --unsafe in the error message.

### DIFF
--- a/src/Ormolu/Diff/ParseResult.hs
+++ b/src/Ormolu/Diff/ParseResult.hs
@@ -1,9 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 -- needed on GHC 9.0 due to simplified subsumption
 {-# LANGUAGE ImpredicativeTypes #-}
-{-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns #-}
@@ -30,7 +27,7 @@ data ParseResultDiff
   = -- | Two parse results are the same
     Same
   | -- | Two parse results differ
-    Different [SrcSpan]
+    Different [RealSrcSpan]
   deriving (Show)
 
 instance Semigroup ParseResultDiff where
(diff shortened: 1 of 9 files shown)


src/Ormolu/Diff/ParseResult.hs
```diff
@@ -30,7 +27,7 @@ data ParseResultDiff
   = -- | Two parse results are the same
     Same
   | -- | Two parse results differ
-    Different [SrcSpan]
+    Different [RealSrcSpan]
```
mrkkrp, line 30: This seems to be a bit more precise typing because `isGoodSpan` selects `RealSrcSpan`s anyway. Afterwards we can check things like line numbers using total functions.

src/Ormolu/Diff/Text.hs
```diff
@@ -70,45 +85,59 @@ diffText a b path =
...
+selectSpans :: [RealSrcSpan] -> TextDiff -> TextDiff
+selectSpans ss textDiff = textDiff {textDiffSelectedLines = xs}
+  where
+    xs = foldl' addOneSpan (textDiffSelectedLines textDiff) ss
+    addOneSpan linesSoFar s =
+      let start = srcSpanStartLine s
+          end = srcSpanEndLine s
+       in IntSet.union
+            linesSoFar
+            (IntSet.fromAscList (take (1 + end - start) [start ..]))
```
amesgen, line 99:
```suggestion
            (IntSet.fromAscList [start .. end])
```

src/Ormolu/Diff/Text.hs
```diff
@@ -284,3 +310,12 @@ mapDiff f = fmap $ \case
 
 third :: (Int, Int, Text) -> Text
 third (_, _, x) = x
+
+isSelectedLine :: IntSet -> Hunk -> Bool
+isSelectedLine selected Hunk {..} =
+  -- If the set of selected lines is empty, everything is selected.
+  IntSet.null selected
+    || not (IntSet.null (IntSet.intersection selected hunkOriginalLines))
```
amesgen, line 318:
```suggestion
    || not (IntSet.disjoint selected hunkOriginalLines)
```

src/Ormolu/Diff/Text.hs
```diff
@@ -8,26 +8,35 @@
...
 
 ----------------------------------------------------------------------------
 -- Types
 
 -- | Result of diffing two 'Text's.
-data TextDiff = TextDiff FilePath DiffList
+data TextDiff = TextDiff
+  { textDiffPath :: FilePath,
+    textDiffDiffList :: DiffList,
+    textDiffSelectedLines :: IntSet
```
amesgen, line 34: Maybe add a comment also here that mentions that an empty `textDiffSelectedLines` means "all lines are selected"?

amesgen (approved):
LGTM apart from a few minor comments below!

I guess that there is no easy way already in the codebase to print a `TextDiff` without a terminal context, so we can't print it in Ormolu Live. But I don't think this is blocking or even necessary at all.

src/Ormolu/Diff/Text.hs
mrkkrp: Done!

PR #884, opened 2022-05-10
Fix extract hackage info

### LINKED ISSUE #883: Popularity estimation of Hackage packages does not work anymore
It looks like Hackage maintainers have changed [this page][hackage-browse] and it is now populated with JavaScript. This makes the instructions in `extract-hackage-info/README.md` obsolete. We need a way to make this work again.

Perhaps a script to automate the whole process of extraction would make sense, too.

[hackage-browse]: https://hackage.haskell.org/packages/browse

mrkkrp:
Close #883.


extract-hackage-info.sh
```diff
@@ -0,0 +1,29 @@
...
+OUTPUT="$WDIR/hackage-info.json"
+
+trap cleanup 0 1 2 3 15
+
+cleanup()
+{
+    rm -rf "$WDIR"; exit
+}
+
+nix-build -A extractHackageInfo --no-out-link
```
amesgen, line 17: I think this line should be unnecessary, as `extractHackageInfo` is built as part of `EXTRACTION_APP`?

amesgen (approved):
Looks good to me, there is probably no better way without a better Hackage API.

tbagrel1 (approved):
Looks good to me too!

extract-hackage-info.sh
amesgen: > I wasn't sure that only the final build output goes to stdout.

Yes, the logs go to stderr; we already use this e.g. here:

https://github.com/tweag/ormolu/blob/276dfdd9a22dfad914578984b891d2a6d33300e6/ormolu-live/deploy.sh#L3

mrkkrp: Fixed :+1:

PR #885, opened 2022-05-12
GHC 9.4

amesgen:
Soft blocked by https://github.com/input-output-hk/haskell.nix/issues/1595 for GHC 9.4 on CI.

---

Overview:

 - Switch to `ghc-lib-parser-9.4.*`. This includes:
    - Support for the new syntactic features in GHC 9.4 ([changelog](https://downloads.haskell.org/ghc/9.4.1/docs/users_guide/9.4.1-notes.html#language)):
       - `\cases` via `LambdaCase`
       - `OPAQUE` pragmas
       - Unboxed sum type constructors like `(# | #)`.
    - ghc-lib-parser-9.4.1 builds with GHC 8.10, but it is not explicitly supported. Hence, I switched from 8.10 to 9.2 as the default version. This required some changes in `extract-hackage-info`.
    - GHC AST changes:
       - More unreachable cases got TTG-ed away :tada: 
       - For exact-printing purposes, many constructors now contain [`HsToken`s](https://downloads.haskell.org/ghc/9.4.1/docs/libraries/ghc-9.4.1/Language-Haskell-Syntax-Extension.html#t:HsToken), which we ignore. 
 - Some changes to the nix setup: some previous haskell.nix workarounds are now unnecessary. Also, we now only set the GHC flags added in #793 when actually building the executable, as otherwise the entire `ghc` library is rebuilt due to changed flags (which takes quite some time).
 - Ormolu Live:
    - I needed to patch ghc-lib-parser slightly for it to compile on GHCJS 8.10.7.
    - closure-compiler optimizations no longer preserves the semantics (all inputs suddenly fail to parse). Disabling them works, but increases the JS sizes by a lot. I haven't looked into this in detail, especially as we hopefully will be able to switch away from GHCJS to either the native WASM or JS backend in GHC 9.6.


ysangkok:
@amesgen Hi again :) Can't you just remove `th-lift-instances`? It doesn't seem to be needed on modern GHC versions according to https://github.com/phadej/file-embed-lzma/pull/11 . I tried building Ormolu without it but ghc-lib-parser is taking forever to build (seems unrelated)

amesgen:
Hi @ysangkok! Yeah, we need the `Lift` instance for containers. We could also require `containers >= 0.6.6` (which has those instances natively) and hence remove `th-lift-instances`, but GHCJS (which we use here) does not play nicely with reinstalling boot packages, so I didn't go with it ATM.

expected-failures/brittany.txt
```diff
@@ -0,0 +1,611 @@
...
++   { foo1,
++     foo2 ::
++       loooooooooooooooooooooooooooooooong ->
++       loooooooooooooooooooooooooooooooong,
++     bar :: a,
++     bazz :: b
++   }
++
+
+  AST of input and AST of formatted code differ.
```
amesgen, line 509: This is due to us (knowningly, see #725) not supporting `DatatypeContexts`, but it would be trivial to add.

All other new `expected-failures` seem to be known comment-related idempotency bugs.

src/Ormolu/Printer/Meat/Common.hs
```diff
@@ -130,6 +132,18 @@ p_infixDefHelper isInfix indentArgs name args =
         breakpoint
         inciIf indentArgs $ sitcc (sep breakpoint sitcc args)
 
+-- | Print a Haddock.
+p_hsDoc ::
```
mrkkrp, line 136: It looks like `p_hsDocString` is not really useful on its own after this change. Could we remove it from the export list and possibly merge the two definitions so that we only have a longer definition of `p_hsDoc` directly?

default.nix
```diff
@@ -251,31 +249,32 @@ in {
       find . -name '*.hs' -exec cp --parents {} $out \;
     '';
   };
-  binaries = {
-    Linux = hsPkgs.projectCross.musl64.hsPkgs.ormolu.components.exes.ormolu;
+  binaries = let hsPkgsOpt = hsPkgs.appendModule {
+    modules = [{
```
mrkkrp, line 253: What is the meaning of this manipulation? Does `appendModule` merge configurations or something?

mrkkrp (commented):
Great job, as usual! I think that the main prerequisite for merging it is still the availability of GHC 9.4.1 in `haskell.nix`. Do you know if there have been any updates regarding that?

.buildkite/pipeline.yml
```diff
@@ -22,6 +18,7 @@ steps:
       if [[ $BUILDKITE_BRANCH == "master" ]]; then
         ormolu-live/deploy.sh
       else
-        nix-build -A ormoluLive.ormoluLive --arg ormoluLiveLink false -j 1
+        nix-build -A ormoluLive.ormoluLive -j 1 \
+          --arg ormoluLiveLink false --argstr ormoluCompiler ghc8107
```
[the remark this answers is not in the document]
amesgen: GHCJS does not work on GHC >8.10; it is supposed to be integrated as a [regular GHC backend in 9.6](https://gitlab.haskell.org/ghc/ghc/-/issues/21078) (which is supposed to be released [in January](https://mail.haskell.org/pipermail/ghc-devs/2022-October/020973.html)). Also, the [WASM backend](https://gitlab.haskell.org/ghc/ghc/-/issues/21200) based on Asterius is also planned to land there.

That means that we have two options:

 - Best-effort support for 8.10, in order to keep GHCJS builds for Ormolu Live.
 - Stop updating Ormolu Live until 9.6, and drop GHC 8.10 support.

Also see the bullet point "Ormolu Live" in the PR description.

cabal.project
```diff
@@ -4,3 +4,9 @@ constraints: ormolu +dev
 
 -- miso does not specify bounds on servant
 constraints: servant >=0.18 && <0.20
+
+-- prune unnecessary dependencies
+constraints: warp -x509, wai-app-static -cryptonite
+
+-- due to GHCJS 8.10
+allow-older: ghc-lib-parser:base
```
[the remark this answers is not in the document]
amesgen: See https://github.com/tweag/ormolu/pull/885#discussion_r1012073561

default.nix
amesgen: Yes, it merges (in the Nix module system sense) it with the arguments to `haskell-nix.project`. This is motivated by this (see PR description):

> Also, we now only set the GHC flags added in https://github.com/tweag/ormolu/pull/793 when actually building the executable, as otherwise the entire `ghc` library is rebuilt due to changed flags (which takes quite some time).

cabal.project
```diff
@@ -4,3 +4,9 @@ constraints: ormolu +dev
 
 -- miso does not specify bounds on servant
 constraints: servant >=0.18 && <0.20
+
+-- prune unnecessary dependencies
+constraints: warp -x509, wai-app-static -cryptonite
```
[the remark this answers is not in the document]
amesgen: Not on CI, but when you locally enter the `ormolu-live` directory, this will mean that you don't have to compile the big transitive dependency tree of `tls` (`basement`, `cryptonite` etc.).

src/Ormolu/Printer/Meat/Common.hs
amesgen: Removed `p_hsDocString` :+1:

9999years:
Is this still blocked on `haskell.nix` support? We've had GHC 9.4 in nixpkgs for a while now, and releasing GHC 9.4 support for Ormolu is blocking [GHC 9.4 support for Fourmolu](https://github.com/fourmolu/fourmolu/pull/242). In turn, this is blocking [improved GHC 9.4 support in haskell-language-server](https://github.com/haskell/haskell-language-server/pull/3317). I'd really love to get this released on Hackage, even if `haskell.nix` can't build it, because everything else using GHC 9.4 can.

georgefst:
> Is this still blocked on `haskell.nix` support? We've had GHC 9.4 in nixpkgs for a while now, and releasing GHC 9.4 support for Ormolu is blocking [GHC 9.4 support for Fourmolu](https://github.com/fourmolu/fourmolu/pull/242). In turn, this is blocking [improved GHC 9.4 support in haskell-language-server](https://github.com/haskell/haskell-language-server/pull/3317). I'd really love to get this released on Hackage, even if `haskell.nix` can't build it, because everything else using GHC 9.4 can.

This is slightly confusing: do you mean to suggest that HLS is blocked on Fourmolu, but somehow not on Ormolu directly? It seems to me that Fourmolu is mostly irrelevant to the issue at hand (although we are certainly looking forward to this PR being merged, and grateful as ever for the effort!).

9999years:
> do you mean to suggest that HLS is blocked on Fourmolu, but somehow not on Ormolu directly?

The Fourmolu maintainers don't want to incorporate+release this patch to add GHC 9.4 support until Ormolu does. HLS can't support Fourmolu or Ormolu on GHC 9.4 until they can be built on GHC 9.4.

Fourmolu is directly relevant for me (we use it to format our Haskell code at Mercury) but only related to this issue as an example of downstream work on GHC 9.4 that's blocked (directly or indirectly) on this PR being merged.

Either way, I would like to reiterate that I'm extremely grateful to @amesgen for putting this patch together & for all the maintainers and reviewers who are working to merge it.

If there's anything I can do to accelerate these efforts, please let me know.

mrkkrp (approved):

PR #887, opened 2022-05-13
Fix some bugs in the diff printing code

### LINKED ISSUE #886: Diff displays hunks incorrectly in some cases
When diffing these two inputs:

```haskell
testPermParser :: Permutation Parser String
testPermParser =
  f <$> toPermutationWithDefault 'x' (char 'a')
    <*> toPermutationWithDefault 'y' (char 'b')
    <*> toPermutationWithDefault 'z' (char 'c')
  where
    f a b c = [a, b, c]
```

and

```haskell
testPermParser :: Permutation Parser String
testPermParser =
  f
    <$> toPermutationWithDefault 'x' (char 'a')
    <*> toPermutationWithDefault 'y' (char 'b')
    <*> toPermutationWithDefault 'z' (char 'c')
  where
    f a b c = [a, b, c]
```

The result is

```diff
@@ -148,5 +148,6 @@
  testPermParser :: Permutation Parser String
  testPermParser =
-   f <$> toPermutationWithDefault 'x' (char 'a')
+   f
+     <$> toPermutationWithDefault 'x' (char 'a')
    where
      f a b c = [a, b, c]
```

which is not correct. Instead of the line with `where` it is the line with `<*> toPermutationWithDefault 'y' (char 'b')` that should follow. Proper diff reports:

```diff
@@ -147,7 +147,8 @@ prsp' p = prs' (runPermutation p)

 testPermParser :: Permutation Parser String
 testPermParser =
-  f <$> toPermutationWithDefault 'x' (char 'a')
+  f
+    <$> toPermutationWithDefault 'x' (char 'a')
     <*> toPermutationWithDefault 'y' (char 'b')
     <*> toPermutationWithDefault 'z' (char 'c')
   where
```

We should match that.

This seems to have been broken ever since I introduced the functionality for nice displaying of diffs.

mrkkrp:
Closes #886

### DIFF
--- a/src/Ormolu/Diff/Text.hs
+++ b/src/Ormolu/Diff/Text.hs
@@ -38,10 +38,7 @@ data TextDiff = TextDiff
     -- whole diff will be displayed.
     textDiffSelectedLines :: IntSet
   }
-  deriving (Eq)
-
-instance Show TextDiff where
-  show (TextDiff path _ _) = "TextDiff " ++ show path ++ " _ _"
+  deriving (Eq, Show)
 
 -- | List of lines tagged by 'D.Both', 'D.First', or 'D.Second'.
 type DiffList = [D.Diff [Text]]
@@ -57,6 +54,7 @@ data Hunk = Hunk
     hunkSecondLength :: Int,
     hunkDiff :: DiffList
   }
+  deriving (Show)
 
 ----------------------------------------------------------------------------
 -- API
@@ -211,8 +209,8 @@ toHunks = go 0 False id id []
       [] ->
         if gotChanges
           then
-            let p = reverse (take margin bothHistory)
-                currentAcc' = addBothAfter p currentAcc
+            let currentAcc' = addBothAfter p currentAcc
+                p = take margin (reverse bothHistory)
              in case formHunk (currentAcc' []) of
                   Nothing -> hunksAcc []
                   Just hunk -> hunksAcc [hunk]
@@ -235,12 +233,12 @@ toHunks = go 0 False id id []
           piece ->
             if gotChanges
               then
-                let p = reverse bothHistory
-                    currentAcc' = currentAcc . addBothBefore p (piece :)
+                let currentAcc' = currentAcc . addBothBefore p (piece :)
+                    p = reverse bothHistory
                  in go 0 True hunksAcc currentAcc' [] xs
               else
-                let p = reverse (take margin bothHistory)
-                    currentAcc' = addBothBefore p (piece :)
+                let currentAcc' = addBothBefore p (piece :)
+                    p = reverse (take margin bothHistory)
                  in go 0 True hunksAcc currentAcc' [] xs
     addBothBefore [] acc = acc
     addBothBefore p acc = (D.Both p p :) . acc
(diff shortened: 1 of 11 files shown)


src/Ormolu/Diff/Text.hs
```diff
@@ -235,12 +233,12 @@ toHunks = go 0 False id id []
...
-                let p = reverse bothHistory
-                    currentAcc' = currentAcc . addBothBefore p (piece :)
+                let currentAcc' = currentAcc . addBothBefore p (piece :)
+                    p = reverse bothHistory
                  in go 0 True hunksAcc currentAcc' [] xs
               else
-                let p = reverse (take margin bothHistory)
-                    currentAcc' = addBothBefore p (piece :)
+                let currentAcc' = addBothBefore p (piece :)
+                    p = reverse (take margin bothHistory)
```
amesgen, line 241: Just for clarity: There is no actual change here, `p` and `currentAcc'` were just swapped, right?

src/Ormolu/Diff/Text.hs
```diff
@@ -211,8 +209,8 @@ toHunks = go 0 False id id []
       [] ->
         if gotChanges
           then
-            let p = reverse (take margin bothHistory)
-                currentAcc' = addBothAfter p currentAcc
+            let currentAcc' = addBothAfter p currentAcc
+                p = take margin (reverse bothHistory)
```
amesgen, line 213: The bug fix in this PR is to change
```haskell
p = reverse (take margin bothHistory)
```
to
```haskell
p = take margin (reverse bothHistory)
```
here. Why is this not necessary below, where we will use the first variant?

amesgen (commented):
This indeed fixes the bug, but I have not yet fully understood how `toHunks` works in detail. Still, I left two comments/questions below.

src/Ormolu/Diff/Text.hs
mrkkrp: Yes, I just wanted to have `currentAcc'` and `p` defined in the same order everywhere. Could not resist!

src/Ormolu/Diff/Text.hs
mrkkrp: I have to admit that this `toHunks` function is not the clearest code I have written. In short, the case that I changed is used to pick up lines that are matching in both snippets that being compared when these lines go **after** the differing part. Given that `bothHistory` has the most recent line as its head (so, it is kind of reversed) we should indeed first reverse it back and then take at most `margin` lines:

```
<differing part>
line 1 -----
line 2   |  | take margin (reverse bothHistory)
line 3 --|--
line 4   |
...      | bothHistory, the head is "line N"
line N --
```

the other case is about history lines that precede the differing part:

```
line 1----
line 2    | bothHistory, the head is "line N"
line 3    |
...       |
line N-2 ---
line N-1  | | reverse (take margin bothHistory)
line N------
<differing part>

amesgen: Ah, and we maintain `bothHistory` in reversed form to prevent quadratic time for appending to it. Thanks for the explanation!

amesgen (approved):

PR #896, opened 2022-06-13
Support multithreading

parsonsmatt:
This PR adds support for multithreading when checking multiple files.

Motivation: for whatever reason, `fourmolu` is [extremely slow](https://github.com/fourmolu/fourmolu/issues/205) with the newest version. I'm diagnosing how/why, but as a first pass, I thought I'd try throwing some concurrency at the problem.

EDIT: Converting to draft. This implementation has super garbled output.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -38,6 +38,7 @@ import Paths_ormolu (version)
 import System.Exit (ExitCode (..), exitWith)
 import qualified System.FilePath as FP
 import System.IO (hPutStrLn, stderr)
+import UnliftIO.Async
 
 -- | Entry point of the program.
 main :: IO ()
@@ -58,7 +59,7 @@ main = do
             ExitSuccess -> Nothing
             ExitFailure n -> Just n
       errorCodes <-
-        mapMaybe selectFailure <$> mapM (formatOne' . Just) (sort xs)
+        mapMaybe selectFailure <$> pooledMapConcurrentlyN optThreads (formatOne' . Just) (sort xs)
       return $
         if null errorCodes
           then ExitSuccess
@@ -185,7 +186,9 @@ data Opts = Opts
     -- | Source type option, where 'Nothing' means autodetection
     optSourceType :: !(Maybe SourceType),
     -- | Haskell source files to format or stdin (when the list is empty)
-    optInputFiles :: ![FilePath]
+    optInputFiles :: ![FilePath],
+    -- | The amount of threads to use when parsing multiple files.
+    optThreads :: !Int
   }
 
 -- | Mode of operation.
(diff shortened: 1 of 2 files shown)


app/Main.hs
```diff
@@ -58,7 +59,7 @@ main = do
             ExitSuccess -> Nothing
             ExitFailure n -> Just n
       errorCodes <-
-        mapMaybe selectFailure <$> mapM (formatOne' . Just) (sort xs)
+        mapMaybe selectFailure <$> pooledMapConcurrentlyN optThreads (formatOne' . Just) (sort xs)
```
amesgen, line 62: As most of Ormolu should be CPU-bound, consider adding `-threaded -with-rtsopts=-N`: 

https://github.com/tweag/ormolu/blob/25b04d45b4f3e8db81bc4a863f2fededda7bc384/ormolu.cabal#L155

app/Main.hs
```diff
@@ -262,6 +265,12 @@ optsParser =
       [ metavar "FILE",
         help "Haskell source files to format or stdin (the default)"
       ]
+    <*> (fmap (fromMaybe 1) . optional . option auto . mconcat)
```
amesgen, line 268: I think it would be a bit more idiomatic to use [`value`](https://hackage.haskell.org/package/optparse-applicative-0.17.0.0/docs/Options-Applicative.html#v:value) and [`showDefault`](https://hackage.haskell.org/package/optparse-applicative-0.17.0.0/docs/Options-Applicative.html#v:showDefault) instead of `fmap (fromMaybe 1) . optional`.

amesgen (changes requested):
Thanks for taking a look! I am not sure if this has to live in Ormolu (especially if it turns out to be complicated), but I am not really opposed (there might be opportunities for caching certain things compared to invoking Ormolu multiple times, but I don't think that this is the bottleneck right now). I still left a few comments below :+1:

FTR, this is an easy way to run Ormolu in parallel, using [fd](https://github.com/sharkdp/fd):
```shell
fd -e hs . /path/to/my/dir -x ormolu -m check
```

---

In any case, the slowdown you describe in https://github.com/fourmolu/fourmolu/issues/205 sounds worrying; I just compared Ormolu 0.5.0.0 to 0.3.1.0, and it is a bit slower (~30% for some random project), but not nowhere near as a bad as in your case. I guess that the code is not open source, please let us now when you had time to look into this!

parsonsmatt:
Thanks for the tip on `fd` - I got a complete `check` of the codebase in 4:50 with that. Even if this doesn't land, that *should* get us under the threshold for updating.

Curiously, the multithreading implementation here does not result in a speedup. With `--jobs 8` and the `-threaded -rtsopts=-N` change, we're using all cores, and mostly at 100%, but it's still taking 20 minutes.

parsonsmatt:
I'm going to close this PR out since the implementation garbles output pretty badly in `check` mode and *isn't any faster* than single threaded execution. Very strange.

PR #911, opened 2022-07-15
Resolve directories as inputs (Fixes #910)

### LINKED ISSUE #910: CLI: Support specifying input folders in addition to just input files
It would be nice if fourmolu could take folders as well as files as an input and would process all haskell files in that folder (recursively)
so instead of
```
fourmolu -i $(find src -name '*.hs')
```
I'd be able to do just
```
fourmolu -i src
```
This is consistent what formatters in other languages support and would be a nice quality of life improvement.

matobet:


mrkkrp:
As [was mentioned before](https://github.com/tweag/ormolu/issues/910#issuecomment-1185294029) we prefer to keep the kind of logic outside of Ormolu for better composability and simplicity.

PR #916, opened 2022-08-12
Sort imports by package qualifier if present (Fixes #905)

### LINKED ISSUE #905: Could imports of the same package be grouped when using `PackageImports` ?

Consider:

```
{-# LANGUAGE PackageImports #-}

import "servant-server" Servant (Context (EmptyContext), Get, Handler, NamedRoutes, type (:>))
import "servant" Servant.API.Generic ((:-))
import "servant-blaze" Servant.HTML.Blaze (HTML)
import "servant-server" Servant.Server.Generic (AsServerT, genericServeTWithContext)
```

Ormolu formats them this way, when it might be nice if the imports from the same package (`servant-server`) where together?

To sort first by package name, instead of module name, when `PackageImports` is turned on.

n/a

**Additional context**

n/a

-- edit: ( Removing the `style` tag because I think I can only add that as I'm a member of tweag, not because of the repository settings; I'll leave it to a maintainer to make that call 😅 )

matthew-healy:
This fixes #905 by sorting imports first by package qualifier,
if any are included, and then by module name. Imports without
package qualifiers are always sorted before those with.

### DIFF
--- a/src/Ormolu/Imports.hs
+++ b/src/Ormolu/Imports.hs
@@ -64,8 +64,8 @@ combineImports (L lx ImportDecl {..}) (L _ y) =
 -- the same 'ImportId' they can be merged.
 data ImportId = ImportId
   { importIsPrelude :: Bool,
-    importIdName :: ModuleName,
     importPkgQual :: Maybe LexicalFastString,
+    importIdName :: ModuleName,
     importSource :: IsBootInterface,
     importSafe :: Bool,
     importQualified :: Bool,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+* Imports are now sorted by package qualifier, if one is present.
+  [Issue 905](https://github.com/tweag/ormolu/issues/905).
+
 ## Ormolu 0.5.0.1
 
 * Fixed a bug in the diff printing functionality. [Issue
--- a/data/examples/import/sorted-package-imports-out.hs
+++ b/data/examples/import/sorted-package-imports-out.hs
@@ -1,0 +1,7 @@
+{-# LANGUAGE PackageImports #-}
+
+import D
+import "a" Ab
+import "b" Aa
+import "b" Bb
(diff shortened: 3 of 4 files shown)


src/Ormolu/Imports.hs
```diff
@@ -30,6 +30,7 @@ normalizeImports :: [LImportDecl GhcPs] -> [LImportDecl GhcPs]
 normalizeImports =
   fmap snd
     . M.toAscList
+    . M.mapKeys OrdByPkgQualImportId
```
amesgen, line 33: One could also reuse the `Ord` instance for tuples to signal the "prioritization" of `importPkgQual`, which would not require a `newtype`:
```suggestion
    . M.mapKeys (\i -> (importPkgQual i, i))
```
Or even more direct: Just swap `importPkgQual` and `importIdName` in the definition of `ImportId`, such that the derived `Ord` instance does the work for us.

matthew-healy: This is a great suggestion, thank you! I've gone with the tuple `Ord` suggestion rather than moving `importPkgQual` because it seemed a little more explicit to me. That said, I'm fairly new to writing Haskell so if relying on the ordering of record definitions is idiomatic then let me know & I'll be happy to change it.

amesgen: Yeah, it is certainly more implicit, but upon second look, I think we actually already make use of the field ordering: We have `importIsPrelude :: Bool` as the first field, such that i.e.
```haskell
import A
import Z
import Prelude
```
is a fixed point. If we want to keep that, it would make sense to adapt the field ordering. @mrkkrp WDYT?

matthew-healy: Ah, well if we're already relying on it I guess there's no harm in continuing to do so. I'll update the PR now.

mrkkrp:
LGTM, please add a changelog note (add a new "unreleased" section).

matthew-healy:
@mrkkrp Done!

Issue #908, opened 2022-07-12
Improvements in the online formatter


asarkar:
It'll be very helpful to have the following in the online formatter:
1. A "copy" option for the right pane that copies the formatted code to the clipboard.
2. Line numbers: Without those, an error message referring to a line is practically useless.

matthew-healy:
I'm currently looking at the first request here.

PR #918, opened 2022-08-16
[ormolu-live] Add a button to copy formatted code

matthew-healy:
This was requested in #908.

It adds a "Copy" button which allows a user to copy the formatted code to their clipboard.

## clipboard.js

Because support for the browser [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) isn't consistent between browsers (e.g. Safari will only allow calls to `writeText` to happen in user gesture event handlers, and the [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API) which will allow requesting more extensive clipboard permissions isn't yet implemented in most browsers), I opted to use the [clipboard.js](https://clipboardjs.com/) library. The minified source is included directly in the first commit.

`clipboard.js` is loaded directly into JSaddle when running on GHC, and is included as a script tag when using GHCJS. I also included the JS to instantiate & hold a reference to the `ClipboardJS` JavaScript object in a script tag, though I'd be happy to hear of a better way of doing this as it feels a little dirty to be writing JS inside an HTML DSL in Haskell (but maybe that's just me... 😅).

## UI

The styling of the UI is a little hacked together currently - Bulma doesn't have any support built-in for "floating" a button over other content (even though this UI paradigm is used to copy code blocks in the Bulma documentation...). I settled on an approach whereby the right-hand column is a Bulma `card` (with shadow & radius removed), and the button is both `card-content` and `is-overlay`, which means it acts as an overlay on the parent card, rather than the whole screen. I then had to change the button's relative position to get it to sit over the top-left corner.

On larger screens this looks pretty good:

ormolu-copy-large

However, as the screen width gets smaller, it starts to slip over the edge of the output div:

ormolu-copy-small

If anyone has suggestions on improving the layout here I'd be happy to receive them.


ormolu-live/src/Main.hs
```diff
@@ -96,6 +96,8 @@ viewModel model@Model {..} =
   div_
     []
     [ link_ [rel_ "stylesheet", href_ "bulma.min.css"],
+      script_ [src_ "clipboard.min.js"] "",
+      script_ [] "var c = new ClipboardJS('.copy-output')",
```
amesgen, line 100: I think this is fine, invoking this via JSaddle would be more ceremony without much benefit. Small suggestion:
```suggestion
      script_ [] "new ClipboardJS('.copy-output');",
```

amesgen (changes requested):
Looks great; I think the layout of the copy button is fine, even on mobile.

This needs minor changes to actually work with GHCJS:
```diff
diff --git a/default.nix b/default.nix
index 2a0cb7d..ad1b38d 100644
--- a/default.nix
+++ b/default.nix
@@ -295,8 +295,7 @@ in {
       };
       buildInputs = [ pkgs.closurecompiler ];
       installPhase = ''
-        mkdir -p $out
-        find . \( -name '*.html' -o -name '*.css' \) -exec cp {} $out \;
+        cp -r . $out
         ORMOLU_LIVE=${ormoluLive}/bin/ormolu-live.jsexe
         closure-compiler \
           $ORMOLU_LIVE/all.js --externs $ORMOLU_LIVE/all.js.externs \
diff --git a/ormolu-live/src/Main.hs b/ormolu-live/src/Main.hs
index 2d9e1c4..7b4ccb7 100644
--- a/ormolu-live/src/Main.hs
+++ b/ormolu-live/src/Main.hs
@@ -96,7 +96,6 @@ viewModel model@Model {..} =
   div_
     []
     [ link_ [rel_ "stylesheet", href_ "bulma.min.css"],
-      script_ [src_ "clipboard.min.js"] "",
       script_ [] "var c = new ClipboardJS('.copy-output')",
       section_ [class_ "section"] . pure . div_ [class_ "container is-fluid"] $
         [ h1_ [class_ "title"] [text "Ormolu Live"],
diff --git a/ormolu-live/www/index.html b/ormolu-live/www/index.html
index b7ae6c1..a50f785 100644
--- a/ormolu-live/www/index.html
+++ b/ormolu-live/www/index.html
@@ -4,6 +4,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Ormolu Live</title>
+    <script src="clipboard.min.js"></script>
   </head>
   <body>
     <script src="all.min.js"></script>
```
I have a test deployment here: https://62fcb04342f26c08d08eb888--ormolu-live-testbed.netlify.app/

ormolu-live/src/Main.hs
matthew-healy: Huh, I could have sworn I tried not assigning the `ClipboardJS` object to a variable & it didn't work, but I've just tried it again now and it works fine. Will update now, thanks!

amesgen (approved):

PR #920, opened 2022-08-18
[Ormolu-Live] Add line numbers to editor (Fixes #908)

### LINKED ISSUE #908: Improvements in the online formatter

matthew-healy:
This uses a (mostly) pure-CSS solution (inspired by [this post](https://www.webtips.dev/add-line-numbers-to-html-textarea)) to add line numbers to the text editor in Ormolu Live. It calculates the desired line number count from the input, and then creates an equivalent number of `span` elements. A CSS counter is then incremented for each span, and the content is set to the counter's value.

As part of this I had to stop letting Bulma style the editor, so I took the opportunity to give it a dark background & white text, which I think looks pretty nice. I'm also happy to try and find something closer to the original style if anyone has particularly strong feelings about it. 

I initially looked at bringing in a dependency for this, but every one I found creates its own `textarea` rather than using the one provided, which means it'd be a huge pain to wire the browser input events back up to the `SetInput` action. Doing it this way has some drawbacks (primarily that the editor's `textarea` can no longer scroll separately from the rest of the page, as otherwise the line numbers would get out of sync) but is significantly simpler overall than the alternatives. 

Merging this will close #908.

### How it looks

ormolu-line-numbers

### DIFF
--- a/ormolu-live/src/Main.hs
+++ b/ormolu-live/src/Main.hs
@@ -96,6 +96,7 @@ viewModel model@Model {..} =
   div_
     []
     [ link_ [rel_ "stylesheet", href_ "bulma.min.css"],
+      link_ [rel_ "stylesheet", href_ "editor.css"],
       script_ [] "new ClipboardJS('.copy-output');",
       section_ [class_ "section"] . pure . div_ [class_ "container is-fluid"] $
         [ h1_ [class_ "title"] [text "Ormolu Live"],
@@ -132,9 +133,15 @@ viewModel model@Model {..} =
             [class_ "columns"]
             [ div_
                 [class_ "column is-half is-flex"]
-                [ textarea_
-                    [class_ "textarea is-family-code", onInput SetInput, rows_ "20", autofocus_ True]
-                    [text input]
+                [ div_
+                    [class_ "editor"]
+                    [ div_
+                        [class_ "line-numbers"]
+                        (replicate (editorLineNumbers . fromMisoString $ input) (span_ [] [])),
+                      textarea_
+                        [class_ "is-family-code", onInput SetInput, rows_ "20", autofocus_ True]
+                        [text input]
+                    ]
                 ],
               div_
                 [id_ "output", class_ "column is-half is-flex card is-shadowless is-radiusless"]
@@ -252,6 +259,8 @@ viewModel model@Model {..} =
               . Dump.showAstData Dump.NoBlankSrcSpan Dump.NoBlankEpAnnotations
               $ prParsedSource
           O.RawSnippet r -> r
+
+    editorLineNumbers text = 1 + T.count "\n" text
 
 extractOrmoluException :: SomeException -> Either String O.OrmoluException
 extractOrmoluException = \case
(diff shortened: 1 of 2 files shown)


ormolu-live/www/editor.css
```diff
@@ -0,0 +1,38 @@
...
+
+.line-numbers span {
+    counter-increment: linenumber;
+}
+
+.line-numbers span::before {
+    content: counter(linenumber);
+    display: block;
+    color: #506882;
+}
```
amesgen, line 38: Nitpick: add a trailing newline for consistency?

amesgen (commented):
Looks nice, I like the layout! Only very minor comments below.

One thing that could be discussed: The line numbers are probably particularly useful when the input does *not* fit on a single page: If an error is near the bottom and one could scroll the left pane independently of the right one (which is currently not possible as you mention), one could see the faulty code and the parse error side-by-side. I think this PR is still improves this situation without this exact usecase, but maybe @asarkar has an opinion here as the author of the motivating issue?

Test deployment: https://6300f6cbf0b2c955c2040b18--ormolu-live-testbed.netlify.app

ormolu-live/www/editor.css
asarkar: IMO, the line numbers are always useful, regardless of how long the output is. How else is a user supposed to find a particular line with an error?
Also, please note that copy formatted output option was introduced https://github.com/tweag/ormolu/pull/918, and it’s important to not break that functionality. That is, copied output must not contain the line numbers because that’d make it uncompilable code.

ormolu-live/src/Main.hs
```diff
@@ -253,6 +260,11 @@ viewModel model@Model {..} =
               $ prParsedSource
           O.RawSnippet r -> r
 
+    editorLineNumbers text
+      | T.null text = 1
+      | T.last text == '\n' = 1 + length (T.lines text)
+      | otherwise = length $ T.lines text
```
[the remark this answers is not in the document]
matthew-healy: Ohhh this is much cleaner. Thank you! 🙏

ormolu-live/www/editor.css
matthew-healy: This doesn't impact the copy functionality I added - that copies the text from the output `div`, whereas this adds line numbers beside the input `textarea`, so it shouldn't have any impact. 👍

matthew-healy:
@amesgen Totally agree that the overall editing experience can be improved further. It should be possible to integrate something like [CodeFlask](https://kazzkiq.github.io/CodeFlask/) or [CodeMirror](https://codemirror.net/) with a medium amount of JSaddle wrangling, and either of those should improve the experience (& mean there's less that has to be maintained in Ormolu-Live). As mentioned in the PR description, the main issue I ran into there was that these libraries expect to introduce their own `textarea`, which means we have to find that object in the document and attach the `onInput` handler to it ourselves. It seemed to make sense to get the simplest version of this feature shipped ASAP, and come back to work out an improved version later on. 

In the meantime, I've made the changes you requested, but I'm currently on free airport WiFi which for whatever reason won't let me push to GitHub, so I won't be able to update the PR for a few hours.

amesgen (approved):

PR #932, opened 2022-11-06
Use located in GADT

brandonchinn178:
Is there a particular reason why `p_hsType` was being called on `unLoc quantifiedType` instead of using the `located` combinator? It seems like it was initially implemented when GADT args + result type were separate, so there wasn't a single HsType object representing the entire type to be located, but now that we have `quantifiedType`, it seems better to be consistent between function signatures + GADT constructors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@
 * Handle `UnicodeSyntax` variants more consistently. [Issue
   934](https://github.com/tweag/ormolu/issues/934).
 
+* Fix an inconsistency in formatting of types in GADT declarations in
+  certain cases. [PR 932](https://github.com/tweag/ormolu/pull/932).
+
 ## Ormolu 0.5.0.1
 
 * Fixed a bug in the diff printing functionality. [Issue
--- a/data/examples/declaration/data/gadt/multiline-out.hs
+++ b/data/examples/declaration/data/gadt/multiline-out.hs
@@ -14,12 +14,8 @@ data Foo a where
     Foo 'Int
   -- | But 'Bar' is also not too bad.
   Bar ::
-    Int ->
-    Maybe Text ->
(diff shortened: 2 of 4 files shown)


amesgen (approved):
Nice, the use of `unLoc` here just seems to be a historical oversight :+1:

Just for reference what the inconsistency fixed by this PR is concretely:
<table>
<th>Input</th><th>Output before this PR</th><th>Output with this PR</th>
<tr>
<td>

```haskell
foo ::
  A -> B
```
</td>
<td>

```haskell
foo ::
  A -> B
```
</td>
<td>

```haskell
foo ::
  A -> B
```
</td>
</tr>
<tr>
<td>

```haskell
data Foo where
  Foo :: 
    A -> B
```
</td>
<td>

```haskell
data Foo where
  Foo :: 
    A ->
    B
```
</td>
<td>

```haskell
data Foo where
  Foo :: 
    A -> B
```
</td>
</tr>
</table>

mrkkrp (approved):

mrkkrp:
I can't figure out why CI is not working on this PR :thinking:

brandonchinn178:
Do you have CI enabled for PRs originating from forks?

mrkkrp:
No, that would be unsafe. I vaguely remember that it should work if I force-push on your PR (which I did). Anyway, it worked magically when I pushed `origin/gadt-cleanup`, even though the branch of the PR itself remained unchanged :thinking: 

I also granted you write access. Use it wisely and don't change the indentation step, please.

brandonchinn178:
Ah sure; Fourmolu uses GitHub Actions, which lets you click a button to manually approve running a workflow. Not sure how buildkite does it.

> Use it wisely and don't change the indentation step, please.

Ha. Even though I work on Fourmolu, I actually prefer 2-space indentation. Ironic

Issue #906, opened 2022-07-11
Comment in empty export list moved out


brandonchinn178:
**Describe the bug**
```hs
module Foo
  ( {- comment -}
  )
where
```
gets reformatted to
```hs
module Foo
  (
  )
where

{- comment -}
```

Related: #901 

**To Reproduce**

**Expected behavior**
It should keep the comment in the export list

**Environment**
 - OS name + version: MacOS
 - Version of the code: HEAD

**Additional context**
This isn't an issue if there's anything in the export list, before or after the comment. So it's normally not an issue, since one usually exports at least one thing. But it's messing up my new library, which generates an export item at the spot specified by the comment. I currently workaround with a hack involving named sections, which is fine for my purposes, since haddocks aren't generated for tests:

https://github.com/brandonchinn178/tasty-autocollect/blob/f4890360a4870d2740e01d2c54f424b07c0ef88d/test/Test/Tasty/Ext/TodoTest.hs#L5

brandonchinn178:
@amesgen hm this seems to break with block comments:
```hs
module Foo
  ( {- asdf -}
  )
where
```
Format without checking idempotence:
```hs
module Foo
{- asdf -}(
          )
where
```
Fixed point:
```hs
module Foo
{- asdf -} (
           )
where
```

amesgen:
Thanks, that is unfortunate, I will have a look at this, especially as your PR #933 handles this properly. For some context from my POV:

 - The current comment layout logic has several very nice properties:
    - You basically don't have to think about comments while writing printing code; they are emitted automatically via `located`, which is very convenient and a significant reduction in cognitive load.
    - The comment printing logic is separated from printing the AST. 
    - It *generally* produces very good to decent output, and when it does not, it is usually easy to put the comment in a slightly different place for it to work.
 - OTOH, almost all currently open bugs are related to comments (in particular idempotence), and many style requests are also related to comments (one of the most prominent ones: #469).
 - It is probably possible to fix most of these issues by adding lots of special cases, and also starting to interleave comment and AST printing. Any individual such fix is probably not a big deal, but together, they could introduce significant complexity, in particular to the comparatively low severity of the issues they are resolving.
 - "Proper" solutions could include:
    - Fix these problems by identifying a self-contained change the existing comment printing logic.
    - Switch to a completely different way of printing comments, e.g. using the comment-to-AST association present in the GHC AST via TTG/EPA (trees-that-grow/exact-print-annotations); also see #324 for some considerations.

PR #933, opened 2022-11-06
Fix comment in empty export list

brandonchinn178:
Resolves https://github.com/tweag/ormolu/issues/906

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Fix comment in empty export list being moved out [Issue 906](https://github.com/tweag/ormolu/issues/906)
+
 ## Ormolu 0.5.3.0
 
 * Stop making empty `let`s move comments. [Issue
--- a/data/examples/module-header/multiline-empty-comment-out.hs
+++ b/data/examples/module-header/multiline-empty-comment-out.hs
@@ -1,0 +1,4 @@
+module Foo
+  ( -- test
+  )
+where
--- a/data/examples/module-header/multiline-empty-comment.hs
+++ b/data/examples/module-header/multiline-empty-comment.hs
@@ -1,0 +1,2 @@
+module Foo ( -- test
+           ) where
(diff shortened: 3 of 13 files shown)


brandonchinn178:
@mrkkrp is it possible to get this PR reviewed any time soon?

amesgen:
@mrkkrp and me actually talked about this some time ago, but I forgot to comment here; sorry for the delay!

The changes are well-separated, but due to the heuristic nature of comment placing, the changes in espc. the last commit are rather ad-hoc, and introduce (albeit relatively isolated) complexity that seems somewhat disproprionate compared to the low severity of the issue they are resolving, as you note in #906.

Therefore, I played a bit around to see if there maybe is an easier way to fix the issue, and I think the core issue for this (and related bugs) is that when a specific region with a list of items (like a module exports list, but also a record field declaration, see the following example) is empty, there is no chance for a comment to be printed inside because there is no "last" element before which it could be printed. Concretely, this means that the comments in this example "escape" their enclosing brackets:
```haskell
module Foo
  ( -- a
  )
where

data Test = Test {
  -- b
  }
```
Hence, an idea would be to use located with bogus spans corresponding to the opening and closing bracket. I implemented this [here](https://github.com/tweag/ormolu/compare/master...amesgen/issue-906-alternative), and it indeed seems to work in principle (and might even be adapted to work e.g. with the record example above): it passes all tests, but there are some idempotency issues in the `hackageTests`.

I might have some time this week to see if the idempotency issue of this approach can be resolved relatively easily, but also feel free to have a look, as you might have considered something similar when working on this PR, but discarded it (or other approaches).

brandonchinn178:
That does look much better 😃 I do like some of the commits in this branch that clean up some of the comments code (e.g. the popComment commit is much more readable IMO), so maybe I can put those up in a separate cleanup PR. But :+1: to your approach

brandonchinn178:
@amesgen any update on your separate approach?

amesgen:
> @amesgen any update on your separate approach?

Sorry, I finally opened #1013. I agree that some commits in this PR would be nice to incorporate just as generic code cleanup, feel free to open a PR for those!

Issue #639, opened 2020-07-31
Allow GHC DynFlags to passed directly to the API


bubba:
**Is your feature request related to a problem? Please describe.**
Currently in the Ormolu plugin in haskell-language-server, we have access to the GHC `ParsedModule`. The whole plugin is short enough to include here so here it is in all its glory:

```haskell
provider :: FormattingProvider IO
provider _lf ideState typ contents fp _ = do
  let
    fromDyn :: ParsedModule -> IO [DynOption]
    fromDyn pmod =
      let
        df = ms_hspp_opts $ pm_mod_summary pmod
        pp =
          let p = D.sPgm_F $ D.settings df
          in  if null p then [] else ["-pgmF=" <> p]
        pm = map (("-fplugin=" <>) . moduleNameString) $ D.pluginModNames df
        ex = map (("-X" <>) . show) $ S.toList $ D.extensionFlags df
      in
        return $ map DynOption $ pp <> pm <> ex

  m_parsed <- runAction "Ormolu" ideState $ getParsedModule fp
  fileOpts <- case m_parsed of
          Nothing -> return []
          Just pm -> fromDyn pm

  let
    fullRegion = RegionIndices Nothing Nothing
    rangeRegion s e = RegionIndices (Just s) (Just e)
    mkConf o region = defaultConfig { cfgDynOptions = o,  cfgRegion = region }
    fmt :: T.Text -> Config RegionIndices -> IO (Either OrmoluException T.Text)
    fmt cont conf =
      try @OrmoluException (ormolu conf (fromNormalizedFilePath fp) $ T.unpack cont)

  case typ of
    FormatText -> ret <$> fmt contents (mkConf fileOpts fullRegion)
    FormatRange r ->
      let
        Range (Position sl _) (Position el _) = normalize r
      in
        ret <$> fmt contents (mkConf fileOpts (rangeRegion sl el))
```
One hairy thing is that although we have direct access to the `ParsedModule`'s `DynFlags`, we still need to print it to a string with this hacky `fromDyn` flag, to then parse to Ormolu.  

**Describe the solution you'd like**
To avoid this workaround it would be nice if the public API exposed a field that allowed `DynFlags` to be set directly instead of doing this roundtrip from `DynFlags` to strings back to `DynFlags`, which seem to get converted back here:
https://github.com/tweag/ormolu/blob/35f4fb39326ab7d4847c5c8940f4d4185f3d3d24/src/Ormolu/Parser.hs#L193-L207

Something like

[...]

PR #948, opened 2022-12-26
Use `ghc` instead of `ghc-lib-parser` when possible

sol:
This significantly reduces the build time with recent version of GHC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ dist-*/
 dist/
 stack.yaml.lock
 result*
+
+# These are touched on every :reload in GHCi.
+# https://gitlab.haskell.org/ghc/ghc/-/issues/22669
+*.o-boot
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -230,7 +230,7 @@ optsParserInfo =
               $gitBranch,
               $gitHash
             ],
-          "using ghc-lib-parser " ++ VERSION_ghc_lib_parser
+          ghcApiVersion
         ]
     exts :: Parser (a -> a)
     exts =
@@ -239,6 +239,13 @@ optsParserInfo =
           help "Display extensions that need to be enabled manually"
         ]
     displayExts = unlines $ sort (showOutputable <$> manualExts)
+
+ghcApiVersion :: String
+#ifdef VERSION_ghc_lib_parser
+ghcApiVersion = "using ghc-lib-parser " ++ VERSION_ghc_lib_parser
+#else
+ghcApiVersion = "using ghc " ++ VERSION_ghc
+#endif
 
 optsParser :: Parser Opts
 optsParser =
--- a/ormolu.cabal
+++ b/ormolu.cabal
@@ -105,7 +105,6 @@ library
         dlist >=0.8 && <2.0,
         exceptions >=0.6 && <0.11,
         filepath >=1.2 && <1.5,
-        ghc-lib-parser >=9.4 && <9.5,
         megaparsec >=9.0,
         mtl >=2.0 && <3.0,
         syb >=0.7 && <0.8,
@@ -113,7 +112,14 @@ library
         text >=0.2 && <3.0,
         th-lift-instances >=0.1 && <0.2
 
-    mixins:           ghc-lib-parser hiding (Language.Haskell.TH.Syntax)
+    if impl(ghc >=9.4 && <9.5)
+        build-depends:
+            ghc,
+            ghc-boot-th
+
+    else
+        build-depends: ghc-lib-parser >=9.4 && <9.5
+        mixins:        ghc-lib-parser hiding (Language.Haskell.TH.Syntax)
 
     if (flag(fixity-th) && impl(ghc <9.4))
         cpp-options: -DFIXITY_TH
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -27,3 +27,7 @@ dist-*/
 dist/
 stack.yaml.lock
 result*
+
+# These are touched on every :reload in GHCi.
+# https://gitlab.haskell.org/ghc/ghc/-/issues/22669
+*.o-boot
```
sol, line 33: Without this files being ignored [`hspec/sensei`](https://github.com/hspec/sensei) goes into a reload-loop.

With this change `sensei` can be used during development, e.g. try:

    cabal exec -- sensei -XHaskell2010 tests/Spec.hs -isrc -itests --focused-only

Or add a `.ghci` file with:

```
-- .ghci
:set -isrc
:set -itests
:set -XHaskell2010
```
so that this becomes:

    cabal exec -- sensei tests/Spec.hs --focused-only

amesgen (commented):
That definitely makes sense in general, and other tools like `hlint` also offer a flag to do this, though hlint disabled using `ghc` by default when possible, but I do not yet fully understand why they chose to do so: https://github.com/ndmitchell/hlint/pull/1410

mrkkrp:
I applied the gitignore-related change separately in #953.

We are a bit undecided about the main part though. One thing we do not want to do for sure is introducing a cabal flag because then we would need a bigger CI matrix and longer CI times. If we force `ghc` for 9.4 then there are two reservations:
* Ormolu might stop being agnostic to whether `ghc-lib-parser` or `ghc` in used at some point, e.g. if a change like https://github.com/tweag/ormolu/issues/639 is implemented.
* Sometimes there are GHC parsing bugs on just one patch version. E.g. suppose there is a bug in 9.4.1 which is fixed in 9.4.2, and you are for some reason forced to compile Ormolu with 9.4.1, then after this PR you can't get a working Ormolu anymore.

sol:
> I applied the gitignore-related change separately in #953.

Thanks for that! That makes sense!

> We are a bit undecided about the main part though. One thing we do not want to do for sure is introducing a cabal flag

👍 less flags are always a good thing, especially if they are `manual`!

> * Ormolu might stop being agnostic to whether `ghc-lib-parser` or `ghc` in used at some point, e.g. if a change like [Allow GHC DynFlags to passed directly to the API #639](https://github.com/tweag/ormolu/issues/639) is implemented.

I would hope this can be addressed by CI. Testing e.g. `9.2.5` and `9.4.4` on CI should make sure that both `ghc-lib-parser` and `ghc` work.

> * Sometimes there are GHC parsing bugs on just one patch version. E.g. suppose there is a bug in 9.4.1 which is fixed in 9.4.2, and you are for some reason forced to compile Ormolu with 9.4.1, then after this PR you can't get a working Ormolu anymore.

From what I understand you have the same issue if you accept multiple versions of `ghc-lib-parser`, right now we are accepting `9.4.*`. From what I understand making sure that we have "parity" between the versions of `ghc` and `ghc-lib-parser` that we allow can address this.

Examples:

Right now we have:
```cabal
if impl(ghc >=9.4 && <9.5)
  build-depends: ghc
else
  build-depends: ghc-lib-parser >=9.4 && <9.5
```

If we discovered a bug in `9.4.1` then we would change this to:
```cabal
if impl(ghc >=9.4.2 && <9.5)
  build-depends: ghc
else
  build-depends: ghc-lib-parser >=9.4.2 && <9.5
```

If at some later point we decided that everybody should be on the same single lastest version then we would change this again to:

```cabal
if impl(ghc ==9.4.4)
  build-depends: ghc
else
  build-depends: ghc-lib-parser ==9.4.4.*
```
Does this make sense?

I am not committed to this change by any means though. It just seemed like a low hanging fruit in terms of build  times. But I'm OK either way.

sol:
> CPP might be necessary in that case

How is that?

I was under the impression that the packages are interchangeable as long as you do not exercise too much of `ghc` (say `ghc-lib-parser` strictly being a subset of `ghc`). 

But I might be wrong here.

> I think it is easier to just specify a different version of `ghc-lib-parser` (the one where the problem is fixed, this is even possible without editing the `.cabal` file)

Do you mean that a user would do this themselves by manually constraining the version?

(say by passing `--constraint="ghc-lib-parser ==9.4.4.*` to `cabal` or alternatively constraining the version through other means like `stack` or `nix`)

> In general, I guess we just prefer simplicity and are not very much effected by longer build times.

That's fine with me. Please feel free to close this PR.

mrkkrp:
> How is that?
> 
> I was under the impression that the packages are interchangeable as long as you do not exercise too much of `ghc` (say `ghc-lib-parser` strictly being a subset of `ghc`).

I think there some parts where the two packages diverge, but I do not know for sure. @amesgen might know more details.



> Do you mean that a user would do this themselves by manually constraining the version?
> 
> (say by passing `--constraint="ghc-lib-parser ==9.4.4.*` to `cabal` or alternatively constraining the version through other means like `stack` or `nix`)

Yes, this is what I had in mind.

amesgen:
> > How is that?
> > I was under the impression that the packages are interchangeable as long as you do not exercise too much of `ghc` (say `ghc-lib-parser` strictly being a subset of `ghc`).
> 
> I think there some parts where the two packages diverge, but I do not know for sure. @amesgen might know more details.

Yeah, differences for the same patch version would probably be a bug, but for different patch versions, they might differ, although I don't think we have ever experienced this so far (concretely: suppose one is compiling with GHC 9.4.1, but ghc-lib-parser has version 9.4.2).

The original point was that

> Ormolu might stop being agnostic to whether `ghc-lib-parser` or `ghc` in used at some point, e.g. if a change like

which was about a usability concern of downstream users such as HLS, i.e. that was the motivation why hlint changed the default: https://github.com/ndmitchell/hlint/issues/1376

mrkkrp:
Okay, closing for now.

PR #949, opened 2022-12-26
Don't use `unsafePerformIO` without need

sol:


amesgen (approved):
Thanks, this reminds me that we could also switch to an actual golden tests library to avoid reimplementing features like this one.

mrkkrp:
Thanks!

PR #960, opened 2023-01-05
Switch to `binary` for parsing the fixity DB

amesgen:
Closes #959


mrkkrp (approved):
Excellent job!

brandonchinn178:
FYI JSON still referenced here: https://github.com/tweag/ormolu/blob/810eb7839cbf4e358821447e8f4603955999e62d/extract-hackage-info/src/Main.hs#L47-L48

PR #963, opened 2023-01-06
De`String`ify

### LINKED ISSUE #972: Is String type considered legacy?
I have a look into Ormolu code base. 
It is using `String` type everywhere. Is there expectations that switching to `Data.Text` would provide performance gain? 

Recently, on Haskell Weekly I saw an article about xhtml library refactoring shifting towards Text.

amesgen:
Retroactive cause: Closes #972

When thinking about e.g. #959, it always bugged me that we use `String`s in so many places. This PR removes the two primary occurences:

 - Package and operator names in the fixity logic.

   Here, we now use `PackageName` from `Cabal-syntax` (as that is where we get the package names from) and a newtype wrapper around `ShortByteString` for operator names (could also use `FastString` or `ShortText`, should be rather easy to change).
 - As the input type to the `ormolu` function, now `Text`.
 
   For this, I wrote an efficient conversion function `Text -> StringBuffer`.

Apart from better type safety, we get speedups:

For fixity parsing overhead (see #959 for the setup, `ormolu` is 0.5.2.0 here):

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `ormolu --no-cabal test-a.hs` | 1.5 ± 0.4 | 0.9 | 4.5 | 1.00 |
| `result-nostrs/bin/ormolu --no-cabal test-a.hs` | 2.6 ± 0.7 | 1.7 | 7.2 | 1.69 ± 0.62 |
| `ormolu --no-cabal test-b.hs` | 37.1 ± 2.1 | 34.5 | 52.3 | 24.37 ± 6.46 |
| `result-nostrs/bin/ormolu --no-cabal test-b.hs` | 20.1 ± 1.1 | 18.6 | 22.6 | 13.23 ± 3.50 |
| `ormolu --no-cabal test-c.hs` | 84.8 ± 3.2 | 82.0 | 96.4 | 55.71 ± 14.57 |
| `result-nostrs/bin/ormolu --no-cabal test-c.hs` | 60.2 ± 3.2 | 57.0 | 75.0 | 39.58 ± 10.46 |

Test suite execution time also dropped by 10% from 940ms to 840 ms.


src/Ormolu/Utils.hs
```diff
@@ -150,3 +157,17 @@ matchAddEpAnn :: AnnKeywordId -> AddEpAnn -> Maybe EpaLocation
...
+  buf <- mallocPlainForeignPtrBytes (len + 3)
+  withForeignPtr buf $ \ptr -> do
+    TFFI.unsafeCopyToPtr txt ptr
+    -- last three bytes have to be zero for easier decoding
+    pokeElemOff ptr len 0
+    pokeElemOff ptr (len + 1) 0
+    pokeElemOff ptr (len + 2) 0
+  pure StringBuffer {buf, len, cur = 0}
+  where
+    len = TFFI.lengthWord8 txt
```
amesgen, line 173: If desired, this could also be written by using `ByteString` as an intermediary type (which is then more obviously correct), but it would incur two copies instead of just one here.

mrkkrp: I am okay with this implementation, but perhaps `unsafePerformIO` is not necessary? Let it live in `IO` and we can plug in it at some point in `ormolu`, since it already has the `MonadIO m` constraint. Functions called from `ormolu` and its variants could take `StringBuffer` directly instead of accepting `Text` and re-converting it.

src/Ormolu/Fixity/Internal.hs
```diff
@@ -79,8 +95,36 @@ instance Semigroup FixityInfo where
           (Just a, Just b) | a == b -> Just a
           _ -> Nothing
 
+-- | An operator name.
+newtype OpName = MkOpName
+  { -- | Invariant: UTF-8 encoded
+    getOpName :: ShortByteString
```
mrkkrp, line 101: Is there a difference in performance if this is `Text` instead? Why should it be a `ShortByteString` (or even `ShortText`)?

src/Ormolu/Fixity/Internal.hs
```diff
@@ -1,12 +1,19 @@
...
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ViewPatterns #-}
 
 module Ormolu.Fixity.Internal
-  ( FixityDirection (..),
+  ( OpName (..),
```
mrkkrp, line 12: Should the constructor be exported if we attach an invariant to the data inside? Perhaps we should only allow construction via smart constructors?

src/Ormolu/Fixity/Internal.hs
amesgen: I would be surprised if there were a big difference in all cases other than synthetic benchmarks. My motivation to use `ShortByteString` was that GHC already uses them internally, so converting them to `OpName` is free, see `occOpName`.

src/Ormolu/Fixity/Internal.hs
amesgen: Thanks for catching that, the pattern synonym exists exactly such that `OpName` can be opaque :smile: Will remove the constructor export.

src/Ormolu/Utils.hs
amesgen: Hmm, that sounds good, but we use `textToStringBuffer` e.g. in the pure `tokenize` function `Ormolu.Parse.Pragma`, and also manipulate the input via `Data.Text`, so don't see an easy way to both keep the functions there pure and avoid converting back from `StringBuffer` to `Text`.

mrkkrp: Yeah, I don't have a good idea how to avoid `IO` in that context.

mrkkrp (approved):

mrkkrp:
Note to self: return to this once #940 is merged.

src/Ormolu/Utils.hs
yaitskov: [`stringBufferFromByteString`](https://hackage.haskell.org/package/ghc-lib-parser-9.4.4.20221225/docs/GHC-Data-StringBuffer.html#v:stringBufferFromByteString) is defined next to `textToStringBuffer` and doing the same under the hood, but doesn't expose IO.

I have an idea to avoid copying at all.
StringBuffer is wrapper around ForeignPtr:

```
data StringBuffer
 = StringBuffer {
     buf :: {-# UNPACK #-} !(ForeignPtr Word8),
     len :: {-# UNPACK #-} !Int,        -- length
     cur :: {-# UNPACK #-} !Int         -- current pos
  }
```
there is a [mmap](https://hackage.haskell.org/package/mmap-0.5.9/docs/System-IO-MMap.html) library which can mmap file and produce ForeignPtr:
```
System.IO.MMap.mmapFileForeignPtr :: FilePath                     -- ^ name of file to map
                   -> Mode                         -- ^ access mode
                   -> Maybe (Int64,Int)            -- ^ range to map, maps whole file if Nothing
                   -> IO (ForeignPtr a,Int,Int)
```

Let's mmap file in read/write mode with size + 3 bytes and zero last 3 bytes.

In case of shared memory causes exception then fallback to current approach with copying.

yaitskov: Writing zero must not be required because OS should take care of zeroing new blocks because of security.

amesgen: Thanks for the suggestion, but I don't think it is possible to make this work, as we need those three extra zero bytes at the end of the `ForeignPtr`, which does not seem to be possible via `mmapFileForeignPtr` (we don't want to actually modify the file).

ormolu.cabal
```diff
@@ -130,7 +130,7 @@ executable ormolu
         th-env >=0.1.1 && <0.2,
         optparse-applicative >=0.14 && <0.18,
         ormolu,
-        text >=0.2 && <3.0
+        text >=2.0 && <3.0
```
brandonchinn178, line 133: Why bump the lower bound? Stackage still doesn't have `text-2.x`, so this prevents building Ormolu (easily) with Stack: https://github.com/tweag/ormolu/pull/984

amesgen: It is due to the `textToStringBuffer` function that makes use of the UTF-8 encoding, see e.g. here: https://github.com/tweag/ormolu/pull/963#discussion_r1063453308

W.r.t. stack: I thought that as one already has to touch `extra-deps` due to `ghc-lib-parser-9.4` in the current LTS, this wouldn't be a big deal, but maybe I am missing a different downside (I mostly don't use stack).

brandonchinn178: :+1: as long as there's a reason. The problem is that when specifying a boot library in `extra-deps`, Stack ignores all other boot libraries, so you have to explicitly specify all the boot libraries you need. Plus, bumping text requires bumping other deps like `hashable`, so there's a cascading effect. Not a big deal (personally, I don't think there's much reason to install ormolu/fourmolu as a library anyway)

PR #964, opened 2023-01-06
Add WASM-based Ormolu Live

### LINKED ISSUE #951: Online formatter should allow copying parts of the output
Why prevent copying from the output pane? Often times I don't want to copy everything, just a portion of it, and execute that part of the code.

The online formatter doesn't allow copying portions of the code.

It should be possible to select and copy portions of output, although editing shouldn't be allowed.

Copy-pasting into a text editor.

**Additional context**
N/A

amesgen:
New Ormolu Live based on the upcoming [GHC WASM backend](https://www.tweag.io/blog/2022-11-22-wasm-backend-merged-in-ghc/).

Closes #951

Test deployment at https://ormolu-live-testbed.netlify.app/

Diff is severly inflated by lockfile changes.


amesgen:
@asarkar As you have been giving feedback on the old Ormolu Live version, I would be interested in your feedback, in particular if you think that the new version resolves #951.

https://ormolu-live-testbed.netlify.app/

asarkar:
@amesgen Thanks for doing this, I’ll take a look later. One unrelated comment, can you display a verbiage on the page about the storage/usage of code that is being formatted? In other words, if you’re not saving any of the code or phoning home,  an you call it out, so that people formatting sensitive code know what to expect.

amesgen:
> One unrelated comment, can you display a verbiage on the page about the storage/usage of code that is being formatted? In other words, if you’re not saving any of the code or phoning home or using any part of it to train the tool, can you call it out, so that people formatting sensitive code know what to expect.

Good point, I could imagine that it might even be illegal under the GDPR to send the data without explicit consent. I added this line:

> Note that this website is entirely client-side; in particular, your input is never sent to a remote server.

asarkar:
I've had a chance to look at the new page, and here are my findings:
1. The copy button on the right is gone, which, IMO, is a no-go.
2. It's not clear which of the two panes is for the user. A note like "paste or type here" that disappears as soon as something is pasted or typed would make for a better user experience. For reference, see https://codebeautify.org/jsonvalidator and https://text-compare.com/.
3. IMO, "See the GitHub repository" button is redundant because you're already linking to a commit in the GitHub repo.
4. The disclaimer "this website is entirely client-side" is better suited as a footnote. 
5. In addition to the rows, it'll be helpful to show the column at which the cursor is at. Here's why:

Input `abc <- xs` results in:
```
<input>:1:5-6
  parse error on input `<-'
  ```
  
Now, while the last line does say what the problem is, this is an exception rather than the rule as Haskell is notorious for hard-to-understand compiler errors. Column 5 happens to be the beginning of `<-`.

**Comment by amesgen** [comment]
> I've had a chance to look at the new page, and here are my findings:

Thanks, I just updated the site based on your findings with only slight deviations (ghc-lib-parser version is next to the commit link, and I left off point 3 for now, as I find it somewhat neat to be able to directly jump to the README).

**Comment by asarkar** [comment]
LGTM :thumbsup:

**Comment by asarkar** [comment]
> it'll be helpful to show the column

Just want to say that this was a great idea. I pasted the following code at https://ormolu-live.tweag.io/:
```
module Chapter03.Lib () where

class Monad m where
return :: a -> m a
(>>=) :: m a -> (a -> m b) -> m b

then_ :: Maybe a -> (a -> Maybe b) -> Maybe b

instance Monad Maybe where
  return = Just
  (>>=) = then_

-- Exercise3.1. Write the implementation of ap.
ap :: Monad m => m (b -> c) -> m b -> m c
ap f x = f >>= $ \f' -> x >>= $ \x' -> f' x'
```

and it failed with:
```
The GHC parser failed:

<input>:15:16

parse error on input `$'
```
However, column numbers aren't shown in the input, so, it's not immediately apparent which of the two `$` is the problem. Thanks for implementing my suggestions.

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -49,3 +49,88 @@ jobs:
...
+        if: env.NETLIFY_AUTH_TOKEN != ''
+        uses: nwtgck/actions-netlify@v2
+        with:
+          publish-dir: ./frontend
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          alias: ${{ github.event.pull_request.head.sha || github.sha }}
+          enable-pull-request-comment: true
+          enable-commit-comment: false
+          enable-commit-status: true
+      - name: Deploy to Netlify, production
```
mrkkrp, line 133: Why do you use `actions-netlify` in one case and the CLI tool directly in the other?

amesgen: There are three cases:

 - `env.NETLIFY_AUTH_TOKEN == ''` (e.g. a PR from a fork): We want to run neither step.
 - `env.NETLIFY_AUTH_TOKEN != ''` and `github.ref != 'refs/heads/master'` (a PR from a branch in this repo): We want a non-production deployment (using the commit hash as an alias). We use `actions-netlify` as we want e.g. the commit status functionality.
 - `env.NETLIFY_AUTH_TOKEN != ''` and `github.ref == 'refs/heads/master'`: We still want a non-production deployment (as you can't supply an alias when deploying to production, cf. #820), but also a regular deployment, which we do via the CLI as we don't need any functionality from `netlify-action`, so it is more succinct.

Does that make sense?

flake.nix
```diff
@@ -97,13 +109,73 @@
           hooks = {
             nixpkgs-fmt.enable = true;
             deadnix.enable = true;
+            purs-tidy.enable = true;
           };
+          tools = { inherit (ormoluLive) purs-tidy; };
         };
+
+        ormoluLive =
```
mrkkrp, line 117: What do you think about putting this in `default.nix` in `ormolu-live` similar to how it is done for some tests? Isolating Ormolu Live-related Nix expressions could perhaps improve readability of both `flake.nix` and `ormolu-live/default.nix`.

ormolu-live/README.md
```diff
@@ -1,25 +1,65 @@
...
+Make sure to be in the `.#ormoluLive` Nix shell when entering `./ormolu-live`, e.g. conveniently via [nix-direnv][].
+
+### Local interactive development
 
+For building the WASM binary, run
+```console
+wasm32-wasi-cabal update
 ```
-nix-build -A ormoluLive.website
+and then iterate by running sth like
```
mrkkrp, line 26:
```suggestion
and then iterate by running something like
```

ormolu-live/README.md
```diff
@@ -1,25 +1,65 @@
...
+
+https://ormolu-live.tweag.io
+
+## Overview
+
+ATM, the GHC WASM backend only supports emitting [WASI binaries][WASI], which can be run in the browser via e.g. [browser_wasi_shim][] or the more fully-featured [wasmer-wasi][].
+
+Hence, Ormolu Live consists of two parts:
+
+ - A regular Purescript frontend that displays input/output and manages options.
```
mrkkrp, line 13:
```suggestion
 - A regular Purescript frontend (in `src/`) that displays input/output and manages options.
```

ormolu-live/README.md
```diff
@@ -1,25 +1,65 @@
...
+ - A background web worker that formats Haskell source code via the WASM-compiled Ormolu.
```
mrkkrp, line 14:
```suggestion
 - A background web worker (in this directory with the source in `app/Main.hs`) that formats Haskell source code via the WASM-compiled Ormolu.
```

src/Ormolu/Fixity.hs
```diff
@@ -42,12 +42,29 @@ import Data.Set (Set)
...
+import System.IO.Unsafe (unsafePerformIO)
+#endif
 
 packageToOps :: Map PackageName FixityMap
 packageToPopularity :: Map PackageName Int
+#if BUNDLE_FIXITIES
 HackageInfo packageToOps packageToPopularity =
   Binary.runGet Binary.get $
     BL.fromStrict $(embedFile "extract-hackage-info/hackage-info.bin")
+#else
```
mrkkrp, line 60: Can you add a comment explaining why this is necessary?

ormolu.cabal
```diff
@@ -32,6 +32,10 @@ flag dev
     default:     False
     manual:      True
 
+flag internal-bundle-fixities
+    description: INTERNAL FLAG DONT USE
```
mrkkrp, line 36:
```suggestion
    description: An internal ad-hoc flag that is enabled by default, Ormolu live disables it
```

I do not know why, perhaps due to the size of the byte string that would be embedded? Feel free to add a better explanation.

mrkkrp (commented):
Exciting stuff!

src/Ormolu/Fixity.hs
amesgen: Will do :+1: This is due to the WASM backend not yet supporting Template Haskell, see also the commit message of the second commit:

> Should be reverted once the WASM backend supports TH.

ormolu.cabal
amesgen: See https://github.com/tweag/ormolu/pull/964#discussion_r1085635501

mrkkrp (approved):

PR #969, opened 2023-01-17
Accept CRLF line endings in .ormolu files

mrkkrp:


amesgen (approved):
I think the reason why we are even seeing these CRLFs is due to us using the functions from `Ormolu.Utils.IO` which (transitively) uses `withBinaryFile` instead of `withFile` as `Data.Text.IO.readFile` does, but this indeed the right thing to do (see #760 or [the Haddocks of `Data.Text.IO.readFile`](https://hackage.haskell.org/package/text-2.0.1/docs/Data-Text-IO.html#v:readFile)).

Issue #973, opened 2023-01-18
cabal inside nix-shell is not working


yaitskov:
nix-shell starts shell but cabal is not working.

```
[nix-shell:~/pro/ormolu]$ type haskell-language-server
haskell-language-server is /nix/store/cm22flslgspfffa0zwasgzdj8mlfmvmn-haskell-language-server-exe-haskell-language-server-1.8.0.0/bin/haskell-language-server

[nix-shell:~/pro/ormolu]$ type ghc
ghc is /nix/store/9wdivs4vd0gii68d74rw3l6a6dl0c1ap-ghc-shell-for-ormolu-ghc-9.2.4-env/bin/ghc

[nix-shell:~/pro/ormolu]$ type cabal
cabal is hashed (/nix/store/x2ribvhlx8vs7sdg4x993x05xjr86d3b-cabal-install-exe-cabal-3.8.1.0/bin/cabal)

[nix-shell:~/pro/ormolu]$ git log -1
commit 708de707ad44172d9a6019e6f09b48ddc3782170 (HEAD -> master, origin/master, origin/HEAD)
Author: Alexander Esgen <alexander.esgen@tweag.io>
Date:   Fri Jan 6 14:40:38 2023 +0100

    Stop making empty `let`s move comments
```

```

[nix-shell:~/pro/ormolu]$ cabal build
Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
update' to download it.
Resolving dependencies...
Error: cabal: Could not resolve dependencies:
[__0] trying: extract-hackage-info-0.0.0.0 (user goal)
[__1] unknown package: tagsoup (dependency of extract-hackage-info)
[__1] fail (backjumping, conflict set: extract-hackage-info, tagsoup)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: extract-hackage-info, tagsoup


[nix-shell:~/pro/ormolu]$ cabal exec -v0 -- ghc --print-libdir
Error: cabal: Could not resolve dependencies:
[__0] trying: extract-hackage-info-0.0.0.0 (user goal)
[__1] unknown package: tagsoup (dependency of extract-hackage-info)
[__1] fail (backjumping, conflict set: extract-hackage-info, tagsoup)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: extract-hackage-info, tagsoup

```

deleting ~/.cabal doesn't help.

amesgen:
[...]

PR #974, opened 2023-01-18
cabal-update-shell-hook update cabal if it is bad in shellHook

yaitskov:


default.nix
```diff
@@ -99,7 +99,12 @@ in {
     exactDeps = false;
   }; in {
     inherit hsPkgs;
-    ormoluShell = shellFor (ps: [ ps.ormolu ]);
+    ormoluShell = (shellFor (ps: [ ps.ormolu ]))
```
mrkkrp, line 102: It bother me to modify the state of the system implicitly like this, but I will leave the decision to @amesgen.

default.nix
```diff
@@ -99,7 +99,12 @@ in {
...
+      .overrideAttrs(oa: { shellHook = (oa.shellHook or "") + ''
+                             cabal exec -v0 -- ghc --print-libdir && echo "Cabal is good" || \
+                               { echo "Update Cabal"; cabal update; }
+                           '';
+                         });
```
amesgen, line 107:
```suggestion
    ormoluShell = shellFor (ps: [ ps.ormolu ps.extract-hackage-info ]);
```

amesgen (changes requested):
Thanks; I think a conceptually somewhat nicer way to fix this is to have switch to having just one combined shell for `ormolu` and `extract-hackage-info`, as mentioned in https://github.com/tweag/ormolu/issues/973#issuecomment-1387092995 and already done in https://github.com/tweag/ormolu/pull/940, as it avoids touching the global system state, as mentioned by @mrkkrp.

For this, feel free to just apply this diff: https://gist.github.com/amesgen/11eaf0ee65796f97059873016d027bf4

default.nix
amesgen: I suggested an alternative that does not have this caveat: https://github.com/tweag/ormolu/pull/974#pullrequestreview-1262270371

amesgen:
#940 has been merged, which also resolved #973 as per https://github.com/tweag/ormolu/issues/973#issuecomment-1387092995.

PR #975, opened 2023-01-19
Normalize parentheses in constraints

### LINKED ISSUE #264: Normalize parens
A common deviation in styles is the following:

```haskell
foo :: (Show a) => a -> IO ()
bar :: Show a => a -> IO ()
```

There are also contrived examples, where extra parens cannot possibly help readability:

```
baz :: ((Show a)) => a -> IO ()
```

Some might argue that constraints should always have parens around them, even singleton constraints, because it's more regular and less work when additional constraints have to be added to the set. Others prefer to always write the minimal number of lexemes. Point is, this is a common matter of style, that could be normalized away.

mrkkrp:
Close #264.

### DIFF
--- a/src/Ormolu/Parser.hs
+++ b/src/Ormolu/Parser.hs
@@ -50,7 +50,7 @@ import Ormolu.Utils (incSpanLine, showOutputable)
 
 -- | Parse a complete module from string.
 parseModule ::
-  MonadIO m =>
+  (MonadIO m) =>
   -- | Ormolu configuration
   Config RegionDeltas ->
   -- | Fixity map to include in the resulting 'ParseResult's
@@ -90,7 +90,7 @@ parseModule config@Config {..} fixityMap path rawInput = liftIO $ do
   pure (warnings, snippets)
 
 parseModuleSnippet ::
-  MonadIO m =>
+  (MonadIO m) =>
   Config RegionDeltas ->
   LazyFixityMap ->
   DynFlags ->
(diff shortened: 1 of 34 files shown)


src/Ormolu/Parser.hs
```diff
@@ -173,11 +181,23 @@ normalizeModule hsmod =
...
+      x -> x
+    enforceParensInClassConstraints :: TyClDecl GhcPs -> TyClDecl GhcPs
+    enforceParensInClassConstraints = \case
+      ClassDecl {..} -> ClassDecl {tcdCtxt = patchContext <$> tcdCtxt, ..}
+      x -> x
+    patchContext :: LHsContext GhcPs -> LHsContext GhcPs
+    patchContext (L l qs) = L l $ case qs of
+      [x@(L _ (HsParTy _ _))] -> [x]
+      [x@(L lx _)] -> [L lx (HsParTy EpAnnNotUsed x)]
+      xs -> xs
```
amesgen, line 200: Hmm, what about directly matching for `LHsContext` instead of only matching for a fixed list of types wherein it is contained? This would have the advantage of making it impossible to miss cases. The only caveat might be that due to `LHsContext` just being a type synonym for a located list of types, but if GHC were to start using a located list of types for something else, they would probably want different EPAs for them (as there can only be one [`Anno (LHsContext pass)`](https://hackage.haskell.org/package/ghc-9.4.4/docs/Language-Haskell-Syntax-Extension.html#t:Anno), so they would create newtypes.

Concretely, this would catch constraints in existential types and GADTs: Even with this PR, this is a fixed point:
```haskell
data Foo = forall a. Show a => Foo a

data Bar where
  Bar :: Show a => a -> Bar
```
With e.g. https://gist.github.com/amesgen/9ebd0bd7cc03689e6013c1f926a20345, parens are added, and at least the test suite passes.

amesgen (changes requested):

src/Ormolu/Parser.hs
mrkkrp: I also thought of matching on `LHsContext`, however it being a type synonym for a list of located types (a thing that seemed too common to be safe to modify) dissuaded me from following that path. I do agree with your reasoning regarding the `Anno` type family. Good one!

amesgen (approved):
:tada:

PR #976, opened 2023-01-20
Expose a more complete API from the Ormolu module

### LINKED ISSUE #924: Squash warnings about Setup.hs

Most packages use a `Setup.hs` which is not directly used by stack or cabal, but is used by `runhaskell` (as mentioned [here](https://cabal.readthedocs.io/en/3.6/cabal-package.html#pkg-field-build-type)). Ormolu always prints a warning for this file:

```
Found .cabal file myapp/myapp.cabal, but it did not mention myapp/Setup.hs
```

Ormolu could recognize this as a known file when it finds a cabal file and simply attempt to format it.

I've tried marking Setup.hs as an extra source file:

```
extra-source-file:
    Setup.hs
```

but this does not squash the warning and I am not sure if this is appropriate.

### LINKED ISSUE #971: False positive - cabal doesn't have a ref to a source file
ormolu 0.5.0.1 prints warning:

```
Found .cabal file napkin.cabal, but it did not mention integration-tests/haskell/Sleep.hs
```

meanwhile cabal specifies source directory:

```
hs-source-dirs: src
```

I thinks it should not look into directories not mentioned by cabal.

mrkkrp:
Close #970, close #971, close #924.


mrkkrp:
This is a step towards exposing pretty much all logic of Ormolu CLI from the `Ormolu` module and making it "official". It still requires some manual wiring from the user to do what the CLI does, because I wanted to keep the helpers versatile.

src/Ormolu/Utils/Cabal.hs
```diff
@@ -59,20 +59,18 @@ defaultCabalInfo =
...
   FilePath ->
-  -- | Extracted cabal info
-  m CabalInfo
+  -- | Extracted cabal info, if any
+  m (Maybe CabalInfo)
 getCabalInfoForSourceFile sourceFile = liftIO $ do
   findCabalFile sourceFile >>= \case
-    Just cabalFile -> parseCabalInfo cabalFile sourceFile
-    Nothing -> do
-      hPutStrLn stderr $ "Could not find a .cabal file for " <> sourceFile
```
amesgen, line 74: There is another place where we log to stderr in this module:

https://github.com/tweag/ormolu/blob/5a43ffc288df4ebce718340da5e7586931de61db/src/Ormolu/Utils/Cabal.hs#L147-L151

Now the question is whether we want to indicate in the return type whether we couldn't find a `.cabal` file or whether it didn't mention the source file...

amesgen (changes requested):

src/Ormolu/Utils/Cabal.hs
mrkkrp: I think we need to introduce a type that would describe the exact outcome:

```haskell
data CabalSearchResult = CabalNotFound | DidNotMention | CabalFound CabalInfo
```

And then we can do printing as required in `Main.hs` and erase this info by using `defaultCabalInfo` when necessary.

amesgen: Yeah, that sounds good, potentially also with a `FilePath` field for `DidNotMention` containing the `.cabal` file.

The alternative would be to pass in some "tracer" (i.e. a wrapped `String -> IO ()` or even `CabalSearchResult -> IO ()`), which also seems reasonable to me.

app/Main.hs
```diff
@@ -139,27 +156,16 @@ formatOne CabalOpts {..} mode reqSourceType rawConfig mpath =
...
-          }
+    patchConfig mdetectedSourceType mcabalInfo = do
+      let sourceType =
+            fromMaybe
+              ModuleSource
+              (reqSourceType <|> mdetectedSourceType)
+      mfixityOverrides <-
+        case mcabalInfo of
+          Nothing -> return Nothing
+          Just cabalInfo -> Just <$> getFixityOverridesForSourceFile cabalInfo
```
amesgen, line 167:
```suggestion
      mfixityOverrides <- traverse getFixityOverridesForSourceFile mcabalInfo
```

amesgen (approved):

PR #978, opened 2023-01-24
Gitignore .direnv/

mrkkrp:
This seems to be created when direnv uses a nix flake?


amesgen (approved):
I think it comes from [nix-direnv](https://github.com/nix-community/nix-direnv), I also have it in non-flake based projects.

I have this in my global `.gitignore` :smile:

PR #980, opened 2023-01-26
Update flake inputs, add GHC 9.4.4 to CI, minor tweaks 

amesgen:
- Update haskell.nix
    - adds GHC 9.4.4 to CI
    - haskell.nix uses a different cache right now (probably temporarily)
 - Update ghc-wasm-meta
 - Nix 2.13.2 fixed the bug we encountered: https://github.com/NixOS/nix/issues/7644#issuecomment-1404957780
 - Fixes an IFD issue for the macOS binary (it tried to build Linux GHCs...); successful test run is here: https://github.com/amesgen/ormolu/actions/runs/4018230397

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -32,8 +32,8 @@
         inherit (pkgs) lib haskell-nix;
         inherit (haskell-nix) haskellLib;
 
-        defaultGHCVersion = "ghc925";
-        ghcVersions = [ "ghc902" defaultGHCVersion ];
+        ghcVersions = [ "ghc925" "ghc902" "ghc944" ];
+        defaultGHCVersion = builtins.head ghcVersions;
         perGHC = lib.genAttrs ghcVersions (ghcVersion:
           let
             hsPkgs = pkgs.haskell-nix.cabalProject {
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -33,7 +33,7 @@
         inherit (haskell-nix) haskellLib;
 
         defaultGHCVersion = "ghc925";
-        ghcVersions = [ "ghc902" defaultGHCVersion ];
+        ghcVersions = [ "ghc902" defaultGHCVersion "ghc944" ];
```
mrkkrp, line 36: A minor nit: I think it would be actually better to write `"ghc925"` here instead of `defaultGHCVersion`. The reason is that if we change `defaultGHCVersion` to be e.g. `"ghc944"` we need to be careful to remember to update `ghcVersions` accordingly because it could as well become `["ghc902" "ghc944" "ghc944"]`, in other words, the way it is written one cannot just update `defaultGHCVersion` and be done, one has to remember more context about how it is used in order to keep things correct.

mrkkrp (approved):

flake.nix
amesgen: Good point, what about
```nix
ghcVersions = [ "ghc925" "ghc902" "ghc944" ];
defaultGHCVersion = builtins.head ghcVersions;
```
?

mrkkrp: This sounds good to me!

amesgen: https://github.com/tweag/ormolu/compare/b0fdd9b08d69aebe67bbc945f827af6c32ee01fd..0d55b81c4656024dceb5805e51b62cf94d33c84a

PR #984, opened 2023-01-29
Fix stack build

brandonchinn178:


amesgen (approved):
Thanks for paying attention to this, can confirm this compiles locally :+1:

@mrkkrp Do we want to reconsider adding a CI job for Stack?

PR #985, opened 2023-01-30
Implement Term as pure operation

brandonchinn178:
Implement Term as an AST builder + interpreter, so that tests and ormolu-live (and Fourmolu) don't have to use hacky work-arounds to print exceptions.

The monad instance is a bit sus, but no one should really be using `<-` in any `Term` blocks. If someone does, it'll just blow up in tests. We can get rid of the monad weirdness by just writing everything as
```hs
prettyOrmoluException :: OrmoluException -> PrettyText
prettyOrmoluException = \case
  OrmoluParsingFailed s e ->
    [ WithBold [putOutputable s]
    , ...
    ]

type PrettyText = [PrettyNode]
data PrettyNode
  = OutputText Text
  | WithColor Color PrettyText
  | WithBold PrettyText

hPutPretty :: ColorMode -> Handle -> PrettyText -> IO ()
getPretty :: PrettyText -> Text
```
But the do-notation does look a bit nicer, so I didn't want to rewrite all of that for now.

### DIFF
--- a/src/Ormolu/Terminal.hs
+++ b/src/Ormolu/Terminal.hs
@@ -1,13 +1,14 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecordWildCards #-}
 
 -- | An abstraction for colorful output in terminal.
 module Ormolu.Terminal
-  ( -- * The 'Term' monad
+  ( -- * The 'Term' abstraction
     Term,
     ColorMode (..),
     runTerm,
+    runTermPure,
 
     -- * Styling
     bold,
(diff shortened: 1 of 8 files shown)


src/Ormolu/Terminal.hs
```diff
@@ -17,114 +17,121 @@ module Ormolu.Terminal
...
+  fmap _ (TermOutput nodes) = TermOutput nodes
+
+instance Applicative TermOutput where
+  pure _ = TermOutput []
+  TermOutput nodes1 <*> TermOutput nodes2 = TermOutput (nodes1 <> nodes2)
+
+instance Monad TermOutput where
+  TermOutput nodes1 >>= k =
+    let TermOutput nodes2 = k (error "TermOutput unexpectedly evaluated continuation")
+     in TermOutput (nodes1 <> nodes2)
```
amesgen, line 57: If you want, I think we could use `QualifiedDo` here to avoid these somewhat weird instances as you note, i.e. make `TermOutput` monomorphic, `newtype`-derive `Monoid`, define
```haskell
(>>) :: TermOutput -> TermOutput
(>>) = (<>)
```
and then make the call sites qualified and use `Term.do` instead of `do`.

Also see the ["Monoids" example](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0216-qualified-do.rst#examples) in the original proposal.

amesgen (commented):
Seems reasonable, thanks for going for it!

You probably need to account for the removed re-exports in [mtl-2.3](https://hackage.haskell.org/package/mtl-2.3.1/changelog) for Ormolu Live to compile.

src/Ormolu/Terminal.hs
brandonchinn178: Nice! I went even further and removed the Applicative instance (cause I still wasn't crazy about the Functor/Applicative instances). I'm less happy about the fact that I needed to implement new `for'`, `when'`, and `unless'` combinators, but I do like the removal of the phantom type parameter. Thoughts?

src/Ormolu/Terminal.hs
```diff
@@ -17,114 +17,113 @@ module Ormolu.Terminal
...
-import System.IO (Handle, hFlush, hPutStr)
+import System.IO (Handle, hFlush)
 
 ----------------------------------------------------------------------------
 -- The 'Term' monad
 
--- | Terminal monad.
-newtype Term a = Term (ReaderT RC IO a)
-  deriving (Functor, Applicative, Monad)
+newtype Term = Term [TermOutputNode]
```
amesgen, line 38: I'd be surprised if it would make a difference, but as we already depend on `dlist`, what about using them here?

src/Ormolu/Terminal.hs
amesgen: I think I might very slightly prefer `Applicative` as one can then reuse `for_`/`when`/`unless`, but feel free to leave it as is :+1:

Also note that `for' = flip foldMap`, so it could probably be inlined and would not need a dedicated binding.

amesgen: You can even derive the `Applicative` using `DerivingVia`:
```haskell
newtype Term' a = Term (DList TermOutputNode)
  deriving (Functor, Applicative) via Const [TermOutputNode]
```

src/Ormolu/Terminal.hs
brandonchinn178: I personally dislike `dlist`, preferring `Data.Sequence` more. I do notice that we're only using it in one place; I'll use dlist here, but I'll make a followup PR to replace DList with Seq

src/Ormolu/Terminal.hs
```diff
@@ -1,13 +1,14 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecordWildCards #-}
 
 -- | An abstraction for colorful output in terminal.
 module Ormolu.Terminal
   ( -- * The 'Term' monad
```
amesgen, line 7: Maybe "applicative" or "abstraction" instead of "monad"?

src/Ormolu/Terminal/Monad.hs
```diff
@@ -0,0 +1,7 @@
+module Ormolu.Terminal.Monad ((>>)) where
```
amesgen, line 1: Nitpick/feel free to ignore: "Monad" feels slightly misleading as `Term` is no `Monad`; could be `Ormolu.Terminal.Monad` (for comparison: `template-haskell` calls their `QualifiedDo` module [`Language.Haskell.TH.CodeDo`](https://hackage.haskell.org/package/template-haskell-2.19.0.0/docs/Language-Haskell-TH-CodeDo.html)).

amesgen (approved):

PR #986, opened 2023-01-30
Replace DList with Seq

brandonchinn178:
I find `Data.Sequence` has a better API than `DList`, and it avoids adding another dependency (given that most projects use `containers` anyway). The two should have roughly the same performance: https://stackoverflow.com/a/37507346/4966649

Will conflict with https://github.com/tweag/ormolu/pull/985, will resolve after PR is merged


amesgen (approved):

mrkkrp:
Thanks!

PR #988, opened 2023-02-03
Add Ormolu.Terminal.QualifiedDo to the list of modules

mrkkrp:


amesgen (approved):
Ah, we use `-Wno-missing-home-modules`:

https://github.com/tweag/ormolu/blob/4e8fc16479664c845c381d2c8e6cfd952814ffc7/ormolu.cabal#L120

Do we want to change that such that CI can catch these mistakes?

mrkkrp:
It would be nice, do you know why we have `-Wno-missing-home-modules` in the first place?

amesgen:
> It would be nice, do you know why we have `-Wno-missing-home-modules` in the first place?

Git blame points to

https://github.com/tweag/ormolu/blob/9a3d872e33fc57b93686c9cd38d27c24e349ad5c/ormolu.cabal#L124-L125

and https://github.com/haskell/haddock/issues/1116 seems to have been fixed in all GHC/Haddock versions we use.

PR #1007, opened 2023-04-03
Clean up p_hsType

brandonchinn178:
Inline `p_hsType'` after the simplification in #993


brandonchinn178:
ah never mind. `multilineArgs` needs to be computed at the very top level and propagate down

PR #1008, opened 2023-04-03
Fix unicode normalization in diffParseResult

brandonchinn178:
Did a few things:
* Renamed `matchIgnoringSrcSpans`, which was a misleading name (since it ignores a lot more now)
* Consolidated the unicode normalization code to keep it all together (we can do similarly for other rules too)
* Included `HsUniToken` (added in GHC 9.6) in the normalization rules for unicode

### DIFF
--- a/src/Ormolu/Diff/ParseResult.hs
+++ b/src/Ormolu/Diff/ParseResult.hs
@@ -64,11 +64,10 @@ diffCommentStream (CommentStream cs) (CommentStream cs')
 --
 --     * 'SrcSpan's
 --     * ordering of import lists
---     * style (ASCII vs Unicode) of arrows
 --     * LayoutInfo (brace style) in extension fields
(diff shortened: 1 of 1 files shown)


amesgen (commented):
:+1: on the first commit

It is not explicitly stated in your PR, but I guess you were prompted to open this PR by a false positive in AST diffing? I took another look, and indeed, there is something related to GADTs. I opened #1009 with a streamlined fix (when we disregard differences in `HsUniToken`, we can actually remove the special handling of `HsArrow`)  and a regression test.

---

> Included `HsUniToken` (added in GHC 9.6) in the normalization rules for unicode

`HsUniToken` was [already present in 9.4](https://hackage.haskell.org/package/ghc-lib-parser-9.4.4.20221225/docs/Language-Haskell-Syntax-Extension.html#t:HsUniToken), but it is used in more places in 9.6.

PR #1009, opened 2023-04-04
Fix false positives in AST diffing related to `UnicodeSyntax`

amesgen:
Inspired by and alternative to #1008 (except the first commit there).

See the new example snippet for the snippet that caused errors before this PR.

Close #1008.

### DIFF
--- a/src/Ormolu/Diff/ParseResult.hs
+++ b/src/Ormolu/Diff/ParseResult.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeepSubsumption #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ViewPatterns #-}
@@ -51,7 +52,7 @@ diffParseResult
       prParsedSource = hs1
     } =
     diffCommentStream cstream0 cstream1
-      <> matchIgnoringSrcSpans hs0 hs1
+      <> diffHsModule hs0 hs1
 
 diffCommentStream :: CommentStream -> CommentStream -> ParseResultDiff
 diffCommentStream (CommentStream cs) (CommentStream cs')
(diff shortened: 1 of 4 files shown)


src/Ormolu/Diff/ParseResult.hs
```diff
@@ -92,14 +93,17 @@ matchIgnoringSrcSpans a = genericQuery a
...
                   `extQ` classDeclCtxEq
                   `extQ` derivedTyClsParensEq
                   `extQ` considerEqual @EpAnnComments -- ~ XCGRHSs GhcPs
-                  `extQ` considerEqual @TokenLocation
+                  `extQ` considerEqual @TokenLocation -- in LHs(Uni)Token
                   `extQ` considerEqual @EpaLocation
                   `ext2Q` forLocated
+                  -- unicode-related
+                  `extQ` considerEqual @(HsUniToken "->" "→")
+                  `extQ` considerEqual @(HsUniToken "::" "∷")
```
amesgen, line 105: Ideally, we would ignore differences for [`HsUniToken tok utok`](https://hackage.haskell.org/package/ghc-lib-parser-9.4.4.20221225/docs/Language-Haskell-Syntax-Extension.html#t:HsUniToken) for all `Symbol`s `tok` and `utok`. This does not work via `ext2Q` as the `Data (HsUniToken tok utok)` instance requires `KnownSymbol` instead of `Data` for `tok` and `utok`, so a generic solution would instead look like this:
```diff
diff --git a/src/Ormolu/Diff/ParseResult.hs b/src/Ormolu/Diff/ParseResult.hs
index 249c36a..e4a8582 100644
--- a/src/Ormolu/Diff/ParseResult.hs
+++ b/src/Ormolu/Diff/ParseResult.hs
@@ -20,6 +20,7 @@ import GHC.Types.SrcLoc
 import Ormolu.Parser.CommentStream
 import Ormolu.Parser.Result
 import Ormolu.Utils
+import Type.Reflection qualified as TR
 
 -- | Result of comparing two 'ParseResult's.
 data ParseResultDiff
@@ -83,6 +84,9 @@ matchIgnoringSrcSpans a = genericQuery a
           if x' == (y' :: ByteString)
             then Same
             else Different []
+      | Just rep <- isHsUniToken x,
+        Just rep' <- isHsUniToken y =
+          if rep == rep' then Same else Different []
       | typeOf x == typeOf y,
         toConstr x == toConstr y =
           mconcat $
@@ -101,14 +105,18 @@ matchIgnoringSrcSpans a = genericQuery a
                   `extQ` considerEqual @EpaLocation
                   `ext2Q` forLocated
                   -- unicode-related
-                  `extQ` considerEqual @(HsUniToken "->" "→")
-                  `extQ` considerEqual @(HsUniToken "::" "∷")
                   `extQ` considerEqual @(HsLinearArrowTokens GhcPs)
               )
               x
               y
       | otherwise = Different []
 
+    isHsUniToken :: (Typeable a) => a -> Maybe TR.SomeTypeRep
+    isHsUniToken (typeOf -> rep)
+      | TR.SomeTypeRep (TR.App (TR.App t _) _) <- rep =
+          rep <$ TR.eqTypeRep t (TR.typeRep @HsUniToken)
+      | otherwise = Nothing
+
     considerEqualVia ::
       forall a.
       (Typeable a) =>
```
This is more complex, but would have the advantage that we don't have to add more cases here if `HsUniToken` is used for more many unicode variants in the future in the AST.

mrkkrp (approved):
I also applied renaming from #1008 here, since it looks like a good idea.

src/Ormolu/Diff/ParseResult.hs
mrkkrp: :+1: Let's keep this in mind in case we discover more of these. For now a simpler fix will do.

PR #1020, opened 2023-05-04
Allow to specify module re-exports in .ormolu files

### LINKED ISSUE #1017: Handle module re-exports correctly when inferring fixity of operators
After #994 we are now more precise about fixities, but this comes with a problem that some users may perceive as a regression, and they would be right. Now that we infer fixities based on import sections, we need to be precise about what operators are imported and from where. #994 largely handles that decently enough, except for one important case—module re-exports. The Hoogle database that we use to scrap operator/fixity declaration provides zero information about module re-exports, which means that when someone imports e.g. `Control.Lens`, they can't expect the same results as when they import `Control.Lens.Operators`. I think this is serious enough to delay the release of 0.7.0.0 for now.

mrkkrp:
Close #1017.

It turns out that gaining knowledge of all module re-exports in the ecosystem is tricky because the Hoogle database does not contain this information, so one would need to download all 14+k packages from Hackage and parse their source code. We've decided that perhaps hardcoding some common re-exports and otherwise giving the users a way to make Ormolu aware of arbitrary important re-exports is a satisfactory solution for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,12 +187,32 @@ infixl 4 <*>, <*, *>, <**>
 It uses exactly the same syntax as usual Haskell fixity declarations to make
 it easier for Haskellers to edit and maintain.
 
-Besides, all of the above-mentioned parameters can be controlled from the
+As of Ormolu 0.7.0.0, `.ormolu` files can also contain instructions about
+module re-exports that Ormolu should be aware of. This might be desirable
+because at the moment Ormolu cannot know about all possible module
+re-exports in the ecosystem and only few of them are actually important when
+it comes to fixity deduction. In 99% of cases the user won't have to do
+anything, especially since most common re-exports are already programmed
+into Ormolu. (You are welcome to open PRs to make Ormolu aware of more
+re-exports by default.) However, when the fixity of an operator is not
+inferred correctly, making Ormolu aware of a re-export may come in handy.
+Here is an example:
+
+```haskell
+module Control.Lens exports Control.Lens.At
+module Control.Lens exports Control.Lens.Lens
+```
+
+Module re-export declarations can be mixed freely with fixity overrides, as
+long as each declaration is on its own line.
+
+Finally, all of the above-mentioned parameters can be controlled from the
 command line:
 
 * Language extensions can be specified with the `-o` or `--ghc-opt` flag.
 * Dependencies can be specified with the `-p` or `--package` flag.
 * Fixities can be specified with the `-f` or `--fixity` flag.
+* Re-exports can be specified with the `-r` or `--reexport` flag.
 
 Searching for both `.cabal` and `.ormolu` files can be disabled by passing
 `--no-cabal`.
--- a/src/Ormolu/Fixity/Imports.hs
+++ b/src/Ormolu/Fixity/Imports.hs
@@ -6,10 +6,13 @@
 module Ormolu.Fixity.Imports
   ( FixityImport (..),
     extractFixityImports,
+    applyModuleReexports,
   )
 where
 
 import Data.Bifunctor (second)
+import Data.List.NonEmpty qualified as NE
+import Data.Map.Strict qualified as Map
 import Distribution.ModuleName (ModuleName)
 import Distribution.Types.PackageName
 import GHC.Data.FastString qualified as GHC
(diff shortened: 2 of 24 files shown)


weeder.dhall
```diff
@@ -2,6 +2,7 @@
   [ "^Main.main\$"
   , "^Paths_"
   , "^Ormolu.Terminal.QualifiedDo.>>\$" -- https://github.com/ocharles/weeder/issues/112
+  , "^Ormolu.Fixity.Parser.pPackageName"
```
mrkkrp, line 5: This one is used by `extract-hackage-info`.

weeder.dhall
```diff
@@ -2,6 +2,7 @@
...
   ]
-, type-class-roots = False
+, type-class-roots = True
```
mrkkrp, line 7: Necessary because some functions are only used to define `Arbitrary` instances (in tests).

README.md
```diff
@@ -187,12 +187,30 @@ infixl 4 <*>, <*, *>, <**>
 It uses exactly the same syntax as usual Haskell fixity declarations to make
 it easier for Haskellers to edit and maintain.
 
-Besides, all of the above-mentioned parameters can be controlled from the
+As of Ormolu 0.7.0.0, `.ormolu` files can also contain instructions about
+module re-exports that Ormolu should be aware of. This might be desirable
+because at the moment Ormolu cannot know about all possible module
+re-exports in the ecosystem and only few of them are actually important when
+it comes to fixity deduction. In 99% of cases the user won't have to do
+anything, especially that most common re-exports are already programmed into
```
amesgen, line 195: Do we want to explicitly encourage people to create issues when they encounter missing re-exports they think are useful for many users?

src/Ormolu/Fixity/Imports.hs
```diff
@@ -71,3 +74,18 @@ ieToOccNames = \case
...
+    expand i = do
+      case Map.lookup (fimportModule i) reexports of
+        Nothing -> pure i
+        Just exports ->
+          let exportToImport mmodule =
+                i
+                  { fimportPackage = Nothing,
+                    fimportModule = mmodule
+                  }
+           in NE.toList exports >>= expand . exportToImport
```
amesgen, line 91: Do we have/want a dedicated test that module re-exports of re-exports work as expected?

amesgen (approved):
Looks great!

---

I think the only other automatic thing we *could* do regarding re-exports are Cabal mixins:
```cabal
  mixins: 
      base hiding (Prelude)
    , relude (Relude as Prelude)
    , relude
```
I don't think this would be too hard, but OTOH, I think they are not used that pervasively, such that it should be easy to specify them in `.ormolu` files. In any case, it does not have to happen in this PR.

weeder.dhall
mrkkrp: Applied!

src/Ormolu/Fixity/Internal.hs
```diff
@@ -156,6 +160,48 @@ newtype FixityOverrides = FixityOverrides
...
+
+-- | Module re-exports
+newtype ModuleReexports = ModuleReexports
+  { unModuleReexports :: Map ModuleName (NonEmpty ModuleName)
+  }
+  deriving stock (Eq, Show)
+
+-- | Module re-exports to apply by default.
+defaultModuleReexports :: ModuleReexports
+defaultModuleReexports =
```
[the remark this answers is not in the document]
mrkkrp: Added re-exports from `servant` and `optics`.

src/Ormolu/Fixity/Imports.hs
mrkkrp: Indeed we have! https://github.com/tweag/ormolu/blob/mkarpov%40issue-1017/tests/Ormolu/FixitySpec.hs#L232-L246

README.md
mrkkrp: I added a sentence to encourage such contributions :slightly_smiling_face:

mrkkrp:
@amesgen I've addressed your comments. Yeah, I think we can release 0.7.0.0 without support for mixins. This can be added later.

amesgen (approved):

PR #1022, opened 2023-05-14
Relax upper bound to accommodate dependency on `ansi-terminal-1.0`

mpilgrem:
`ansi-terminal-1.0` is released on Hackage.

See also https://github.com/commercialhaskell/stackage/issues/6976.

Tested by building with Stack >= 2.9.3 and `stack.yaml` with these additions:
~~~yaml
# resolver: lts-20.11 # GHC 9.2.5
resolver: lts-20.20 # GHC 9.2.7

extra-deps:
- ansi-terminal-1.0
- ansi-terminal-types-0.11.5

allow-newer: true
allow-newer-deps:
- ansi-wl-pprint
~~~

### DIFF
--- a/ormolu.cabal
+++ b/ormolu.cabal
@@ -100,7 +100,7 @@ library
         Cabal-syntax >=3.10 && <3.11,
         Diff >=0.4 && <1.0,
         MemoTrie >=0.6 && <0.7,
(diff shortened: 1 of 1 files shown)


mrkkrp:
Thanks! I'm going to publish a revision on Hackage too.

PR #1033, opened 2023-05-20
Ensure source file not mentioned in cabal file has same behavior as cabal file not existing

### LINKED ISSUE #1032: Format base infix operators better when cabal file doesn't mention module
**Describe the bug**
If you have a standalone Haskell file in a Cabal project (e.g. a Haskell script), Ormolu will find the cabal file, but it'll see that the cabal file doesn't mention the file and return `ciDependencies = []`.

**To Reproduce**
Make a file `test.hs` with the following contents in a directory containing a `.cabal` file:
```hs
foo =
  unlines $
    hello <> world : ["Line " ++ show x | x <- [10 .. 20]]
  where
    hello = "hello"
    world = " world"
```

It will reformat as
```hs
foo =
  unlines
    $ hello
    <> world
    : ["Line " ++ show x | x <- [10 .. 20]]
  where
    hello = "hello"
    world = " world"
```

**Expected behavior**
Formatting a Haskell file with a cabal file that doesn't mention it should be equivalent to formatting without a cabal file

**Environment**
 - OS name + version:
 - Version of the code: 0.7.0.0

**Additional context**

brandonchinn178:
Resolves #1032

Technically, only the first commit is necessary to solve the issue. But I refactored the `CabalInfo` subsystem to ensure that the former `CabalFileNotFound` + `CabalFileDidNotMention` branches used the same logic. Specifically, this change allows `refineConfig` to be the sole source of truth for defining the Cabal information for both a cabal file not existing and a cabal file not mentioning a source file (previously, this was two different locations: `parseCabalInfo` was doing it for `CabalFileDidNotMention` and `refineConfig` was doing it for `CabalFileNotFound`).

## Test

Add the following Haskell file:
```hs
foo =
  unlines $
    hello <> world : ["Line " ++ show x | x <- [10 .. 20]]
  where
    hello = "hello"
    world = " world"
```
Running `ormolu` on it formats the same in both cases:
1. The Haskell file is in an empty directory
2. The Haskell file is in a directory with a Cabal file, but it's not mentioned in the Cabal file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+* Include `base` fixity information when formatting a Haskell file that's
+  not mentioned in an existing cabal file. [Issue
+  1032](https://github.com/tweag/ormolu/issues/1032)
+
 ## Ormolu 0.7.0.0
 
 * Inference of operator fixity information is now more precise and takes
--- a/src/Ormolu/Utils/Cabal.hs
+++ b/src/Ormolu/Utils/Cabal.hs
@@ -28,6 +28,7 @@ import Distribution.Utils.Path (getSymbolicPath)
 import Language.Haskell.Extension
 import Ormolu.Config
 import Ormolu.Exception
+import Ormolu.Fixity
 import Ormolu.Utils.IO (findClosestFileSatisfying)
 import System.Directory
 import System.FilePath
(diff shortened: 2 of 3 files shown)


mrkkrp (approved):

mrkkrp:
Thanks! I merged only the first commit though because I'm not sure about the benefits vs mental cost of the others.

brandonchinn178:
@mrkkrp Thanks! I do think the refactor is an improvement for overall understandability. e.g. it's a bit confusing IMO for CabalInfo to be different for the same cabal file but different source files. Do you mind if I open a separate PR to discuss that?

PR #1034, opened 2023-05-23
Ensure "source file not mentioned in Cabal file" has same behavior as "no Cabal file found"

brandonchinn178:
Broken off from #1033 

This change has the following improvements:
1. Force user to handle missing stanza information (e.g. dependencies) when the cabal file does not mention a source file
    * This would make it much harder to have the same regression as #1032 somewhere else
3. Make `CabalInfo` return all info for the cabal file, without reference to the source file
    * Simplifies `parseCabalInfo` as just "parse cabal file", not "parse cabal file + extract info relevant to given source file"
    * IMO this is clearer, given the names of the types and functions.


mrkkrp:
I'm happy with the first commit, which I adjusted it a bit to apply the new helper to both caching of cabal files and `.ormolu` files, however I don't think we can easily merge something that changes the signature of `refineConfig` because it changes the API, which we are supposed to keep backwards compatible now as per c81320d8a1069674c17fa5ee52b64c18604f84de.

brandonchinn178:
Ah sure, I could make the changes backwards compatible. Give me a second

brandonchinn178:
@mrkkrp how's that?

mrkkrp:
Eh, I like splitting the code in two branches (old and new) even less. Let's apply d91bbd064511fee913bd348fc413ba5a820e8927 and keep the rest on the shelf for a while until we decide to release 0.9.0.0 at which point we can make a clean break.

PR #1035, opened 2023-05-24
Allow optparse-applicative-0.18

mrkkrp:


amesgen (approved):

amesgen (changes requested):
My approval was premature, can you also update it for `extract-hackage-info`?

PR #1036, opened 2023-05-24
Instruct Ormolu about a common hspec re-export

mrkkrp:


tests/Ormolu/PrinterSpec.hs
```diff
@@ -49,6 +49,8 @@ checkExample srcPath' = it (fromRelFile srcPath' ++ " works") . withNiceExceptio
               Set.fromList
                 [ "base",
                   "esqueleto",
+                  "hspec",
+                  "hspec-expectations",
```
amesgen, line 53: Just to double-check: The `hspec-expectations` would not be necessary with https://github.com/tweag/ormolu/issues/1037, right?

amesgen (approved):

tests/Ormolu/PrinterSpec.hs
mrkkrp: Yes, that's correct.

PR #1042, opened 2023-06-06
Erase lines affected by CPP before attempting to parse pragmas and imports

### LINKED ISSUE #1040: Cannot parse module with only conditional CPP block
**Describe the bug**
The GHC parser (in Haddock mode) and by extension Ormolu cannot parse modules that _only_ contain a conditional CPP block. This is possibly an upstream issue with `ghc-lib-parser`, but I'm filing it here because I'm unsure as GHC compiles the file correctly.

**To Reproduce**
Write the following to `Bug.hs`:
```hs
{-# LANGUAGE CPP #-}

module Bug where

#ifdef flag
constant :: Int
constant = 1312
#endif
```
Then run:
```bash
ormolu Bug.hs
```
Ormolu fails with:
```
Bug.hs:1:1
  {ErrorWithoutFlag
   lexical error in pragma at character 'i'}
```

**Expected behavior**
Ormolu outputs the file unchanged.

**Environment**
 - macOS Ventura
 - ormolu 0.7.0.0
 - ghc-lib-parser 9.6.2.20230523

mrkkrp:
Close #1040.


src/Ormolu/Processing/Cpp.hs
```diff
@@ -60,3 +61,14 @@ cppLines input = IntSet.fromAscList $ go Outside (T.lines input `zip` [1 ..])
             inConditional = case state of
               InConditional {} -> True
               _ -> False
+
+-- | Replace all lines affected by CPP with blank lines.
+eraseCppLines :: Text -> Text
+eraseCppLines input =
+  T.intercalate "\n" . fmap eraseCpp $ T.lines input `zip` [1 ..]
```
amesgen, line 68: Nit: What about `T.unlines` instead of `T.intercalate "\n"` for symmetry/consistency?

amesgen (approved):

src/Ormolu/Processing/Cpp.hs
mrkkrp: Good idea, done!

PR #1049, opened 2023-06-27
Avoid building ghc-lib-parser if GHC 9.6 is already being used

brandonchinn178:
This speeds up a clean build on GHC 9.6 by a ton


amesgen:
See #948 for previous discussion on this change

PR #1050, opened 2023-06-27
Update stack snapshot

brandonchinn178:


amesgen (approved):
Thanks, we need a Stack bump for #1056 in any case :+1:

PR #1056, opened 2023-07-30
ghc-lib-parser 9.8

amesgen:


data/examples/declaration/signature/fixity/infix.hs
```diff
@@ -1,2 +1,4 @@
 infix 0 <?>
 infix 9 <^-^>
+
+infix 2 ->
```
mrkkrp, line 4: Was it previously impossible to specify a custom fixity for `->`?

mrkkrp (approved):
Thanks a lot for performing the migration (again)!

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Unreleased
+
+* Switched to `ghc-lib-parser-9.8`, with the following new syntactic features:
+  * `ExtendedLiterals`: `123#Int8` is a literal of type `Int8#`. (disabled by
+    default)
+  * `TypeAbstractions`: `@k`-binders in data type declarations (enabled by
+    default)
+  * GHC proposal #134: deprecating/warning about exports
```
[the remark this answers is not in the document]
amesgen: Done :+1:

data/examples/declaration/signature/fixity/infix.hs
amesgen: Yes and no: `infix 2 ->` could be parsed also by older GHCs, but all GHCs later reject it ("`Illegal fixity signature of built-in syntax: ->`"). I only included this as I had to add this new case in `p_rdrName`
```haskell
          -- whether the `->` identifier is parenthesized
          NameAnnRArrow {nann_mopen = Just _} -> parens N
```
and I wondered when it is possible for `->` (as an identifier, so not as part of a normal function arrow `A -> B`, which is parsed to `HsFunTy`) to *not* be parenthesized, and a fixity declaration was the only thing I could came up with, so I added a test for it right here.

PR #1063, opened 2023-09-01
Fix interaction of operators starting with `#` and `UnboxedSums`

### LINKED ISSUE #1062: Cannot format operators starting with a hash when using UnboxedSums or UnboxedTuples
**Describe the bug**
Ormolu formats `( #<| )` as `(#<|)` when `-XUnboxedSums` or `-XUnboxedTuples` is set, thus the formatted code generates a parser error. I think it applies to any operator starting with `#`.

**To Reproduce**

Format the following file with `--unsafe` (else it fails).

```haskell
{-# LANGUAGE UnboxedTuples #-}

module Foo (( #<| )) where

( #<| ) :: Int -> Int -> Int
( #<| ) = (+)
```

The generated code does not compile.

**Expected behavior**
Ormolu should keep the spaces to be valid code, i.e.  `( #<| )` should be unchanged.

**Environment**
 - OS name + version: Linux (openSUSE Tumbleweed 20230828)
 - Version of the code: ormolu 0.7.1.0 using ghc-lib-parser 9.6.2.20230523

amesgen:
Closes #1062 

From https://downloads.haskell.org/ghc/9.6.2/docs/users_guide/exts/primitives.html:

> Note that when unboxed tuples are enabled, `(#` is a single lexeme, so for example when using operators like `#` and `#-` you need to write `( # )` and `( #- )` rather than `(#)` and `(#-)`.

It turns that this is already the case for `UnboxedSums` (which is implied by `UnboxedTuples`).

### DIFF
--- a/src/Ormolu/Printer/Meat/Common.hs
+++ b/src/Ormolu/Printer/Meat/Common.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ViewPatterns #-}
 
 -- | Rendering of commonly useful bits.
 module Ormolu.Printer.Meat.Common
@@ -20,8 +21,9 @@ import Data.Text qualified as T
 import GHC.Hs.Doc
 import GHC.Hs.Extension (GhcPs)
 import GHC.Hs.ImpExp
+import GHC.LanguageExtensions.Type (Extension (..))
 import GHC.Parser.Annotation
-import GHC.Types.Name.Occurrence (OccName (..))
+import GHC.Types.Name.Occurrence (OccName (..), occNameString)
 import GHC.Types.Name.Reader
 import GHC.Types.SourceText
 import GHC.Types.SrcLoc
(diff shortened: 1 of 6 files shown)


src/Ormolu/Printer/Meat/Common.hs
```diff
@@ -62,15 +64,27 @@ p_ieWrappedName = \case
...
           NameAnnOnly {nann_adornment = NameParensHash} -> const $ txt "(# #)"
           _ -> id
         EpAnnNotUsed -> id
+
+      -- When UnboxedSums is enabled, `(#` is a single lexeme, so we have to
+      -- insert spaces when we have a parenthesized operator starting with `#`.
+      handleUnboxedSumsAndHashInteraction
+        | unboxedSums,
+          -- Qualified names do not start wth a `#`.
+          Unqual (occNameString -> '#' : _) <- x =
```
amesgen, line 84: I am not 100% sure if the `Orig` or `Exact` constructors of [`RdrName`](https://hackage.haskell.org/package/ghc-lib-parser-9.6.2.20230523/docs/GHC-Types-Name-Reader.html#t:RdrName) could ever be relevant here :thinking:

mrkkrp (approved):

PR #1068, opened 2023-09-05
Fix AST diffing for empty Haddock comments in data decls

### LINKED ISSUE #1065: Differing AST on empty {- ^ -} Haddock markup
**Describe the bug**

If I have an empty `{- ^ -}` Haddock annotation then `ormolu` fails with an `AST of input and AST of formatted code differ` error.

**To Reproduce**

```
$ cat /tmp/example.hs
data T = T {- ^ -}
$ ormolu /tmp/example.hs
/tmp/example.hs
@@ -1,1 +1,3 @@
- data T = T {- ^ -}
+ data T
+   = -- |
+     T

    at /tmp/example.hs:1:10
  To format anyway, use --unsafe.
```

**Expected behavior**

`ormolu` to produce the same reformatted code as above but to _not_ complain that the ASTs differ (I don't think they do!). `{- ^ ... -}` is not mentioned in [the Haddock documentation](https://haskell-haddock.readthedocs.io/en/latest/markup.html#constructors-and-record-fields) as valid Haddock syntax, but it does accept it. Note that if there is text in the markup then everything is fine:

```
$ cat /tmp/example2.hs
data T = T {- ^ text -}
$ ormolu /tmp/example2.hs
data T
  = -- | text
    T
```

**Environment**

 - OS name + version:  Debian GNU/Linux 11.7
 - Version of the code: 

```
$ ormolu --version
ormolu 0.7.1.0
using ghc-lib-parser 9.6.2.20230523
```

**Additional context**

`ormolu` is great, thanks!

Perhaps related: https://github.com/tweag/ormolu/issues/822

amesgen:
Closes #1065, by basically amending #818 with another case where Haddock comments can occur

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * Format parenthesized operators starting with a `#` correctly in the presence
   of `UnboxedSums`. [Issue 1062](https://github.com/tweag/ormolu/issues/1062).
 
+* Fix false positives in AST diffing related to empty Haddock comments in data
+  declarations. [Issue 1065](https://github.com/tweag/ormolu/issues/1065).
+
 ## Ormolu 0.7.1.0
 
 * Include `base` fixity information when formatting a Haskell file that's
--- a/data/examples/other/empty-haddock-out.hs
+++ b/data/examples/other/empty-haddock-out.hs
@@ -5,3 +5,5 @@ where
 
 test ::
   test
+
+data T = T
--- a/data/examples/other/empty-haddock.hs
+++ b/data/examples/other/empty-haddock.hs
@@ -9,3 +9,5 @@ where
 test ::
   -- |
   test
+
+data T = T {- ^ -}
--- a/src/Ormolu/Parser.hs
+++ b/src/Ormolu/Parser.hs
@@ -177,7 +177,7 @@ parseModuleSnippet Config {..} modFixityMap dynFlags path rawInput = liftIO $ do
 normalizeModule :: HsModule GhcPs -> HsModule GhcPs
 normalizeModule hsmod =
   everywhere
-    (extT (mkT dropBlankTypeHaddocks) patchContext)
+    (mkT dropBlankTypeHaddocks `extT` dropBlankDataDeclHaddocks `extT` patchContext)
     hsmod
       { hsmodImports =
           normalizeImports (hsmodImports hsmod),
@@ -204,6 +204,13 @@ normalizeModule hsmod =
       L _ (HsDocTy _ ty s) :: LHsType GhcPs
         | isBlankDocString s -> ty
       a -> a
+    dropBlankDataDeclHaddocks = \case
+      ConDeclGADT {con_doc = Just s, ..} :: ConDecl GhcPs
+        | isBlankDocString s -> ConDeclGADT {con_doc = Nothing, ..}
+      ConDeclH98 {con_doc = Just s, ..} :: ConDecl GhcPs
+        | isBlankDocString s -> ConDeclH98 {con_doc = Nothing, ..}
+      a -> a
+
     patchContext :: LHsContext GhcPs -> LHsContext GhcPs
     patchContext = fmap $ \case
       [x@(L _ (HsParTy _ _))] -> [x]


tomjaguarpaw:
Oh no! This is very unfortunate, and so is https://github.com/tweag/ormolu/pull/818. Empty Haddock comments are semantically meaningful. They cause the Haddock output to wrap.

For further information see

* http://h2.jaguarpaw.co.uk/posts/improving-the-typed-process-documentation/ under "Forced type signatures to wrap"
* [The source](https://hackage.haskell.org/package/typed-process-0.2.10.1/docs/src/System.Process.Typed.html#setStdin)
* [The Haddock output](https://hackage.haskell.org/package/typed-process-0.2.10.1/docs/System-Process-Typed.html#v:setStdin)

Ormolu obliterates this nice formatting :(  Granted, using `-- ^`/`-- |` to force Haddock to wrap signatures is a hack, but there is no other way (I believe).

amesgen:
Interesting, that is a very nice trick; but it doesn't seem to work anymore :sob:: You link to `setStdin` from typed-protocols 0.2.10.1, but in 0.2.11.0 (the current version), this trick doesn't seem to have any effect anymore (which is consistent that it does not show up as a Haddock comment in the AST anymore; how GHC associates Haddock comments to the AST was changed in GHC 9.0):

 - [The source](https://hackage.haskell.org/package/typed-process-0.2.11.0/docs/src/System.Process.Typed.Internal.html#setStdin)
 - [The Haddock output](https://hackage.haskell.org/package/typed-process-0.2.11.0/docs/System-Process-Typed.html#v:setStdin)

Maybe this should be considered to be an upstream GHC issue? If empty Haddock comments appear in the AST again, it will also be easier for Ormolu to preserve them. WDYT?

amesgen:
I can also confirm locally that your trick works on GHC 8.10, but does no longer work since 9.0.

tomjaguarpaw:
Thanks for a very speedy reply! Oh, yes, I see. I didn't realise that this had changed, and I didn't realise I hadn't checked the latest documentation. I guess the two versions were generated by the builder at times straddling the change to GHC 9.x for Haddock generation.

> Maybe this should be considered to be an upstream GHC issue? If empty Haddock comments appear in the AST again, it will also be easier for Ormolu to preserve them. WDYT?

I suppose that seems reasonable but I'd still prefer to have

* `-- |` preserved
* `-- ^` changed to `-- |`
* only `{- ^ -}` deleted

even though that won't get me any benefit under GHC 9.x (yet ...). Anyway, it's your application (which I enjoy using a lot) and your effort, so please do what you think is most appropriate.

In fact, now that I come to think of it, I don't understand why formatting `data T = T {- ^ -}` to `data T = T` passes the AST equivalency check anyway ...

amesgen:
> > Maybe this should be considered to be an upstream GHC issue? If empty Haddock comments appear in the AST again, it will also be easier for Ormolu to preserve them. WDYT?
> 
> I suppose that seems reasonable but I'd still prefer to have
> 
>     * `-- |` preserved
> 
>     * `-- ^` changed to `-- |`
> 
>     * only `{- ^ -}` deleted
> 
> 
> even though that won't get me any benefit under GHC 9.x (yet ...).

I actually like your suggested behavior (though I would probably not treat `{- ^ -}` specially and still turn it into `-- |`); it is just annoying to implement as `-- |` does not appear in the AST as an Haddock comment, only as a "regular" comment in an EPA extension point, so we would have to manually look in those to find them and treat them like a proper Haddock comment. I have not investigated this in detail, but a priori, the implementation seems too fiddly in relation to the relatively low severity of this stylistic concern.

I think the proper fix is to actually change the GHC parser behavior as mentioned above, I opened https://gitlab.haskell.org/ghc/ghc/-/issues/23935 for that. I think that would also re-enable your Haddock trick, I tagged you there (hope that is fine).

> In fact, now that I come to think of it, I don't understand why formatting `data T = T {- ^ -}` to `data T = T` passes the AST equivalency check anyway ...

We preprocess the AST to remove Haddock comments that are all blank (this PR is extending this preprocessing step).

---

I will open a new issue here about preserving blank Haddock comments (as suggested by you) if the reaction in the GHC ticket is positive; once we upgrade to a GHC that has the fix, we can remove the code that eg this PR is touching. In the meantime, I think that this PR is still an incremental improvement (it certainly doesn't make anything worse).

tomjaguarpaw:
> I would probably not treat `{- ^ -}` specially and still turn it into `-- |`

I think that would be even better if it fits with the constraints. I still don't fully understand the behaviour of the GHC parser and how that impinges.

> I tagged you there (hope that is fine).

Yes, absolutely, thanks for doing that.

mrkkrp (approved):

PR #1069, opened 2023-09-11
Warn when fixities are different from base

### LINKED ISSUE #1060: Make it easier to debug fixity issues
In https://github.com/fourmolu/fourmolu/issues/357#issuecomment-1691915590, it took a bit before realizing that NoImplicitPrelude is causing Prelude operators from getting the correct fixity. Is there some way we can make debugging this easier, e.g. when `--debug` is turned on? Perhaps when `--debug` is turned on, show warnings when operators from base are detected with different fixity info? I think base is an important enough package to special case this logic. Or perhaps show the warning if any operators have different fixity than in the Hackage fixity map? Since this would only show up with `--debug`, I think being a bit noisy with false positives is better than otherwise

brandonchinn178:
Resolves #1060 

Tested with:
```hs
{-# LANGUAGE NoImplicitPrelude #-}

import MyCustomPrelude

main :: IO ()
main =
  putStrLn $
    "hello " ++ "world"
```
This shows the following with `--debug`:
```
*** WARNING *** Operator is possibly using the wrong fixity. Got: FixityApproximation {faDirection = Just InfixL, faMinPrecedence = 9, faMaxPrecedence = 9}, Fixities being shadowed: [FixityInfo {fiDirection = InfixR, fiPrecedence = 5},FixityInfo {fiDirection = InfixR, fiPrecedence = 5},FixityInfo {fiDirection = InfixR, fiPrecedence = 5},FixityInfo {fiDirection = InfixR, fiPrecedence = 5},FixityInfo {fiDirection = InfixR, fiPrecedence = 5}]
*** WARNING *** Operator is possibly using the wrong fixity. Got: FixityApproximation {faDirection = Just InfixL, faMinPrecedence = 9, faMaxPrecedence = 9}, Fixities being shadowed: [FixityInfo {fiDirection = InfixR, fiPrecedence = 0},FixityInfo {fiDirection = InfixR, fiPrecedence = 0},FixityInfo {fiDirection = InfixR, fiPrecedence = 0}]
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -40,6 +40,7 @@ import System.IO (stderr)
 main :: IO ()
 main = do
   Opts {..} <- execParser optsParserInfo
+  initializeLogging optConfig
   let formatOne' =
         formatOne
           optConfigFileOpts
@@ -85,11 +86,11 @@ formatOne ConfigFileOpts {..} mode reqSourceType rawConfig mpath =
           cabalSearchResult <- getCabalInfoForSourceFile sourceFile
           case cabalSearchResult of
             CabalNotFound -> do
-              logDebug rawConfig "CABAL FILE" $ "Could not find a .cabal file for " <> sourceFile
+              logDebugM "CABAL FILE" $ "Could not find a .cabal file for " <> sourceFile
               return Nothing
             CabalDidNotMention cabalInfo -> do
               relativeCabalFile <- makeRelativeToCurrentDirectory (ciCabalFilePath cabalInfo)
-              logDebug rawConfig "CABAL FILE" $
+              logDebugM "CABAL FILE" $
                 "Found .cabal file "
                   <> relativeCabalFile
                   <> ", but it did not mention "
@@ -116,7 +117,7 @@ formatOne ConfigFileOpts {..} mode reqSourceType rawConfig mpath =
             ormoluStdin config >>= TIO.putStr
             return ExitSuccess
           InPlace -> do
-            logError "In place editing is not supported when input comes from stdin."
+            logErrorM "In place editing is not supported when input comes from stdin."
             -- 101 is different from all the other exit codes we already use.
             return (ExitFailure 101)
           Check -> do
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -24,6 +24,7 @@ import Options.Applicative
 import Ormolu
 import Ormolu.Diff.Text (diffText, printTextDiff)
 import Ormolu.Fixity
+import Ormolu.Logging
 import Ormolu.Parser (manualExts)
 import Ormolu.Terminal
 import Ormolu.Utils (showOutputable)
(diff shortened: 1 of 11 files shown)


app/Main.hs
```diff
@@ -82,22 +84,17 @@ formatOne ConfigFileOpts {..} mode reqSourceType rawConfig mpath =
...
-                  "Found .cabal file "
-                    <> relativeCabalFile
-                    <> ", but it did not mention "
-                    <> sourceFile
+              relativeCabalFile <- makeRelativeToCurrentDirectory (ciCabalFilePath cabalInfo)
+              logDebugM "CABAL FILE" $
+                "Found .cabal file "
+                  <> relativeCabalFile
+                  <> ", but it did not mention "
+                  <> sourceFile
```
brandonchinn178, line 97: This is technically different behavior than the previous version: if debug is not enabled, the new version still relativizes the cabal file path, even though it's not printed.

I could create a version of `logDebugM` that takes a to-be-computed `IO String` message, but I don't think adding such a bespoke version is worth this relatively cheap call.

src/Ormolu/Logging.hs
```diff
@@ -0,0 +1,84 @@
...
+loggerConfig :: IORef LoggerConfig
+loggerConfig = unsafePerformIO $ newIORef (error "Logger not configured yet")
+{-# NOINLINE loggerConfig #-}
+
+initializeLogging :: Config region -> IO ()
+initializeLogging cfg =
+  writeIORef loggerConfig $
+    LoggerConfig
+      { debugEnabled = cfgDebug cfg
+      }
```
brandonchinn178, line 31: Yeah, this is a bit hacky and unprincipled, smuggling the config through the global environment. But I figured this was much neater than storing the config in `R`'s environment and passing it through `addFixityInfo`.

I call this function immediately in all the entrypoints (`main`, `ormolu`, and for tests), so we should be good in principle. The only issue might be if someone imports from Ormolu's submodules + calls functions directly. I guess I could set a default logger config instead of erroring with "logger not configured yet".

mrkkrp:
Could you add such a warning without introducing `Ormolu.Logging`?

brandonchinn178:
So without a logging framework, we would have to do the following:
1. Add debug/quiet/log level to the environment of R
2. Pass the log level through from main to where R is set up
3. Pass log level to reassociateOpTree
4. Pass log level to addFixityInfo
5. Call traceM in addFixityInfo

This seemed rather annoying and not scalable, and I figured it'd be better
to actually come up with a logging system that works in pure code.

I assume the biggest issue with this solution is the global ref? I think
having a proper logging system would be better than the smattering of
trace/putStrLn statements currently in the codebase. If the concern is with
the global ref (which I agree is a bit code smelly), maybe I could store a
LogContext in R and have the log functions take in the context?

Or maybe turn addFixityInfo into a StateT that outputs fixities different
from base and print out the warnings when it gets back into the R monad?

brandonchinn178:
@mrkkrp do you have any thoughts on the above approaches?

PR #1071, opened 2023-10-23
Improve ormolu live wasm size

TerrorJack:
This PR does even more aggressive cleanup of wasm module before snapshotting, let's see if it helps in further shrinking the wasm module size :)

### DIFF
--- a/ormolu-live/cbits/init.c
+++ b/ormolu-live/cbits/init.c
@@ -1,9 +1,21 @@
+#include "Rts.h"
+
 #include "Main_stub.h"
-#include <Rts.h>
+
+void malloc_inspect_all(void (*handler)(void *start, void *end,
+                                        size_t used_bytes, void *callback_arg),
+                        void *arg);
+
+static void malloc_inspect_all_handler(void *start, void *end,
+                                       size_t used_bytes, void *callback_arg) {
+  if (used_bytes == 0) {
+    memset(start, 0, (size_t)end - (size_t)start);
+  }
+}
 
 __attribute__((export_name("wizer.initialize"))) void __wizer_initialize(void) {
   char *args[] = {
-      "ormolu-live.wasm", "+RTS", "--nonmoving-gc", "-H64m", "-RTS", NULL};
+      "ormolu-live.wasm", "+RTS", "-H64m", "-RTS", NULL};
   int argc = sizeof(args) / sizeof(args[0]) - 1;
   char **argv = args;
   hs_init_with_rtsopts(&argc, &argv);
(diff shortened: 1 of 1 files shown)


ormolu-live/cbits/init.c
```diff
@@ -1,14 +1,28 @@
...
+      "ormolu-live.wasm", "+RTS", "-H64m", "-RTS", NULL};
   int argc = sizeof(args) / sizeof(args[0]) - 1;
   char **argv = args;
   hs_init_with_rtsopts(&argc, &argv);
   evaluateFixityInfo();
   hs_perform_gc();
   hs_perform_gc();
   rts_clearMemory();
+  malloc_inspect_all(malloc_inspect_all_handler, NULL);
+  memset(0, 0, (size_t)__builtin_frame_address(0));
```
amesgen, line 27: Is this line necessary for correctness, or is it another optimization? Size-wise, it seems to cause a reduction by ~4KB from 18.596MB to 18.592MB.

ormolu-live/cbits/init.c
```diff
@@ -1,14 +1,28 @@
...
+void malloc_inspect_all(void (*handler)(void *start, void *end,
+                                        size_t used_bytes, void *callback_arg),
+                        void *arg);
+
+static void malloc_inspect_all_handler(void *start, void *end,
+                                       size_t used_bytes, void *callback_arg) {
+  if (used_bytes == 0) {
+    memset(start, 0, (size_t)end - (size_t)start);
+  }
+}
```
amesgen, line 14: Maybe add a link to eg https://github.com/WebAssembly/wasi-libc/blob/be1704a9568e98ae073f074f7271cae68c23d161/dlmalloc/src/malloc.c#L1027-L1052 (or is there a better reference)? But I guess you might add this trick to [ghc-wasm-meta](https://gitlab.haskell.org/ghc/ghc-wasm-meta/), in which case it doesn't need to be explained here.

amesgen (commented):
Thanks, I can confirm the size reduction in the uncompressed case; though interestingly, with `brotli -5`, I actually get an increase from 4.4MB to 4.6MB :thinking: For zstd (with default options), it is an increase from 5.4MB to 5.5MB.

ormolu-live/cbits/init.c
TerrorJack: You mean the `memset` line? It's meant to wipe the stack region which starts from `0x0` (because ghc is configured to link with `-Wl,--stack-first`) up to the current stack pointer. It's an independent optimization, suitable to be done just before exiting the wizer initialize function.

ormolu-live/cbits/init.c
TerrorJack: Yes, I'll add this to `ghc-wasm-meta` documentation with more explanation. `malloc_inspect_all` is a `dlmalloc` internal function to traverse the entire heap, not exposed by upstream `wasi-libc`, but I enabled it in our fork for this purpose.

amesgen (approved):
I don't think the compression weirdness mentioned in https://github.com/tweag/ormolu/pull/1071#pullrequestreview-1700064205 has to block merging this.

PR #1072, opened 2023-10-23
Correct dependency bounds

mrkkrp:
Inspired by https://github.com/mrkkrp/megaparsec/pull/548, which see.

### DIFF
--- a/extract-hackage-info/extract-hackage-info.cabal
+++ b/extract-hackage-info/extract-hackage-info.cabal
@@ -12,14 +12,14 @@ executable extract-hackage-info
     ghc-options:      -O2 -Wall -rtsopts -Wunused-packages
     build-depends:
         Cabal-syntax >=3.10 && <3.11,
-        base >=4.12 && <5.0,
+        base >=4.12 && <5,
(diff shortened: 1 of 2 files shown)


extract-hackage-info/extract-hackage-info.cabal
```diff
@@ -12,14 +12,14 @@ executable extract-hackage-info
     ghc-options:      -O2 -Wall -rtsopts -Wunused-packages
     build-depends:
         Cabal-syntax >=3.10 && <3.11,
-        base >=4.12 && <5.0,
+        base >=4.12 && <5,
         binary >=0.8 && <0.9,
         bytestring >=0.10 && <0.13,
         containers >=0.6 && <0.8,
         directory >=1.0 && <2.0,
```
amesgen, line 19:
```suggestion
        directory >=1.0 && <2,
```

ormolu.cabal
amesgen, line 1: There is still is an occurrence of `base >=4.14 && <5.0` here.

amesgen (approved):
Thanks, I think I spotted two more such cases, otherwise, LGTM

PR #1073, opened 2023-10-24
Add hie.yaml to .gitignore

mrkkrp:
This file is typically automatically derived with [`implicit-hie`](https://hackage.haskell.org/package/implicit-hie).


amesgen (approved):
Looks good, personally, I never had to generate a `hie.yaml` for Ormolu as HLS's default detected configuration never caused any problems for me.

PR #1092, opened 2024-01-21
Fix comments in if blocks

### LINKED ISSUE #998: Comment placement on then/else branches
Hi, I'm not sure if the AST would allow us to do this, but it would be great if this formatting of comment placement was preserved by ormolu:

```haskell
if something
  -- then comment
  then do
    stuff
    stuff
  -- else comment
  else do
    stuff
    stuff
```

When formatted, this looks like

```haskell
if something
  then -- then comment
  do
    stuff
    stuff
  else -- else comment
  do
    stuff
    stuff
```

One workaround is to place comments at the top of each alternative, which is okay, but seems a little less natural.

```haskell
if something
  then do
    -- then comment
    stuff
    stuff
  else do
    -- else comment
    stuff
    stuff
```

Thanks!

brandonchinn178:
Resolves #998 

The issue has two bugs associated with it, which are both solved in this PR.

## Comments between then/else tokens and body

This bug happens if there's a comment between the `then`/`else` tokens and the body of the branch:
```hs
-- without comments
if cond
  then case a of
    Just b -> ...
  else
    foo
      True
      ""

-- expected
if cond
  then
    -- comment
    -- comment
    case a of
      Just b -> ...
  else
    -- comment
    -- comment
    foo
      True
      ""

-- previous behavior
if cond
  then -- comment
  -- comment
  case a of
    Just b -> ...
  else -- comment
  -- comment

    foo
      True
      ""
```

There are a few bugs here:
1. The comment is moved up next to the keyword instead of staying on the next line
2. Subsequent comments are not indented
3. Expressions that normally "hang" like `case` are not indented
4. Expressions that don't hang like the second branch here get an extra line added

The issue is that the code does the following steps:
1. Print `then`/`else` token
2. `located` A: Spit out comments
3. `located` B: Set breakpoint layout
4. `placeHanging` Apply hanging format
5. Print body

But we need to spit out the comments AFTER applying the hanging layout. Additionally, we need to never hang when there are comments. So I fixed this by moving "Apply hanging format" before "Spit out comments", and manually setting the breakpoint layout before:
1. Print `then`/`else` token
2. **`switchLayout`: Set breakpoint layout**
3. **`placeHanging` Apply hanging format**
3. `located` A: Spit out comments
4. `located` B: Set breakpoint layout
4. ~`placeHanging` Apply hanging format~
5. Print body

## Comments before then/else keywords

In this bug, comments before the `then`/`else` keywords get moved to after:
```hs
-- input + expected
if cond
  -- asdf
  then foo
  -- asdf
  then bar

-- previous behavior
if cond
  then -- asdf
    foo
  else -- asdf
    bar
```

This happens because we only spit out comments when we call `located`, but we dont call `located` for keywords; we just do a plain `txt "then"`.

The fix for this was relatively easy: I just wrap `txt "then"` in a `located` containing the SrcSpan for the `then` keyword from the annotations. I think there's future work here to always do this for all keywords, now that GHC has prettyprinting information with SrcSpans for all keywords


src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -971,28 +971,56 @@ p_if ::
...
+    locatedToken mTokenSpan token =
+      case mTokenSpan of
+        Just tokenSpan -> located (L tokenSpan ()) $ \_ -> txt token
+        Nothing -> txt token
+
+    placeHangingLocated mTokenSpan bodyLoc@(L _ body) = do
+      let hasComments = fromMaybe False $ do
+            tokenSpan <- mTokenSpan
+            bodySpan <- srcSpanToRealSrcSpan $ getLoc' bodyLoc
+            any (\commentSpan -> tokenSpan < commentSpan && commentSpan < bodySpan) <$> mCommentSpans
```
amesgen, line 1020: Maybe reduce the line length by moving the lambda to a let binding?

src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -971,28 +971,56 @@ p_if ::
...
+    placeHangingLocated mThenSpan then'
     breakpoint
-    txt "else"
+    locatedToken mElseSpan "else"
     space
-    located else' $ \x ->
-      placeHanging (placer x) (render x)
+    placeHangingLocated mElseSpan else'
+  where
+    (mThenSpan, mElseSpan, mCommentSpans) =
```
amesgen, line 997: Currently, these have types
```haskell
mThenSpan, mElseSpan :: Maybe RealSrcSpan
mCommentSpans :: Maybe [RealSrcSpan]
```
What about changing this to
```haskell
thenSpan, elseSpan :: SrcSpan
commentSpans :: [RealSrcSpan]
```
This allows some minor simplifications below. As a replacement for `Nothing :: Maybe RealSrcSpan`, you can use eg [`noSrcSpan`](https://hackage.haskell.org/package/ghc-lib-parser-9.6.4.20240109/docs/GHC-Types-SrcLoc.html#v:noSrcSpan).

amesgen (approved):
Thanks, this looks good to me. It introduces some complexity, but it is localized, and I think the direction of using something like `locatedToken` could be promising to fix these kinds of subtle comment placement issues; there is also a movement in GHC to store the token spans in a systematic way: https://gitlab.haskell.org/ghc/ghc/-/issues/23447

I diffed our `hackageTests` output before and after this PR, the results look good to me: https://gist.github.com/amesgen/88b2b92cc9e0d7bc13e7fc8357db057a (pipe through eg [delta](https://github.com/dandavison/delta/) for a prettier/easier-to-read diff).

brandonchinn178:
hm Thanks for the hackage diff. It's showing two diffs that I didn't expect (or want):
1. There are some function calls that are now always being put below then/else. I dont want to change behavior there
2. I should also add a test for comments on the same line as then/else. Those should not float above the then/else

I'll take a look tonight

amesgen:
> hm Thanks for the hackage diff. It's showing two diffs that I didn't expect (or want):
> 
>     1. There are some function calls that are now always being put below then/else. I dont want to change behavior there
> 
>     2. I should also add a test for comments on the same line as then/else. Those should not float above the then/else
> 
> 
> I'll take a look tonight

To avoid any confusion: The diff is between the Ormolu output before and after this PR, not between the input and the Ormolu output; the original is not visible (but you can find it on Hackage ofc). Ideally, this would be a three-way diff, but I think I would have to set up a git repo for this?

src/Ormolu/Printer/Meat/Declaration/Value.hs
brandonchinn178: hm I'm open, but I'm not sure how `mThenSpan :: SrcSpan` would simplify it. Can you elaborate?

e.g. the primary operation I need is `thenSpan < commentSpan`. But `commentSpan` is already a `RealSrcSpan`, so I might as well keep `thenSpan` a `RealSrcSpan`. Plus, `SrcSpan` doesn't have an `Ord` instance

amesgen: Eg the diff would look like this (should also remove the `m` prefix of `mThenSpan` and `mElseSpan`):
```diff
diff --git a/src/Ormolu/Printer/Meat/Declaration/Value.hs b/src/Ormolu/Printer/Meat/Declaration/Value.hs
index 1c65bba..5cc1ffb 100644
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -997,28 +997,26 @@ p_if placer render epAnn if' then' else' = do
     (mThenSpan, mElseSpan, commentSpans) =
       case epAnn of
         EpAnn {anns = AnnsIf {aiThen, aiElse}, comments} ->
-          ( Just $ epaLocationRealSrcSpan aiThen,
-            Just $ epaLocationRealSrcSpan aiElse,
+          ( loc' $ epaLocationRealSrcSpan aiThen,
+            loc' $ epaLocationRealSrcSpan aiElse,
             map (anchor . getLoc) $
               case comments of
                 EpaComments cs -> cs
                 EpaCommentsBalanced pre post -> pre <> post
           )
         EpAnnNotUsed ->
-          (Nothing, Nothing, [])
+          (noSrcSpan, noSrcSpan, [])
 
-    locatedToken mTokenSpan token =
-      case mTokenSpan of
-        Just tokenSpan -> located (L tokenSpan ()) $ \_ -> txt token
-        Nothing -> txt token
+    locatedToken tokenSpan token =
+      located (L tokenSpan ()) $ \_ -> txt token
 
-    placeHangingLocated mTokenSpan bodyLoc@(L _ body) = do
+    placeHangingLocated tokenSpan bodyLoc@(L _ body) = do
       let hasComments = fromMaybe False $ do
-            tokenSpan <- mTokenSpan
+            realTokenSpan <- srcSpanToRealSrcSpan $ tokenSpan
             bodySpan <- srcSpanToRealSrcSpan $ getLoc' bodyLoc
-            pure $ any (\commentSpan -> tokenSpan < commentSpan && commentSpan < bodySpan) commentSpans
+            pure $ any (\commentSpan -> realTokenSpan < commentSpan && commentSpan < bodySpan) commentSpans
           placement = if hasComments then Normal else placer body
-      switchLayout ([loc' tokenSpan | Just tokenSpan <- pure mTokenSpan] <> [getLoc' bodyLoc]) $
+      switchLayout [tokenSpan, getLoc' bodyLoc] $
         placeHanging placement (located bodyLoc render)
 
 p_let ::
```
But this is a minor thing, feel free to also leave it as is :+1:

brandonchinn178: ah perfect, thanks!

data/examples/declaration/value/function/if-with-comment-next-to-keyword-out.hs
```diff
@@ -0,0 +1,10 @@
+foo =
+  if x
+    then -- comment 1
+    -- comment 2
```
mrkkrp, line 4: This is the case that you wanted to add, I think. Note that this line is not indented. Not sure if this is what you intended or not. If this is difficult to fix I'd say it is probably fine.

src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -971,28 +971,56 @@ p_if ::
...
+            map (anchor . getLoc) $
+              case comments of
+                EpaComments cs -> cs
+                EpaCommentsBalanced pre post -> pre <> post
+          )
+        EpAnnNotUsed ->
+          (noSrcSpan, noSrcSpan, [])
+
+    locatedToken tokenSpan token =
+      located (L tokenSpan ()) $ \_ -> txt token
```
mrkkrp, line 1011:
```suggestion
      switchLayout [tokenSpan] (txt token)
```

amesgen: This doesn't work (you will see local test failures) as `switchLayout` doesn't emit comments.

mrkkrp: Eh, indeed.

mrkkrp:
@brandonchinn178 Do you consider this ready or is there anything else you'd like to add here?

mrkkrp:
It is nice that keyword spans are now available, it wasn't the case when I was originally implementing this.

brandonchinn178:
This is ready on my end!

brandonchinn178:
oh you added commits. I would appreciate it if you could avoid adding commits to my PRs without consulting me. It's a bit disorienting; if you just left a comment, I would have been perfectly happy to make the changes.

Let me take another look

brandonchinn178:
Sigh I wanted to see if I could improve the comment-next-to-keyword case, but it doesn't look trivial. It seems like the comments are attached to "after the `else` keyword", and there's no way to increase the indentation for comments.

I doubt this will happen often, and this PR is already a huge improvement, so let's merge. Thanks @mrkkrp!

mrkkrp (approved):

mrkkrp:
Thanks a lot!

PR #1095, opened 2024-02-01
Prioritize fixity overrides and module re-exports from command line

### LINKED ISSUE #1030: Fixity info from .ormolu overrides CLI flags
**Describe the bug**

Reading the code here:
https://github.com/tweag/ormolu/blob/49eb083d2b94ad882b1ff5f8b65f9870ee94a53d/src/Ormolu.hs#L193-L207

It seems like configuration in `.ormolu` files take precedence over command line flags, when the help text (and standard behavior of flags vs config files) seems to indicate that flags take precedence. Specifically, `Map.union` is left-biased, so `fixityOverrides` (coming from `mfixityOverrides`, which comes from `mDotOrmolu`) is preferred over `cfgFixityOverrides rawConfig` (which is parsed from command line flags)

**To Reproduce**
1. Create some Haskell file with fixities
2. Specify the wrong fixity in `.ormolu`
3. Specify the right fixity with `--fixity`
4. Observe ormolu formatting with the wrong fixity

**Expected behavior**
Ormolu should format with the fixity specified on the command line

**Environment**
 - OS name + version:
 - Version of the code:

**Additional context**

mrkkrp:
Close #1030.


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
     default)
   * GHC proposal [#134](https://github.com/ghc-proposals/ghc-proposals/blob/0b652bd70258e354dfe4a05940182007596f8bf7/proposals/0134-deprecating-exports-proposal.rst): deprecating/warning about exports
   * GHC proposal [#541](https://github.com/ghc-proposals/ghc-proposals/blob/0b652bd70258e354dfe4a05940182007596f8bf7/proposals/0541-warning-pragmas-with-categories.rst): warning categories
+* Now command line options for fixity overrides and module re-exports
+  overwrite information from `.ormolu` files. [Issue
+  1030](https://github.com/tweag/ormolu/issues/1030).
```
amesgen, line 12: This should probably go to the "unreleased" section (it already exists, but this PR has to be rebased).

fixity-tests/test-0-cli-overrides-dotormolu-expected.hs
amesgen, line 1: This is the same as `test-0-no-extra-info-expected.hs`, maybe this file (`test-0-cli-overrides-dotormolu-expected.hs`) could be deleted, and maybe even additionally, you could do a rename like

 - `test-0-no-extra-info-expected.hs` ⟶ `test-0-ugly-expected.hs`
 - `test-0-with-fixity-info-expected.hs` ⟶ `test-0-beautiful-expected.hs`

amesgen (approved):
Agreeing that the approach described in https://github.com/tweag/ormolu/issues/1030#issuecomment-1556696659 looks good :+1:

PR #1097, opened 2024-02-09
Make Ormolu print debug info about operator fixity inference

### LINKED ISSUE #1060: Make it easier to debug fixity issues

mrkkrp:
Close #1060, close #1069.

### DIFF
--- a/tests/Ormolu/FixitySpec.hs
+++ b/tests/Ormolu/FixitySpec.hs
@@ -261,7 +261,7 @@ checkFixities dependencies fixityImports expectedResult =
   where
     actualResult =
       fmap
-        (\(k, _) -> (k, inferFixity k resultMap))
+        (\(k, _) -> (k, inferFixity False k resultMap))
         expectedResult
     resultMap =
       moduleFixityMap
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 * Respect newlines in data declarations in more cases. [Issue
   1077](https://github.com/tweag/ormolu/issues/1077) and [issue
   947](https://github.com/tweag/ormolu/issues/947).
+* The `-d / --debug` command line option now makes Ormolu print out debug
+  information regarding operator fixity inference. [Issue
+  1060](https://github.com/tweag/ormolu/issues/1060).
 
 ## Ormolu 0.7.3.0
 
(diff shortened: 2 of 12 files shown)


brandonchinn178 (approved):
My other PR also added troubleshooting docs to the README. Perhaps that would still be useful to add?

I'm still not a fan of manually passing Bools all the way down the stack. I also think it would be better if there were one way to log things, instead of some places using `hPutStrLn stderr` and some using `trace`.

tests/Ormolu/FixitySpec.hs
```diff
@@ -261,7 +261,7 @@ checkFixities dependencies fixityImports expectedResult =
   where
     actualResult =
       fmap
-        (\(k, _) -> (k, inferFixity k resultMap))
+        (\(k, _) -> (k, inferFixity False k resultMap))
```
amesgen, line 264: Call sites like this would make it nice if we used some dedicated type instead of `Bool`, eg
```haskell
data DebugLogging = DebugLoggingEnabled | DebugLoggingDisabled
```
(unless this is too annoying, but I don't immediately see a big obstacle)

amesgen (approved):

tests/Ormolu/FixitySpec.hs
mrkkrp: Yeah, but then in general it applies to pretty much any `Bool` and we have quite a few of them, see `Config` for example. Maybe a separate PR could be opened for this kind of refactoring.

amesgen: Do you have an opinion whether [`choice`](https://github.com/mboes/choice/#choice-a-solution-to-boolean-blindness) might be worth it? I think it could be a nice as an alternative to our various "blind `Bool`s" (and even some ad-hoc two-constructor types, like `IsApplicand` or `ForAllVisibility`).

mrkkrp: `choice` looks nice!

Issue #901, opened 2022-06-22
Ormolu drops a Haddock section header in module export


ocharles:
**Describe the bug**

The following code

```haskell
module Beckhoff.ADS
  ( -- $intro
    -- * Connections
    ConnectInfo(..)
  ) where
```

becomes

```haskell
module Beckhoff.ADS
  ( -- $intro
    ConnectInfo (..),
  )
where
```

Note that `* Connections` is no longer in the result.

**To Reproduce**
Try formatting the "before" code in https://ormolu-live.tweag.io/.

**Expected behavior**
Both `$intro` and `* Connections` are present in the formatted result.

**Environment**
Ormolu live

ocharles:
I note that 

```haskell
module Beckhoff.ADS
  ( -- $intro

    -- * Connections
    ConnectInfo(..)
  ) where
```

works, so I do have a workaround.

amesgen:
Hmm, I think that Ormolu's output is actually "Haddock-preserving" here. The rendered Haddock of
```haskell
module MyLib
  ( -- $intro
    -- * Connections
    Foo (..)
  ) where

data Foo = Foo
```
and
```haskell
module MyLib
  ( -- $intro
    Foo (..)
  ) where

data Foo = Foo
```
is in both cases
![2022-06-23_16-43](https://user-images.githubusercontent.com/15369874/175326948-42cf2fc1-3c41-4062-99b7-19bcbbae6c18.png)

while for
```haskell
module MyLib
  ( -- $intro

    -- * Foo
    Foo (..)
  ) where

data Foo = Foo
```
yields
![2022-06-23_16-45](https://user-images.githubusercontent.com/15369874/175327372-ad422668-d7be-460f-bbba-3f725f0cc858.png)

So while I agree it is surprising, maybe Ormolu's behavior is actually helpful as it only mirrors the Haddock behavior of removing everything after `-- $intro` in the same comment?

ocharles:
Ah, interesting. That said, I do find Ormolu makes it easy to now just lose the comment entirely if you're not careful reading the diff

amesgen:
> That said, I do find Ormolu makes it easy to now just lose the comment entirely if you're not careful reading the diff

Agreed; so we basically have two options:

 - Consider this to be an upstream (Haddock/GHC parser) issue.
 - Change our approach to format Haddock snippets.

   Right now, Haddock comments occur in two places in the GHC AST:
   
    1. In the list of "ordinary" comments, e.g.
[...]

PR #1103, opened 2024-03-11
ghc-lib-parser 9.10

amesgen:
Update to `ghc-lib-parser-9.10`, build with GHC 9.10.1 in CI, drop GHC 9.4, update bounds.

 - See the changelog for details about the new syntactic features in GHC 9.10.
 - We can remove our `HasSrcSpan` in favor of `ghc-lib-parser`'s `HasLoc`.
 - The EPA extension points in constructors do (usually) no longer contain the comments (they are now usually in the annotation part of a `Located`). We relied on this in `p_if` (as of #1092). Instead of laboriously passing the comments down from where we unwrap the `LHsExpr`, I added a new function `getEnclosingComments` to the `R` monad that leverages the comment stream, as that seemed much less invasive.
 - See the review comments for some further minor points.


src/Ormolu/Fixity/Parser.hs
```diff
@@ -45,6 +45,17 @@ import Text.Megaparsec.Char.Lexer qualified as L
...
+-- TODO Support fixity namespacing?
+-- https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0065-type-infix.rst
+--
+-- Note that currently, our fixity machinery does *not* do any namespacing:
+--
+--  - https://github.com/tweag/ormolu/pull/994#pullrequestreview-1396958951
+--    brought this up in the past
+--
+--  - https://github.com/tweag/ormolu/pull/1029#issue-1718217029
+--    has a concrete example (morley-prelude) where namespacing would matter
```
amesgen, line 57: WDYT?

data/examples/declaration/value/function/pragmas-out.hs
```diff
@@ -1,4 +1,4 @@
-sccfoo = {-# SCC foo #-} 1
+sccfoo = {-# SCC "foo" #-} 1
```
amesgen, line 1: This is due to a change in an `Outputable` instance; we could choose to not use it and keep the previous behavior.

tchoutri:
I'm very excited for this PR to land, I am starting a project with GHC 9.10. :)

data/examples/declaration/value/function/pragmas-out.hs
mrkkrp: I like the new behavior, it adds a little bit more of normalization :-)

mrkkrp (approved):

src/Ormolu/Fixity/Parser.hs
amesgen: Opened https://github.com/tweag/ormolu/issues/1113

data/examples/backpack/signature-0-out.hsig
```diff
@@ -15,10 +15,10 @@ instance Primitive Prim
 
 name :: String
 
--- \| The name of the primitive used as the seed stretcher
--- \| Test line 2
+-- | The name of the primitive used as the seed stretcher
+-- | Test line 2
 --  | Test line 3
--- \|Test line 4
+-- |Test line 4
```
brandonchinn178, line 21: I was curious why this changed. Seems like GHC [merged a change](https://gitlab.haskell.org/ghc/ghc/-/commit/b444c16f4ff64938a8bec9587bd90209bda682b9) that causes signature modules to now parse Haddocks correctly. So now ghc-lib-parser returns an AST containing `HsDocStringChunk "  | Test line 2"` instead of the previously incorrect `EpaLineComment "  \\| Test line 2"`.

However, now I see that these docs are in two places: in an `EpaComments` annotation and in a `DocD` declaration. I can't find the logic that prevents Ormolu from printing these out twice. Anyone know where that is?

amesgen: > However, now I see that these docs are in two places: in an `EpaComments` annotation and in a `DocD` declaration.

Yeah, this has always been the case, see e.g. https://github.com/tweag/ormolu/issues/901#issuecomment-1166252693 for another example/more context.

> I can't find the logic that prevents Ormolu from printing these out twice. Anyone know where that is?

It is here, specifically line 75: https://github.com/tweag/ormolu/blob/a9e996b5f1ad6bb00e3c487e8411a858696580de/src/Ormolu/Parser/CommentStream.hs#L71-L78

PR #1116, opened 2024-06-02
Indent multiline bracketed expr with minimal amount of space

brandonchinn178:
Supercedes #1101 

AFAICT, the only motivation for `BracketStyle` is to prevent the closing bracket from being in the same column as the opening bracket, which can be a parse error in certain contexts, like case patterns. So this only needs to indent by a single space, not a full indentation.

It's not that big of an issue here (1 vs 2 spaces), but it's a bigger issue with larger indentations. So I had to make this fix in Fourmolu anyway, but I'm opening this PR just in case you want to merge it upstream. IMO this encodes the intention better, but I recognize that the breakage might not be worthwhile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Fix Haddock comments on infix constructors
   [Issue 758](https://github.com/tweag/ormolu/issues/758)
+* Multiline bracketed expressions indent closing bracket by just 1 space instead of a full indentation
 
 ## Ormolu 0.7.5.0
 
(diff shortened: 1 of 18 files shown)


mrkkrp:
In my head, a code snippet is correctly indented iff all its lines start at columns that are divisible by the indentation step. I can see why it looks weird in Fourmolu with larger indentation, but Ormolu does not have that problem and so this change seems unnecessary.

brandonchinn178:
@mrkkrp Understood. FWIW, there is at least one example where Ormolu does not indent along multiples of 2:

https://github.com/tweag/ormolu/blob/d90413c0c228fa43da0967d22c7a55c18969cdac/data/examples/declaration/value/function/required-type-arguments-out.hs#L16-L21

But I understand your point in general. I'll close this PR

PR #1130, opened 2024-07-08
Format `.cabal` files with cabal-gild

amesgen:
See https://github.com/tweag/ormolu/pull/1129#issuecomment-2214104633 and https://github.com/tweag/ormolu/pull/1129#issuecomment-2215060315

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,11 +7,7 @@
     };
     nixpkgs.follows = "haskellNix/nixpkgs-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-    pre-commit-hooks = {
-      url = "github:cachix/pre-commit-hooks.nix";
-      inputs.nixpkgs.follows = "nixpkgs";
-      inputs.flake-utils.follows = "flake-utils";
-    };
+    pre-commit-hooks.url = "github:cachix/git-hooks.nix";
 
     # for Ormolu Live
     ghc-wasm-meta.url = "gitlab:ghc/ghc-wasm-meta?host=gitlab.haskell.org";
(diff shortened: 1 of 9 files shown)


flake.nix
```diff
@@ -7,11 +7,7 @@
     };
     nixpkgs.follows = "haskellNix/nixpkgs-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-    pre-commit-hooks = {
-      url = "github:cachix/pre-commit-hooks.nix";
-      inputs.nixpkgs.follows = "nixpkgs";
-      inputs.flake-utils.follows = "flake-utils";
-    };
+    pre-commit-hooks.url = "github:cachix/git-hooks.nix";
```
mrkkrp, line 10: Why are the "follows" attributes no longer necessary?

nix/format.sh
```diff
@@ -2,10 +2,6 @@
 
 export LANG="C.UTF-8"
 
-cabal format
```
mrkkrp, line 5: Previously we instructed contributors to run `nix run .#format` (in `CONTRIBUTING.md`) which would take care of formatting cabal files. Should we add something there about the new way of doing things?

flake.nix
amesgen: The `flake-utils` input is no longer necessary since https://github.com/cachix/git-hooks.nix/pull/446, and the `nixpkgs` input of `haskell.nix` is too old to contain `cabal-gild`, so I now just use the `nixpkgs` input by `git-hooks.nix`.

nix/format.sh
amesgen: The most direct answer would be: Run `pre-commit` in the Nix shell. E.g. the `nixpkgs-fmt` check was only enforced by that, not by `format.sh`.

I don't feel super strongly about the pre-commit hooks, we could also move everything to `format.sh`.

mrkkrp: Hmm, I see. I never actually ran `pre-commit` myself. I think we can use `pre-commit`, let's just mention it in `CONTRIBUTING.md`.

mrkkrp (approved):

PR #1133, opened 2024-09-13
Implement support for fractional precedence

### LINKED ISSUE #1106: Support “fine-grained” operator precedence
I sometimes end up with long chains of operators with identical precedence. E.g., in some parser code, I have
```haskell
infixr 3 >~< -- run two parsers with whitespace between them
infixr 3 |~| -- run two parsers with optional whitespace between them
infixr 3 <~> -- run two parsers without whitespace between them
```
They _should_ have the same precedence, because they’re mutually associative, so that’s all good. But when Ormolu formats a chain split across multiple lines, it can get noisy
```haskell
startFormTok
  |~| messageTag
  >~< startMessageTok
  |~| name
  >~< p'
  |~| endMessageTok
  |~| endFormTok
```
What I would like is for Ormolu to treat them so that required whitespace is lower precedence than optional, is lower than illegal whitespace, so I can have it formatted like
```haskell
startFormTok |~| messageTag
  >~< startMessageTok |~| name
  >~< p' |~| endMessageTok |~| endFormTok
```

I think a fairly clean solution would be for the .ormolu file to parse the precedences as floating point, rather than integers. Existing .ormolu files should work as is, but I can edit it to look like
```haskell
infixr 3 >~<
infixr 3.3 |~|
infixr 3.7 <~>
```
and get the behavior I want.

As far as working with #846, perhaps regeneration could update the integral part without affecting the fractional part, so the fine-grained relationships are preserved even if the operator precedence has shifted.

What I currently do is fill my .ormolu with lies. Using incorrect precedences like
```haskell
infixr 3 >~<
infixr 4 |~|
infixr 5 <~>
```
gets me the behavior I want, but I have to be careful to adjust any other precedences that bump into these.

mrkkrp:
Close #1106.


src/Ormolu/Fixity/Internal.hs
```diff
@@ -116,10 +116,10 @@ data FixityApproximation = FixityApproximation
     faDirection :: Maybe FixityDirection,
     -- | Minimum precedence level found in the (maybe conflicting)
     -- definitions for the operator (inclusive)
-    faMinPrecedence :: Int,
+    faMinPrecedence :: Double,
     -- | Maximum precedence level found in the (maybe conflicting)
     -- definitions for the operator (inclusive)
-    faMaxPrecedence :: Int
+    faMaxPrecedence :: Double
```
amesgen, line 122: Hmm, sadly, the default `Binary` instance for `Double` is quite inefficient, see https://github.com/kolmodin/binary/issues/69.

In particular, we get a 565KB increase (~60%!, from 0.99MB to 1.56MB) for `hackage-info.bin`, and at least a ~12ms slowdown for parsing it:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `echo 1+++++1 \| master/bin/ormolu --no-cabal` | 29.9 ± 7.3 | 22.5 | 77.2 | 1.00 |
| `echo 1+++++1 \| 1106/bin/ormolu --no-cabal` | 42.4 ± 2.2 | 32.5 | 44.2 | 1.42 ± 0.35 |


If we instead write a manual `Binary` instance for these types via [`putDoublele`](https://hackage.haskell.org/package/binary-0.8.9.2/docs/Data-Binary-Put.html#v:putDoublele)/`getDoublele`, the size only increases by 12KB (from 0.99MB to 1.00MB) and the parsing time increases only by 2-3ms:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `echo 1+++++1 \| master/bin/ormolu --no-cabal` | 29.1 ± 4.9 | 21.9 | 33.6 | 1.00 |
| `echo 1+++++1 \| 1106/bin/ormolu --no-cabal` | 42.1 ± 2.6 | 32.0 | 53.0 | 1.45 ± 0.26 |
| `echo 1+++++1 \| 1106-better-double/bin/ormolu --no-cabal` | 32.7 ± 1.6 | 22.2 | 34.1 | 1.13 ± 0.20 |

```diff
diff --git a/src/Ormolu/Fixity/Internal.hs b/src/Ormolu/Fixity/Internal.hs
index 40e6e3e..4c84602 100644
--- a/src/Ormolu/Fixity/Internal.hs
+++ b/src/Ormolu/Fixity/Internal.hs
@@ -32,6 +32,9 @@ where
 
 import Control.DeepSeq (NFData)
 import Data.Binary (Binary)
+import Data.Binary qualified as Binary
+import Data.Binary.Get qualified as Binary
+import Data.Binary.Put qualified as Binary
 import Data.ByteString.Short (ShortByteString)
 import Data.ByteString.Short qualified as SBS
 import Data.Choice (Choice)
@@ -99,7 +102,17 @@ data FixityInfo = FixityInfo
     fiPrecedence :: Double
   }
   deriving stock (Eq, Ord, Show, Generic)
-  deriving anyclass (Binary, NFData)
+  deriving anyclass (NFData)
+
+instance Binary FixityInfo where
+  put FixityInfo {..} = do
+    Binary.put fiDirection
+    Binary.putDoublele fiPrecedence
+
+  get = do
+    fiDirection <- Binary.get
+    fiPrecedence <- Binary.getDoublele
+    pure FixityInfo {..}
 
 -- | Fixity info of the built-in colon data constructor.
 colonFixityInfo :: FixityInfo
@@ -122,7 +135,19 @@ data FixityApproximation = FixityApproximation
     faMaxPrecedence :: Double
   }
   deriving stock (Eq, Ord, Show, Generic)
-  deriving anyclass (Binary, NFData)
+  deriving anyclass (NFData)
+
+instance Binary FixityApproximation where
+  put FixityApproximation {..} = do
+    Binary.put faDirection
+    Binary.putDoublele faMinPrecedence
+    Binary.putDoublele faMaxPrecedence
+
+  get = do
+    faDirection <- Binary.get
+    faMinPrecedence <- Binary.getDoublele
+    faMaxPrecedence <- Binary.getDoublele
+    pure FixityApproximation {..}
 
 -- | Gives the ability to merge two (maybe conflicting) definitions for an
 -- operator, keeping the higher level of compatible information from both.
```

(We could also use sth like the [`OnField` microsurgery](https://hackage.haskell.org/package/generic-data-1.1.0.0/docs/Generic-Data-Internal-Microsurgery.html#t:OnField) to still generically derive this `Binary` instance :smile:)

src/Ormolu/Fixity/Printer.hs
```diff
@@ -75,3 +76,12 @@ renderSingleModuleReexport (exportingModule, exports) =
...
 renderModuleName = B.fromString . intercalate "." . ModuleName.components
+
+-- | Render precedence using integer representation for whole numbers.
+renderPrecedence :: Double -> Builder
+renderPrecedence x =
+  let (n :: Int, fraction :: Double) = properFraction x
+      isWholeEnough = fraction < 0.0001
+   in if isWholeEnough
+        then B.decimal n
+        else B.realFloat x
```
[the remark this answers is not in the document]
mrkkrp: Yeah, I do not have great hope to make floats roundtrip anyway, see e.g. mrkkrp/megaparsec#479, so I intentionally went for this "crude" approach.

src/Ormolu/Fixity/Internal.hs
mrkkrp: I will adjust this as you suggest. I noticed the increase in the size of the database, but did not investigate what caused it.

amesgen (approved):

sellout:
Fantastic! I was definitely just waiting for the inevitable “won’t fix” on this one, but I’m excited to see it implemented!

PR #1142, opened 2024-11-06
Allow Diff-1.0

mrkkrp:


amesgen (approved):
Also locally tested that it builds fine

PR #1170, opened 2025-07-16
Make RequiredTypeArguments consistent with types

brandonchinn178:
Input:
```hs
typeSig ::
  ( Int
  -> Int -> Int
  )
  
requiredType =
  f ( Int
    -> Int -> Int
    )
```

Current output:
```hs
typeSig ::
  ( Int ->
    Int ->
    Int
  )

requiredType =
  f
    ( Int ->
      Int -> Int
    )
```

After this PR:
```hs
typeSig ::
  ( Int ->
    Int ->
    Int
  )

requiredType =
  f
    ( Int ->
      Int ->
      Int
    )
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
   fixity, such as `infix +`. [Issue
   1166](https://github.com/tweag/ormolu/issues/1166).
 
+* Make multiline function signatures in RequiredTypeArguments consistent with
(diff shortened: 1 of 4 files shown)


amesgen (approved):
Thanks for catching that!

Can you add a changelog entry?

brandonchinn178:
Not sure why formatting took forever, but should be good to go

PR #1186, opened 2025-12-31
allow optparse-applicative 0.19

juhp:
This would also unlock ormolu in Stackage


mrkkrp:
Thanks!

juhp:
Perhaps a Hackage revision could be made?

mrkkrp:
Yes, done.

PR #1175, opened 2025-08-20
ghc-lib-parser 9.14

amesgen:
Draft until GHC 9.14.1 gets released

 1. Update flake inputs and GHCs, bump `.cabal` version bounds

    Also drop GHC 9.8 (which allows us to upgrade to GHC2024 in a follow-up PR).

 2. Update to `ghc-lib-parser-9.14`. See the changelog entry for new syntactic features.
 3. Update to `Cabal-syntax-3.14`.
 4. Fix #1165 (depends on an upstream GHC parser fix).

PR #1190, opened 2026-02-06
Fix build on GHC 9.14

georgefst:
Note that this targets #1175, not `master`.


src/Ormolu/Diff/ParseResult.hs
```diff
@@ -153,6 +153,7 @@ diffHsModule = genericQuery
     epAnnEq :: EpAnn a -> b -> ParseResultDiff
     epAnnEq _ _ = Same
 
+    importDeclQualifiedStyleEq :: forall a. (Data a) => ImportDeclQualifiedStyle -> a -> ParseResultDiff
```
georgefst, line 156: I actually have no idea what change makes this necessary.

mrkkrp:
Thanks! Applied as part of #1175.

PR #1199, opened 2026-05-04
Do not indent body on multiline or-pattern case match

### LINKED ISSUE #1188: Or pattern case branches are doubly indented
**Describe the bug**
When using Or-Patterns, case branches are indented twice.

**To Reproduce**
```hs
main =
  case x of
    1
    2 ->
      putStrLn "asdf"
    _ ->
      putStrLn "asdf"
```

Results in
```hs
main =
  case x of
    1
    2 ->
        putStrLn "asdf"
    _ ->
      putStrLn "asdf"
```

**Expected behavior**
The above repro should be a fixed point

**Environment**
 - OS name + version:
 - Version of the code:

**Additional context**

joaosreis:
Multiline or-pattern case match would trigger a body indentation, while using `placeHanging Normal` to render the body, resulting in a second unwanted body indentation.
This PR fixes this issue by adding a clause to trigger the body indentation that checks that the we are not in a case with or-patterns.
Test cases were updated to cover a situation affected by this change.
Fixes #1188.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@
 * Fixed printing of multi-line or-patterns inside as-patterns. [Issue
   1183](https://github.com/tweag/ormolu/issues/1183).
 
+* Fixed an issue where or-patterns would be indented twice. [Issue
+  1188](https://github.com/tweag/ormolu/issues/1188).
+
 ## Ormolu 0.8.0.2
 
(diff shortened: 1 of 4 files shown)


mrkkrp (changes requested):
LGMT, can you add a changelog entry and make sure we avoid exceeding 80 characters in line lengths if possible (line 208 in particular).

mrkkrp (approved):

PR #1200, opened 2026-05-11
Add support for explicit level imports in import declarations

### LINKED ISSUE #1192: Ormolu support for ExplicitLevelImports
**Describe the bug**

Ormolu will drop the `splice` (or `quote`) `import` informations when the code use `ExplicitLevelImports`.

**To Reproduce**

Test.hs:
```haskell
{-# LANGUAGE ExplicitLevelImports #-}
import splice PyF (fmt)
import PyF
```

`ormolu Test.hs`:

```diff
Test.hs
@@ -1,3 +1,4 @@
  {-# LANGUAGE ExplicitLevelImports #-}
- import splice PyF (fmt)
+
  import PyF
+ import PyF (fmt)

  AST of input and AST of formatted code differ.
    at Test.hs:2:1-23
  Please, consider reporting the bug.
  To format anyway, use --unsafe.
```

See how it dropped the `splice`

**Expected behavior**

Ormolu should correctly handle `splice` and `quote`

**Environment**
 - OS name + version:
 - Version of the code: Ormolu built using #1190

**Additional context**

joaosreis:
Introduce support for import declaration level specifiers and provide test examples for explicit level imports.

Fixes #1192.

### DIFF
--- a/data/examples/import/explicit-level-imports.hs
+++ b/data/examples/import/explicit-level-imports.hs
@@ -6,3 +6,5 @@ import Data.ByteString (e)
 import {-# SOURCE #-} safe qualified A splice as QA hiding (a, b, c, d, e, f)
 import quote qualified B as QB
 import qualified C splice as SC
+import A splice
+import qualified D splice
--- a/data/examples/import/explicit-level-imports.hs
+++ b/data/examples/import/explicit-level-imports.hs
@@ -3,3 +3,6 @@
 import splice Data.Text (a, b, c)
 import Data.ByteString.Lazy quote (d)
 import Data.ByteString (e)
+import {-# SOURCE #-} safe qualified A splice as QA hiding (a, b, c, d, e, f)
+import quote qualified B as QB
+import qualified C splice as SC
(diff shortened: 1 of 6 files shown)


data/examples/import/explicit-level-imports.hs
```diff
@@ -0,0 +1,5 @@
+{-# LANGUAGE ExplicitLevelImports #-}
```
mrkkrp, line 1: Can we try to cover all possible positions/combinations of `splice` and `quote` in pre- and post- forms, in all acceptable positions relative to `qualified`, etc. I was reading this https://ghc-proposals.readthedocs.io/en/latest/proposals/0682-explicit-level-imports.html#syntax-for-imports

src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -44,6 +44,11 @@ p_hsmodImport ImportDecl {..} = do
   space
   when (ideclSource == IsBoot) (txt "{-# SOURCE #-}")
   space
+  case ideclLevelSpec of
```
mrkkrp, line 49: It would probably be nice to reduce duplication a little bit by having a helper `ImportDeclLevel -> R ()`.

mrkkrp (changes requested):
LGTM, modulo the little remarks I made. As usual, let's make sure we have a proper changelog entry for this.

mrkkrp (approved):

PR #1204, opened 2026-06-03
feat: improve constructor Haddock placement

### LINKED ISSUE #583: Haddock docstring placement for constructors

The following snippet

```haskell
data Foo
 = Bar -- ^ bar
 | Baz -- ^ baz
```

is formatted as

```haskell
data Foo
  = -- | bar
    Bar
  | -- | baz
    Baz
```

It's a bit awkward: the Haddock comments are placed before the constructor, probably because this is the placement for record fields. However, the reason why they are placed first in fields is because the punctuation is a the end of the line; in constructor, though, the “punctuation” is at the beginning of the line.

I think either of the two formatting would be more consistent:

- punctuation at the end/Haddock at the beginning
    ```haskell
    data Foo =
      -- | bar
      Bar |
      -- | baz
      Baz
     ```
- punctuation at the begining/Haddock at the end
    ```haskell
    data Foo
     = Bar
       -- ^ bar
     | Baz
       -- ^ baz
    ```

mtibben:
Ormolu currently formats nullary constructors with trailing Haddocks by moving the documentation between the sum separator and the constructor name, which makes enum-like declarations harder to scan:

```haskell
data Foo
  = -- | bar
    Bar
  | -- | baz
    Baz
```

This changes the H98 prefix-constructor printer so single-line Haddocks on simple nullary constructors remain next to the constructor name:

```haskell
data Foo
  = Bar -- ^ bar
  | Baz -- ^ baz
```

The change is intentionally limited to single-line docs on constructors with no fields, context, or explicit binders. Constructors with arguments, multi-line constructor docs, records, GADTs, and infix constructors continue to use the existing placement so constructor docs are not confused with argument docs.

Fixes #583.

### DIFF
--- a/data/examples/declaration/data/multiline-out.hs
+++ b/data/examples/declaration/data/multiline-out.hs
@@ -2,11 +2,12 @@ module Main where
 
 -- | Here we have 'Foo'.
 data Foo
-  = -- | One
-    Foo
+  = Foo -- ^ One
   | -- | Two
     Bar Int
-  | -- | Three
-    Baz
+  | Baz -- ^ Three
+  | -- | Four
+    -- more about four
+    Qux
   deriving
     (Eq, Show)
--- a/data/examples/declaration/data/multiline.hs
+++ b/data/examples/declaration/data/multiline.hs
@@ -6,5 +6,8 @@ data Foo
   = Foo -- ^ One
   | Bar Int -- ^ Two
   | Baz -- ^ Three
+  | Qux
+    -- ^ Four
+    -- more about four
   deriving
     (Eq, Show)
--- a/data/examples/declaration/data/partly-documented-out.hs
+++ b/data/examples/declaration/data/partly-documented-out.hs
@@ -1,5 +1,4 @@
 data Optimisation
   = PETransform
-  | -- | partial eval and associated transforms
-    GeneralisedNatHack
+  | GeneralisedNatHack -- ^ partial eval and associated transforms
   deriving (Show, Eq, Generic)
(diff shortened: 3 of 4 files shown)


mrkkrp:
Hi, thanks for taking a look. I do realize there is that very old issue about it, which I kept open for the sake of conversation, not because I agree it should be implemented. It is not different from many other issues tagged "style" which are often just time sinks once we start arguing about aesthetics and opinions.

The arguments for keeping the status quo would be:

* It is the way we've been formatting things for many years now and when a project enjoys as wide adoption as Ormolu does we must be conservative with respect to formatting changes that are purely stylistic (i.e. there is no solid technical reason to prefer one way or another).
* You yourself mention that the change special-cases "single-line docs on constructors with no fields, context, or explicit binders". It can be argued that consistency is more valuable then brevity and familiarity of `-- ^`. It seems undesirable to me to have `-- |` vs `-- ^` based on whether a Haddock happens to be single-line or multiline or whether the corresponding constructor takes an argument or not. With this change, it is not impossible to have both styles of Haddock in a single data type definition:

```haskell
data Foo
  = -- | This is my foo.
    Foo Int
  | Bar -- ^ And this is my bar
```

Which does not really even look good.

Compare with the status quo:

```haskell
data Foo
  = -- | This is my foo.
    Foo Int
  | -- | And this is my bar
    Bar
```

So, as much as I hate it, sometimes I have to make some choices on style in this project. I like all my choices in open source to be logical and defendable, but the argumentation can only go so far here, since it is largely an aesthetic choice + conservatism. At this point I am inclined to preserve the status quo rather than merge this change.

mtibben:
Understood, and thank you for the clear explanation!

PR #1206, opened 2026-06-22
Force multiline layout for export lists with doc comments

### LINKED ISSUE #1051: Haddock comment in single line export list breaks format
**Describe the bug**
A multiline haddock in a single line export list incorrectly changes to a single line haddock comment.

**To Reproduce**
```hs
module Foo ({- | asdf -} foo) where
```

Currently, formatting this with `--unsafe` returns:
```hs
module Foo (-- | asdf foo) where
```

**Expected behavior**
Should reformat to
```hs
module Foo
  ( -- | asdf
    foo
  ) where
```
or stay with multiline haddock

**Environment**
 - OS name + version:
 - Version of the code:

**Additional context**

joaosreis:
This PR forces a multiline layout when export lists contain Haddock comments.

Fixes #1051.
This is the most straightforward way of fixing the aforementioned issue. An alternative approach would be to maintain a block comment in such situations. However, this is not the current expected behavior of ormolu (see #641).

### DIFF
--- a/src/Ormolu/Printer/Meat/ImportExport.hs
+++ b/src/Ormolu/Printer/Meat/ImportExport.hs
@@ -8,6 +8,7 @@
 module Ormolu.Printer.Meat.ImportExport
   ( p_hsmodExports,
     p_hsmodImport,
+    enterMultilineLayoutIfContainsDocEntries,
   )
 where
 
@@ -25,12 +26,13 @@ import Ormolu.Utils (RelativePos (..), attachRelativePos)
 
 p_hsmodExports :: [LIE GhcPs] -> R ()
 p_hsmodExports xs =
-  parens N $ do
-    layout <- getLayout
-    sep
-      breakpoint
-      (\(p, l) -> sitcc (located (addDocSrcSpan l) (p_lie layout p)))
-      (attachRelativePos xs)
+  enterMultilineLayoutIfContainsDocEntries xs $
+    parens N $ do
+      layout <- getLayout
+      sep
+        breakpoint
+        (\(p, l) -> sitcc (located (addDocSrcSpan l) (p_lie layout p)))
+        (attachRelativePos xs)
   where
     -- In order to correctly set the layout when a doc comment is present.
     addDocSrcSpan lie@(L l ie) = case ieExportDoc ie of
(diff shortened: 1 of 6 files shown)


src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -19,18 +20,20 @@ import GHC.LanguageExtensions.Type
 import GHC.Types.PkgQual
 import GHC.Types.SrcLoc
 import Ormolu.Printer.Combinators
+import Ormolu.Printer.Internal (enterLayout)
```
mrkkrp: The pattern of use for `enterLayout` suggests that it should become a public combinator, so let's just re-export it from `Ormolu.Printer.Combinators` and remove these imports from `Ormolu.Printer.Internal`.

src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -19,18 +20,20 @@ import GHC.LanguageExtensions.Type
...
 
 p_hsmodExports :: [LIE GhcPs] -> R ()
 p_hsmodExports xs =
-  parens N $ do
-    layout <- getLayout
-    sep
-      breakpoint
-      (\(p, l) -> sitcc (located (addDocSrcSpan l) (p_lie layout p)))
-      (attachRelativePos xs)
+  (if any (isDocEntry . unLoc) xs then enterLayout MultiLine else id) $
```
mrkkrp: There is a duplication of logic in this module and in `Ormolu.Printer.Meat.Module`. Let us introduce a helper that does a little more than `isDocEntry`:

```haskell
enterMultilineLayoutIfContainsDocEntries :: [LIE GhcPs] -> R () -> R ()
enterMultilineLayoutIfContainsDocEntries xs =
  if any (isDocEntry . unLoc) xs
    then enterLayout MultiLine
    else id
```

Then this helper can be exported instead of `isDocEntry` and also used in `ImportExport.hs` directly.

src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -37,7 +38,8 @@ p_hsModule mstackHeader pragmas HsModule {..} = do
   let XModulePs {..} = hsmodExt
       deprecSpan = maybe [] (pure . getLocA) hsmodDeprecMessage
       exportSpans = maybe [] (pure . getLocA) hsmodExports
-  switchLayout (deprecSpan <> exportSpans) $ do
```
mrkkrp, line 40: It may be more natural to leverage the fact that `switchLayout Multiline` will override the layout of the enclosing block. So, we could write:

```haskell
  switchLayout (deprecSpan <> exportSpans) $
    enterMultilineLayoutIfContainsDocEntries (maybe [] unLoc hsmodExports) $ do
```

...which arguably reads nicer.

mrkkrp (changes requested):
Very nice! Can you please also add a changelog entry? This would go in a new "Unreleased" section.

mrkkrp (approved):

PR #1214, opened 2026-08-06
Don't single out Prelude imports when NoImplicitPrelude is on

### LINKED ISSUE #1189: Don’t single out `Prelude` import with `NoImplicitPrelude`
`Prelude` always gets sorted to the end of the imports.

``` haskell
import Control.Applicative (Alternative, (<|>))
import Data.Maybe (Maybe (Nothing), maybe)
import System.IO (IO)
import Prelude ((+))
```

This makes sense in the default case, when `ImplicitPrelude` is on.

However, with `NoImplicitPrelude`, `Prelude` no longer has any special status, so it would be nice to sort it among the other imports.

``` haskell
import Control.Applicative (Alternative, (<|>))
import Data.Maybe (Maybe (Nothing), maybe)
import Prelude ((+))
import System.IO (IO)
```

And similarly with package-qualiifed imports:

``` haskell
import "base" Control.Applicative (Alternative, (<|>))
import "base" Data.Maybe (Maybe (Nothing), maybe)
import "base" System.IO (IO)
import "yaya" Yaya.Fold (cata, ana)
import "base" Prelude ((+))
```

would become

``` haskell
import "base" Control.Applicative (Alternative, (<|>))
import "base" Data.Maybe (Maybe (Nothing), maybe)
import "base" Prelude ((+))
import "base" System.IO (IO)
import "yaya" Yaya.Fold (cata, ana)
```

gkaracha:
Closes #1189.

Since `importIsPrelude` came first in the fields of `ImportId`, sorting imports through `ImportId` resulted in the implicit separation between `Prelude` and non-`Prelude` imports, effecting placing non-`Prelude` imports first (because `False < True`). In order to avoid having to define a manual `Ord` instance for `ImportId`, I did two things:
1. Make `importIsPrelude` the last field of `ImportId`, so that the initial sorting does not separate `Prelude` imports, and
2. Explicitly relocate the `Prelude` to the end of the list in `normalizeImports` if `ImplicitPrelude` is on.

### DIFF
--- a/data/examples/import/no-implicit-prelude-out.hs
+++ b/data/examples/import/no-implicit-prelude-out.hs
@@ -1,0 +1,6 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+
+import Control.Applicative (Alternative, (<|>))
+import Data.Maybe (Maybe (Nothing), maybe)
+import Prelude ((+))
+import System.IO (IO)
--- a/data/examples/import/no-implicit-prelude-qual-out.hs
+++ b/data/examples/import/no-implicit-prelude-qual-out.hs
@@ -1,0 +1,8 @@
+{-# LANGUAGE PackageImports #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+import "base" Control.Applicative (Alternative, (<|>))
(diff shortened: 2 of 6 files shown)


gkaracha:
Superseded by #1216.

PR #1215, opened 2026-08-18
Fix preservation of the position of comments around `where`

gkaracha:
Closes #784. Replaces #1210.

PR #1216, opened 2026-08-18
Do not place `Prelude` at the end with `NoImplicitPrelude`

### LINKED ISSUE #1189: Don’t single out `Prelude` import with `NoImplicitPrelude`

gkaracha:
Closes #1189. Depends on #1215 (superficial - changelog only). Supersedes #1214.
(Cherry-picked https://github.com/mrkkrp/ormolu/commit/e67dbfe2faca57769755bb3d2240c0f6cc07a0a5.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 * Fixed preservation of the position of comments around the `where` keyword.
   [Issue 784](https://github.com/tweag/ormolu/issues/784).
 
+* Do not sort `Prelude` to the end of the import list when the
+  `NoImplicitPrelude` extension is enabled; instead sort it like any other
+  import. [Issue 1189](https://github.com/tweag/ormolu/issues/1189). Cherry-picked [mrkkrp/ormolu@e67dbfe](https://github.com/mrkkrp/ormolu/commit/e67dbfe2faca57769755bb3d2240c0f6cc07a0a5).
+
 ## Ormolu 0.8.1.1
 
 * Add missing braces for case expressions in single‑line do blocks. [Issue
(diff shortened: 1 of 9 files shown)


qnikst:
Being cherry-picked and for keeping compatibility with Mark's version it makes sense.

After playing a bit more with an examples: package imports, qualified imports, mixed case (I'm not sure if it is reasonable to add that to examples, but could be).

I've used that one:

```
{-# LANGUAGE PackageImports #-}
{-# LANGUAGE NoImplicitPrelude #-}

import Prelude ((+))
import System.IO (IO)
import "base" Control.Applicative (Alternative, (<|>))
import "base" Data.Maybe (Maybe (Nothing), maybe)
import "yaya" Yaya.Fold (ana, cata)
```

To be honest I would attribute Mark explicitly for the fix in the changelog given the state of the project, as it would be fair.


P.S. CI is red at the moment so we will need to sort that out before merging

qnikst (approved):

joaosreis (approved):

PR #1221, opened 2026-09-06
Version: 0.9.0.0

qnikst:
Co-authored-by: Mark Karpov <markkarpov92@gmail.com>

This commit keeps all the changes that makes a version exactly equal to the lastest version released on the hackage.

However as upstream respository was removed it's not possible to recreate exact commit history. And this commit exists in order to keep historical version for the community and not lose work that was done by Mark.

Commit is fully implemented by Mark Karpov, with detailed explanations done in https://web.archive.org/web/20260819093310/https://markkarpov.com/post/ormolu-one-refactor-that-fixed-everything.html


qnikst:
It would be nice to apply this commit to have a version that we can work on that have the latest changes.

I've tried to properly attribute all the changes, but commit sequence recreation will just be incorrect and unfair.
So it worth to apply other improvements on top of this PR.

hseg:
Checking the Wayback Machine, I can only get the titles and shas of the commits this touched, but perhaps this can help in reconstructing the original commit sequence.
Thank you for preserving this work, it'd be a shame to lose it.

https://web.archive.org/web/20260730151921/https://github.com/mrkkrp/ormolu/commits?author=mrkkrp

